PR #34, opened 2024-03-14
fix missing cpuinfo lib and another issue

wenhuach21:


requirements.txt
```diff
@@ -2,6 +2,7 @@ accelerate
 auto-gptq
 datasets
 py-cpuinfo
```
hshen14: already there

PR #44, opened 2024-03-21
add hqq acc data.md

pursure-D:
add HQQ data


docs/acc.md
```diff
@@ -84,6 +84,21 @@ For wikitext2/ptb-new/c4-new ppl, we follow the code of gptq and set the sequenc
         <td>47.44</td>
         <td>61.36</td>
     </tr>
```
wenhuach21, line 86: "With these configurations, the tuning costs for GPTQ, AWQ, and ours are comparable, while HQQ is much faster."

PR #47, opened 2024-03-22
remove amp autocast default cuda setting.

lkk12014402:
remove amp autocast default cuda setting

### DIFF
--- a/auto_round/autoround.py
+++ b/auto_round/autoround.py
@@ -445,8 +445,16 @@ class AutoRound(object):
         self.amp_dtype = torch.float16
         if self.model.dtype != torch.float32:
             self.amp_dtype = self.model.dtype
-        if self.device == "cpu" or self.device == "hpu":
+        if self.device == "cpu":
             self.amp_dtype = torch.bfloat16
+            self.amp_device_type = "cpu"
+
+        if "hpu" in self.device:
+            self.amp_dtype = torch.bfloat16
+            self.amp_device_type = "hpu"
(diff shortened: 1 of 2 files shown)


WeiweiZhang1 (approved):

wenhuach21:
please make sure "cuda:3" and "hpu:4" could work

lkk12014402:
> please make sure "cuda:3" and "hpu:4" could work

add new variable self.amp_device_type

auto_round/autoround.py
```diff
@@ -876,10 +884,16 @@ def quant_block(self, block, input_ids, input_others, q_input=None, device=torch
...
+                    current_input_ids,
+                    current_input_others,
+                    self.amp,
+                    self.amp_dtype,
+                    self.amp_device_type,
+                    device,
                 )
                 if self.amp and not check_is_cpu(device):
-                    with autocast(device_type="cuda", dtype=self.amp_dtype):
+                    with autocast(device_type=self.amp_device_type, dtype=self.amp_dtype):
```
wenhuach21, line 896: a better way is just detect whether ":" in device string, and truncate the string, as there may more device like xpu, npu, etc

lkk12014402: As I know, some code do not check npu or xpu device and we don't validate such devices. If we want to adapt other device, we need validate it and revise code.

wenhuach21:
please review pr https://github.com/intel/auto-round/pull/64

PR #67, opened 2024-03-29
qwen1.5 acc update

yintong-lu:
acc update of qwen1.5 recipes


wenhuach21:
Please identify the reason for the fluctuations in GSM8K.

examples/language-modeling/scripts/Qwen1.5-7B-Chat.sh
```diff
@@ -10,3 +10,11 @@ python3 main.py \
 --sym \
 --minmax_lr 0.01 
 
+# Evaluation command for CEVAL, CMMLU, MMLU and GSM8K tasks
+lm_eval --model hf \
+    --model_args pretrained="gpu_model_path",autogptq=True,gptq_use_triton=True,trust_remote_code=True \
+    --device cuda:0 --tasks ceval-valid,cmmlu,mmlu --batch_size 16
```
wenhuach21: the default num_few_shot should be zero, move the cmd to acc.md

yintong-lu: the default num_fewshot for gsm8k is 5 and 0 for the rest tasks. 
It's been moved to acc.md

PR #68, opened 2024-04-01
update custom data

JustinLin610:
this pr adds custom dataset support. the custom data is a jsonl file where each line is a dict with key 'text'. in the shell script, use `--dataset custom --data_path path_to_jsonl`.

### DIFF
--- a/examples/language-modeling/main.py
+++ b/examples/language-modeling/main.py
@@ -44,6 +44,13 @@ if __name__ == '__main__':
 
     parser.add_argument("--iters", default=200, type=int,
                         help=" iters")
+    
+    parser.add_argument("--dataset", default="NeelNanda/pile-10k", type=str,
+                        choices=["NeelNanda/pile-10k", "mbpp", "custom"],
+                        help="The dataset for quantization training. It can be a custom one.")
+    
+    parser.add_argument("--data_path", default=None, type=str,
+                        help="The data path for quantization training.")
 
     parser.add_argument("--use_quant_input", action='store_true',
                         help="whether to use the output of quantized block to tune the next block")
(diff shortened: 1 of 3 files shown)


examples/language-modeling/main.py
```diff
@@ -254,7 +261,7 @@ def get_library_version(library_name):
                     f"{n} will not be quantized due to its shape not being divisible by 32, resulting in an exporting issue to autogptq")
 
     autoround = round(model, tokenizer, args.bits, args.group_size, sym=args.sym, batch_size=args.train_bs,
-                      seqlen=seqlen, n_blocks=args.n_blocks, iters=args.iters, lr=args.lr,
+                      dataset=args.dataset, data_path=args.data_path, seqlen=seqlen, n_blocks=args.n_blocks, iters=args.iters, lr=args.lr,
```
wenhuach21, line 264: Hi Justin, Thanks for the pr. 
As there are many args for dataset now, I think a better way is merging dataset and data_path.  This could be implemented by 

1 register a local file calibration dataset
@register_dataset("local")
def get_local_dataloader(
    tokenizer, seqlen, dataset_name="./tmp.json", split=["train", "validation", "test"], seed=42, bs=4
):

2 then change the code in (https://github.com/intel/auto-round/blob/main/auto_round/autoround.py#L638)
 from .calib_dataset import CALIB_DATASETS
 if is_local_file(self.dataset_name):
      get_dataloader = CALIB_DATASETS.get("local")
  else:
      get_dataloader = CALIB_DATASETS.get(self.dataset_name, CALIB_DATASETS["NeelNanda/pile-10k"])
 self.dataloader = get_dataloader(
                self.tokenizer,
                self.seqlen,
                seed=self.seed,
                bs=self.train_bs,
                split=self.dataset_split,
                dataset_name=self.dataset_name,
            )

If this works for you, I can implement based on your pr

JustinLin610: sure feel free to make changes. it is all good to me

PR #77, opened 2024-04-16
Support XPU export

Kaihui-intel:
Support XPU export

### DIFF
--- a/auto_round/export/export_to_itrex/export.py
+++ b/auto_round/export/export_to_itrex/export.py
@@ -72,6 +72,50 @@ def save_quantized_as_itrex(output_dir, inplace=True, **kwargs):
         logger.error("Fail to save configure file and weights due to {}.".format(e))
     return compressed_model
 
+@register_format("itrex_xpu")
+def save_quantized_as_itrex_xpu(output_dir, inplace=True, **kwargs):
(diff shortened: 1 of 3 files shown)


auto_round/export/export_to_itrex/export.py
```diff
@@ -145,10 +201,11 @@ def pack_model(
             scale_dtype=scale_dtype,
             zp=zp is not None,
             bias=m.bias is not None,
-            device=device,
-            use_optimum_format=True,
+            device="cuda" if device == "xpu" else device,
```
WeiweiZhang1, line 204: It might be a bit obscure here, can you change it? for example, add a temporary variable quant_device

auto_round/export/export_to_itrex/export.py
```diff
@@ -127,6 +180,9 @@ def pack_model(
         if not isinstance(scale, torch.Tensor):
             scale = torch.tensor(scale, dtype=convert_dtype)
             zp = torch.tensor(zp, dtype=torch.int32)
+            if device == "xpu":
+                scale = torch.tensor(v["scale"], dtype=torch.float32)
```
WeiweiZhang1, line 184: better add a log here, if convert_dtype != torch.float32 there maybe lead to loss of acc

PR #93, opened 2024-04-26
handle invalid layername in weight_config

WeiweiZhang1:


auto_round/export/export_to_itrex/export.py
```diff
@@ -170,17 +170,23 @@ def pack_model(
...
+        try:
+            if "float" in v["data_type"]:
+                continue
+            dtype = v["data_type"]
+            num_bits = v["bits"]
+            group_size = v["group_size"]
+            sym = v["sym"]
+            scale_dtype = v["scale_dtype"]
+            scale, zp = v["scale"], v["zp"]
+        except:
```
wenhuach21: why using try except,   having no scale info in the dict could detect this issue

auto_round/export/export_to_autogptq.py
```diff
@@ -80,7 +80,7 @@ def save_quantized_as_autogptq(output_dir, use_triton=True, inplace=True, **kwar
                 break
         if not is_supported_type:
             continue
-        if not check_to_quantized(m):
+        if not check_to_quantized(m, n):
```
wenhuach21: I still suggest that direct raise the error if the layer name is wrongly set

auto_round/utils.py
```diff
@@ -294,7 +294,6 @@ def get_module(module, key):
     for name in name_list:
         if hasattr(module, name):
             module = getattr(module, name)
-            module = module
```
wenhuach21, line 297: It's unreasonable to return the orig model if the key could not found.  Please change to return None instead

auto_round/autoround.py
```diff
@@ -1319,6 +1319,8 @@ def get_quantized_layer_names_outside_blocks(self):
                 continue
             try:
                 layer = get_module(self.model, key)
+                if layer == self.model:
```
wenhuach21: For a straightforward approach, consider listing all the supported_layers in self.model.named_module

wenhuach21 (approved):

PR #97, opened 2024-04-29
 fix critic bug for gradient_accumulate_steps!=1 and reduce cpu memory  of lm-head tuning

WeiweiZhang1:
Resolve the issue of excessive CPU memory usage caused by saving layer outputs.
Fix the issue of generating random indices when using 'rand' sampler.

### DIFF
--- a/auto_round/autoround.py
+++ b/auto_round/autoround.py
@@ -914,8 +914,8 @@ class AutoRound(object):
             train_bs = self.train_bs
             for i in range(0, self.n_samples, train_bs):
                 end_index = min(self.n_samples, i + train_bs)
-                tmp_inputs = inputs[i:end_index, ...].to(device)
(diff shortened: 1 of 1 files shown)


auto_round/autoround.py
```diff
@@ -914,8 +914,8 @@ def quant_layer(self, layer_name, inputs, q_inputs=None, device=torch.device("cp
             train_bs = self.train_bs
             for i in range(0, self.n_samples, train_bs):
                 end_index = min(self.n_samples, i + train_bs)
-                tmp_inputs = inputs[i:end_index, ...].to(device)
-                tmp_output = layer.forward(tmp_inputs).to(cache_device)
+                tmp_input = inputs[i:end_index, ...].to(device)
```
wenhuach21: this could not reduce cpu memory, move the logic to the for loop in the following, add compute the fp output  each iteration.

PR #153, opened 2024-06-08
fix qbits issue

wenhuach21:


auto_round_extension/qbits/qlinear_qbits.py
```diff
@@ -20,12 +20,6 @@
 import torch.nn as nn
 from auto_round.utils import convert_dtype_torch2str, logger
 QBITS_AVAILABLE = True
-try:
```
zhewang1-intc, line 23: this change will lead qbits.repack_quantized_weight, qbits.woq_linear throw `no attribute` err.
the solution is to add a member named self.qbits, in post_init stage, pls set `self.qbits=intel_extension_for_transformers.qbits`

don't forget to import itrex when we try to get itrex-version in req_check.

i have validated that autoround works well after updating these changes.

zhewang1-intc (changes requested):

auto_round/auto_quantizer.py
```diff
@@ -317,6 +317,13 @@ def convert_model(self, model: nn.Module):
         return model
 
     def _dynamic_import_inference_linear(self, bits, backend):
+        if (not torch.cuda.is_available()) or "qbits" in backend or "intel/cpu" in backend:
+            try:
+                from intel_extension_for_transformers import qbits  # pylint: disable=E0401
+            except Exception as e:
+                raise ImportError("Please install Intel Extension for Transformers via 'pip install "
+                                  "intel-extension-for-transformers' to  inference on Intel CPU")
```
zhewang1-intc: amd CPUs also can gain qbits perf benefits.
consider change to x86 CPUs

zhewang1-intc (approved):
LGTM

WeiweiZhang1 (approved):

PR #160, opened 2024-06-19
example_bugfix

WeiweiZhang1:
add trust_remote_code for eval_042, 
example bugfix, 
enable amp quant_layer of cpu device

### DIFF
--- a/auto_round/autoround.py
+++ b/auto_round/autoround.py
@@ -891,7 +891,7 @@ class AutoRound(object):
                 output_q = block_forward(
                     block, current_input_ids, current_input_others, self.amp, self.amp_dtype, device
                 )
-                if self.amp and not check_is_cpu(device):
+                if self.amp:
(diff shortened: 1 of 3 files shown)


auto_round/autoround.py
```diff
@@ -891,7 +891,7 @@ def quant_block(self, block, input_ids, input_others, q_input=None, device=torch
                 output_q = block_forward(
                     block, current_input_ids, current_input_others, self.amp, self.amp_dtype, device
                 )
-                if self.amp and not check_is_cpu(device):
+                if self.amp:
```
wenhuach21, line 894: amp should be false if cpu does not support bf16

WeiweiZhang1: > amp should be false if cpu does not support bf16

this is done in the set_amp_dtype()

wenhuach21:
change the title

wenhuach21 (approved):

PR #165, opened 2024-06-28
enable llava & Qwen-VL multimodal model quantization

WeiweiZhang1:


auto_round/autoround.py
```diff
@@ -449,6 +457,14 @@ def calib(self, nsamples, bs):
                 for key in data.keys():
                     data_new[key] = data[key].to(self.model.device)
                 input_ids = data_new["input_ids"]
+            elif isinstance(data, tuple) or isinstance(data, list):
+                if self.multimodal:
+                    data_new = {"input_ids": data[0].to(self.model.device), \
```
wenhuach21: better not add key 'multimodal',  only support dict if there are multiples inputs?

auto_round/autoround.py
```diff
@@ -449,6 +457,14 @@ def calib(self, nsamples, bs):
...
+                                "images": data[1].to(self.model.device, dtype=self.model.dtype), "image_sizes": data[2]}
```
wenhuach21: why user still need to tell the image_size, the image_size should already be in images

auto_round/utils.py
```diff
@@ -132,7 +132,7 @@ def get_scale_shape(weight, group_size):
     return shape
 
 
-def to_device(input, device=torch.device("cpu")):
+def to_device(input, device=torch.device("cpu"), multimodal=False):
     """Moves input data to the specified device.
```
wenhuach21: seems never used the multimodel arg

n1ck-guo (approved):

auto_round/utils.py
```diff
@@ -559,12 +595,13 @@ def get_layer_names_in_block(model, supported_types=[torch.nn.Linear, transforme
         if isinstance(m, tuple(supported_types)):
             m.tmp_name = n
     layers_in_block = []
-    block_names = get_block_names(model)
-    for block_name in block_names:
-        block = get_module(model, block_name)
-        for n, m in block.named_modules():
-            if hasattr(m, "tmp_name"):
-                layers_in_block.append(m.tmp_name)
+    all_blocks = get_block_names(model, multimodal)
```
wenhuach21, line 598: is there any way to check whether it is a mutlimodal model from model itself/inputs?

WeiweiZhang1: For a multimodal model there should be a arg that determines whether or not to quantize the part of vision, and the arg here are for that purpose, not to identify the model type.

PR #179, opened 2024-07-09
Fix autoround format accuracy issue

wenhuach21:


auto_round/export/export_to_autoround/export.py
```diff
@@ -159,12 +159,12 @@ def save_quantized_as_autoround(output_dir, inplace=True, backend="autoround:exl
     """
     if ":" not in backend:
         backend = "autoround:exllamav2"
+        backend = backend.replace("autoround", "auto_round")
+        backend = backend.replace("auto-round", "auto_round")
```
WeiweiZhang1: Shouldn't these two lines be outside the if ?

test/test_export.py
```diff
@@ -0,0 +1,154 @@
+import copy
+import shutil
+import sys
+import unittest
+
+sys.path.insert(0, "..")
```
WeiweiZhang1, line 6: should remove this line

WeiweiZhang1 (approved):

test/test_export.py
wenhuach21: for local test

wenhuach21:
auto-gptq format still has accuracy issue

PR #191, opened 2024-07-17
Fix multimodal and moe issue

WeiweiZhang1:
fix mixtral-8x7b model quantization issue

### DIFF
--- a/auto_round/autoround.py
+++ b/auto_round/autoround.py
@@ -43,6 +43,7 @@ from .utils import (
     to_dtype,
     get_layer_names_in_block,
     mv_module_from_gpu,
+    get_multimodal_block_names,
 )
 
 from .low_cpu_mem.utils import get_layers_before_block
@@ -140,11 +141,12 @@ class AutoRound(object):
             dynamic_max_gap: int = -1,
             data_type: str = "int",
             scale_dtype: str = "fp16",
-            multimodal:bool = False,
             act_bits: int = 32,
             act_group_size: int = None,
             act_sym: bool = None,
             act_dynamic: bool = True,
+            multimodal:bool = False,
+            quant_vision:bool = False,
             **kwargs,
     ):
         self.quantized = False
@@ -177,6 +179,7 @@ class AutoRound(object):
 
         self.iters = iters
         self.multimodal = multimodal
+        self.quant_vision = quant_vision
         if self.iters <= 0:
             logger.warning("iters must be positive, reset it to 200")
             self.iters = 200
@@ -235,7 +238,10 @@ class AutoRound(object):
         The quantized model and layer configurations.
         """
         # logger.info("cache block input")
-        all_blocks = get_block_names(self.model, self.multimodal)
+        if self.multimodal:
+            all_blocks = get_multimodal_block_names(self.model, self.quant_vision)
+        else:
+            all_blocks = get_block_names(self.model)
                     
         if len(all_blocks) == 0:
             logger.warning("could not find blocks, exit with original model")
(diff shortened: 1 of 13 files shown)


auto_round/export/export_to_autoround/export.py
```diff
@@ -166,6 +166,7 @@ def save_quantized_as_autoround(output_dir, inplace=True, backend="auto_round:ex
 
     model = kwargs["model"]
     multimodal = kwargs["multimodal"]
+    safe_serialization = True if 'safe_serialization' not in kwargs.keys() else  kwargs["safe_serialization"]
```
wenhuach21, line 169: why need this

auto_round/special_model_handler.py
```diff
@@ -12,11 +12,13 @@
...
 
-from .utils import logger, torch
+import torch
 
 share_attention_mask_tuple = ("baichuan",)
 special_states_dim_tuple = ("chatglm",)
-not_share_position_ids_tuple = ("llava","phi3_v")
+not_share_position_ids_tuple = ("llava", "phi3_v",)
+Vison_blocks_tuple = ("vision", "visual",)
+MoE_blocks_tuple = ("moe", "expert",)
```
wenhuach21: do not handle moe as special model . there are many moe models, which may not name block as expert.

auto_round/export/export_to_autoround/export.py
WeiweiZhang1: for phi3-vision save model process

auto_round/autoround.py
```diff
@@ -140,11 +141,12 @@ def __init__(
             dynamic_max_gap: int = -1,
             data_type: str = "int",
             scale_dtype: str = "fp16",
-            multimodal:bool = False,
             act_bits: int = 32,
             act_group_size: int = None,
             act_sym: bool = None,
             act_dynamic: bool = True,
+            multimodal:bool = False,
+            quant_vision:bool = False,
```
wenhuach21: how about just pass the block_names [list of list] to support this, while the example of mutlimodel could support this args

auto_round/export/export_to_autoround/export.py
wenhuach21: if safe serialization  has no effect for other models, we could use it as the default in the future

auto_round/utils.py
```diff
@@ -232,22 +232,75 @@ def check_is_cpu(device):
     return device == torch.device("cpu") or device == "cpu"
 
 
-def get_block_names(model, multimodal=False):
+def validate_modules(module_names):
+        """
+        Validate a list of modules validity.
```
wenhuach21: could not understand the comment

auto_round/autoround.py
```diff
@@ -100,12 +99,11 @@ class AutoRound(object):
         data_type (str): The data type to be used (default is "int").
         scale_dtype (str): The data type of quantization scale to be used (default is "float16"), different kernels
                            have different choices.
-        multimodal(bool): Enable multimodal model quantization, (default is "False").
         act_bits (int): Number of bits for activation quantization. Default is 32.
         act_group_size (int): Group size for activation quantization. Default is None.
         act_sym (bool): Whether to use symmetric activation quantization. Default is None.
         act_dynamic (bool): Whether to use dynamic activation quantization. Default is True.
-
+        quant_block_list (bool): A list whose elements are list of block's layer names to be quantized.
```
wenhuach21: not bool

auto_round/utils.py
```diff
@@ -234,13 +234,13 @@ def check_is_cpu(device):
 
 def validate_modules(module_names):
         """
-        Validate a list of modules validity.
```
wenhuach21: where it will be used

wenhuach21 (approved):

auto_round/utils.py
WeiweiZhang1: in get_multimodal_block_names

PR #197, opened 2024-07-24
Enable phi3v tuning

WeiweiZhang1:
add phi3v quantization example
enable multimodal autoround format inference


wenhuach21:
Please prioritize supporting inference for these three models

wenhuach21 (approved):

PR #205, opened 2024-08-05
set autoround format as default to unify CPU/HPU/CUDA

wenhuach21:


auto_round/auto_quantizer.py
```diff
@@ -311,10 +326,18 @@ def convert_model(self, model: nn.Module):
 
         layer_names = get_layer_names_in_block(model)
         quantization_config = model.config.quantization_config
+        if hasattr(quantization_config, "backend"):  # pragma: no cover
+            backend = quantization_config.backend
+            if "hpu" in backend and model.dtype != torch.bfloat16:
+                logger.info("We suggest you to set `torch_dtype=torch.bfloat16` for better efficiency with AutoRound.")
+                model = model.to(torch.bfloat16)
+            elif model.dtype != torch.float16:
```
WeiweiZhang1, line 334: only for gpu, better add a judge "cpu" not in backend

WeiweiZhang1 (approved):

PR #211, opened 2024-08-07
limit the scale minimum value not to 0

WeiweiZhang1:
Limit the scale minimum value not to 0
tested on common models(>=10)
currently, only effects Qwen2-57B-A14B-Instruct model


wenhuach21:
please also try just clippling without the if in the future

wenhuach21 (approved):

PR #214, opened 2024-08-07
update xpu format exporting

WeiweiZhang1:
The XPU backend has aligned its convert format with Hugging Face. auto-round should follow up on this change.
Testing on XPU is done

### DIFF
--- a/auto_round/export/export_to_itrex/export.py
+++ b/auto_round/export/export_to_itrex/export.py
@@ -20,7 +20,7 @@ from typing import Dict, List, Optional, Union
 import torch
 import transformers
 from auto_round.export.register import register_format
-from auto_round.utils import get_module, logger, set_module
+from auto_round.utils import get_module, logger, set_module, detect_device
 
 from .config import QuantConfig
 from .model_wrapper import WeightOnlyLinear
@@ -109,7 +109,7 @@ def save_quantized_as_itrex(output_dir, inplace=True, **kwargs):
 
 @register_format("itrex_xpu")
 def save_quantized_as_itrex_xpu(output_dir, inplace=True, **kwargs):
-    """Save configure file and weights for CPU backend inference."""
+    """Save configure file and weights for XPU backend inference."""
     model = kwargs["model"]
     layer_config = kwargs["layer_config"]
     sym = kwargs["sym"]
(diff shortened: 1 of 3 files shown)


auto_round/export/export_to_itrex/export.py
```diff
@@ -185,15 +185,11 @@ def pack_model(
...
-        compression_dim=0,
-        use_optimum_format=False,
-        scale_dtype=convert_dtype_str2torch(config.scale_dtype),
-        device="xpu",
     """
+    # Due to XPU doesn't support tuning yet
+    device = detect_device() if device == "xpu" else device
     if model.device.type == 'meta':
-        model = model.to("cuda" if device == "xpu" else device)
+        model = model.to(device)
```
wenhuach21, line 192: could xpu device run the exporting successfully?

wenhuach21 (approved):

auto_round/export/export_to_itrex/export.py
```diff
@@ -185,15 +185,11 @@ def pack_model(
...
 
-    xpu args:
-        compression_dtype=torch.int8,
-        compression_dim=0,
-        use_optimum_format=False,
-        scale_dtype=convert_dtype_str2torch(config.scale_dtype),
-        device="xpu",
     """
+    # Due to XPU doesn't support tuning yet
+    device = detect_device() if device == "xpu" else device
```
Kaihui-intel: xpu may be detected as "xpu:0". Please verify this function on xpu device.

Kaihui-intel (approved):

auto_round/export/export_to_itrex/export.py
WeiweiZhang1: xpu device will be detected as 'cpu'

wenhuach21: this is not a good  code style. A workaround is  hardcode "cpu" for xpu in this code. In the future, we should support detect xpu in detect_device

WeiweiZhang1: ok，will update it to the cpu device

auto_round/export/export_to_itrex/export.py
WeiweiZhang1: It is actually running on the CPU, so it will logically succeed. However, the speed depends on the machine's CPU efficiency.

PR #222, opened 2024-08-12
[WIP] hadamard support

wenhuach21:


wenhuach21:
negative impact with AutoRound in W2 and W4A4 MXFP4 for llama3

PR #229, opened 2024-08-22
change the scale thresh generally

WeiweiZhang1:
change the scale thresh generally
0.0 --->  1e-5
The effect of scale clip was evaluated on 4 commonly used models with acceptable accuracy fluctuations.
Adding scale clip prevents weight values from disappearing and reduces model instability after quantization.

### DIFF
--- a/auto_round/data_type/int.py
+++ b/auto_round/data_type/int.py
@@ -59,8 +59,6 @@ def quant_tensor_asym(weight, bits=4, v=0, min_scale=1.0, max_scale=1.0, scale_d
     wmax[tmp] = +1
     scale = ((wmax - wmin) / maxq).to(scale_dtype)
     scale = torch.clamp(scale, min=q_scale_thresh)
-    if (scale == 0.).any():
-        scale = torch.clamp(scale, min=1e-5)
     zp = round_ste(-wmin / scale)  # pylint: disable=E1130
     scale = scale.unsqueeze(dim=-1)
     zp = zp.unsqueeze(dim=-1)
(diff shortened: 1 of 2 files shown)


auto_round/data_type/int.py
```diff
@@ -59,8 +59,6 @@ def quant_tensor_asym(weight, bits=4, v=0, min_scale=1.0, max_scale=1.0, scale_d
     wmax[tmp] = +1
     scale = ((wmax - wmin) / maxq).to(scale_dtype)
     scale = torch.clamp(scale, min=q_scale_thresh)
-    if (scale == 0.).any():
-        scale = torch.clamp(scale, min=1e-5)
```
wenhuach21, line 63: where is the clip? better using q_scale_thresh

wenhuach21 (approved):

auto_round/quantizer.py
```diff
@@ -46,7 +46,7 @@ def reshape_tensor(v, group_size=-1):
 
 def quant_tensor(
         quant_func, data, bits=4, group_size=-1, v=0, min_scale=1.0, max_scale=1.0, scale_dtype=torch.float16,
-        weight_min=None, weight_max=None, q_scale_thresh=0.0, **kwargs,
+        weight_min=None, weight_max=None, q_scale_thresh=1e-5, **kwargs,
```
wenhuach21, line 49: we have a thresh value in wrapper linear, better pass that value

wenhuach21:
add clip in unwrapper function

wenhuach21 (approved):

PR #262, opened 2024-09-18
tmp branch, delete later

n1ck-guo:


auto_round/quantizer.py
```diff
@@ -255,6 +257,8 @@ def __init__(self, orig_layer, enable_minmax_tuning=True, enable_norm_bias_tunin
                 torch.zeros(self.orig_layer.weight.shape, device=self.device, dtype=weight_dtype),
                 self.group_size),
             requires_grad=True)
+        if self.bits == 8:
+            self.value = torch.tensor(0.0, device=self.device, dtype=weight_dtype)
```
wenhuach21: have a check that value is not changed. I haven't coded like this before

examples/language-modeling/main.py
```diff
@@ -448,9 +470,11 @@
             user_model = None
             if args.act_bits <= 8:
                 user_model = model.to(device_str)
+                user_model = user_model.to(torch.float16)
```
wenhuach21: why using this, this could be removed

PR #263, opened 2024-09-18
better support quant_lm_head for larger models

wenhuach21:


wenhuach21:
@mxjmtxrm this should could solve your problem

n1ck-guo (approved):

PR #281, opened 2024-10-16
qwen2_bugfix, add adamround vision UT

WeiweiZhang1:


auto_round/autoround.py
```diff
@@ -721,7 +720,7 @@ def forward(m, hidden_states, *positional_args, **kwargs):
                             else:
                                 self.inputs[name][key] = list(torch.split(alibi.to("cpu"), 1, dim=0))
                     elif "position_ids" in key or 'cache_position' in key:
-                        if self.train_bs == 1 and self.not_share_rotary_pos_emb_flag:
+                        if self.train_bs == 1 and self.not_share_position_ids_flag:
```
wenhuach21, line 724: we could refine this later,  we could add a shareable, contatable, etc for each key. In the future, we'd better not change the code in autoround.py but change the code in special_model_handle

wenhuach21 (approved):

wenhuach21 (approved):

PR #297, opened 2024-11-04
mllm eval bug fix

n1ck-guo:
* fix bug for evalkit verbose use
* remove require for dataset arg when only use auto_round --eval command

### DIFF
--- a/auto_round/script/mllm.py
+++ b/auto_round/script/mllm.py
@@ -47,7 +47,7 @@ class BasicArgumentParser(argparse.ArgumentParser):
         self.add_argument("--asym", action='store_true',
                             help=" asym quantization")
 
-        self.add_argument("--dataset", required=True, type=str,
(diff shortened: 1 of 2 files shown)


wenhuach21 (approved):

auto_round/script/mllm.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, *args, **kwargs):
         self.add_argument("--asym", action='store_true',
                             help=" asym quantization")
 
-        self.add_argument("--dataset", required=True, type=str,
+        self.add_argument("--dataset", type=str, default=None,
```
wenhuach21, line 50: why there is no default dataset?

n1ck-guo: Cause for llava dataset, the argument 'dataset' must be a path

wenhuach21: how about using one in https://huggingface.co/datasets?search=llava

PR #299, opened 2024-11-05
refine model config file for mixed precision quantization

wenhuach21:


wenhuach21:
close due to conflict

PR #304, opened 2024-11-05
fix glm4-9b batch dim issue

wenhuach21:


auto_round/autoround.py
```diff
@@ -283,8 +289,10 @@ def quantize(self):
             keys = inputs.keys()
             input_id_str = [key for key in keys if key.startswith('hidden_state')]
             if len(input_id_str) != 1:
-                raise RuntimeError("hidden_states arg mismatch error," \
-                                   " please check the input kwargs of block forward for more details.")
+                raise RuntimeError(f"hidden_states arg mismatch error,"
+                                   "please raise an issue in https://github.com/intel/auto-round/issues"
+                                   " or try to set the `batch_size` to 1 and "
```
WeiweiZhang1: This dismatch is independent of batch_size, just indicating that this model's hidden_state is not a regular name

wenhuach21: Setting batch_size=1 should handle these scenarios, as we cache all the information when the batch size is 1. If this is not yet supported, please address it in a future update.

WeiweiZhang1 (approved):

PR #321, opened 2024-11-13
fix multiple device bug

wenhuach21:


auto_round/utils.py
```diff
@@ -906,7 +906,7 @@ def compile_func_on_hpu(func):
 
 
 def compile_func_on_cuda_or_cpu(func, enable_torch_compile):
-    if enable_torch_compile or TORCH_VERSION_AT_LEAST_2_6_PRE_RELEASE:
+    if enable_torch_compile or (TORCH_VERSION_AT_LEAST_2_6_PRE_RELEASE and enable_torch_compile!=False):
```
WeiweiZhang1, line 909: Can it be simplified to 'if enable_torch_compile' ?

WeiweiZhang1 (approved):

auto_round/utils.py
wenhuach21: when enable_torch_compile is set to None and torch version is >=2.6,  automatically set enable_torch_compile  to True

PR #329, opened 2024-11-20
fix mllm dataset config bug and add gptq cuda backend

wenhuach21:


wenhuach21:
@n1ck-guo  should support calibration dataset automatic switch for quant_nontext_module

n1ck-guo (approved):

PR #345, opened 2024-11-27
[HPU]Enhance `numba` check

yiliu30:
Porting https://github.com/intel/neural-compressor/pull/2068, to handle cases where `TBB` is installed but not properly configured.

### DIFF
--- a/auto_round/utils.py
+++ b/auto_round/utils.py
@@ -980,3 +980,64 @@ def compile_func(fun, device, enable_torch_compile):
         return compile_func_on_cuda_or_cpu(fun, enable_torch_compile)
 
 
+
+def is_numba_available():  # pragma: no cover
+    """Check if Numba is available."""
+    try:
+        import numba
+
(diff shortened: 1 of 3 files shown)


auto_round/utils.py
```diff
@@ -980,3 +980,64 @@ def compile_func(fun, device, enable_torch_compile):
...
+        return True
+    except ImportError:
+        return False
+
+
+def _is_tbb_installed():  # pragma: no cover
+    import importlib.metadata
+
+    try:
+        importlib.metadata.version("tbb")
```
wenhuach21, line 998: better add a function, check_package_install or similar.  you may implement it via from transformers.utils.versions import require_version

wenhuach21 (approved):

auto_round/utils.py
yiliu30: Agree, will port such helper functions in a separate PR.

PR #362, opened 2024-12-02
fix bug of only_text_test check due to inference issue on cpu

n1ck-guo:


auto_round/script/mllm.py
```diff
@@ -269,6 +269,8 @@ def tune(args):
             args.device = ",".join(map(str, range(len(devices))))
             devices = args.device.replace(" ", "").split(',')
         use_auto_mapping = True
+    elif args.device == "auto":
```
wenhuach21, line 274: change this in llm too.  llm may need to change two place, tuning and evaluation

wenhuach21: besides, if you want to use `auto` to use multigpu,  it's not the expected behavior in the main code.  The gpu device number is controlled by --devices now

wenhuach21 (changes requested):
could not simply use auto to enable auto-mapping

auto_round/mllm/autoround_mllm.py
```diff
@@ -157,7 +157,8 @@ def __init__(
         from ..calib_dataset import CALIB_DATASETS
         from .mllm_dataset import MLLM_DATASET
         if isinstance(dataset, str):
-            if quant_nontext_module or (dataset in CALIB_DATASETS.keys() and not _only_text_test(model, tokenizer, device)):
+            if quant_nontext_module or \
+                (dataset in CALIB_DATASETS.keys() and not _only_text_test(model, tokenizer, device)):
```
wenhuach21, line 161: you should need to set the check running on GPU even though the default device has been changed

wenhuach21 (approved):

Issue #361, opened 2024-11-30
auto-round asym quantization inference needs to remind user install from source in the code


wenhuach21:

wenhuach21:
https://github.com/intel/auto-round/pull/363

Issue #360, opened 2024-11-30
fp_layers fuzzy match unexpected behaviour


wenhuach21:
for example, model.layers.1 will exclude model.layers.1 and model.layers.11, model.layer12 ,etc.

wenhuach21:
Besides, to_quant_block_names has not been changed correctly, which will impact auto-gptq format

Issue #357, opened 2024-11-29
export to awq exception due to transformers 4.7 API change


wenhuach21:
from transformers.modeling_utils import shard_checkpoint
ImportError: cannot import name 'shard_checkpoint' from 'transformers.modeling_utils'

PR #363, opened 2024-12-02
fix awq backend and fp_layers issue

wenhuach21:
fixed https://github.com/intel/auto-round/issues/361
fixed https://github.com/intel/auto-round/issues/360
fixed https://github.com/intel/auto-round/issues/357


auto_round/utils.py
```diff
@@ -1041,3 +1043,41 @@ def can_pack_with_numba():  # pragma: no cover
...
+            all_layer_names.append(n)
+    not_to_quantized_layers = []
+
+    for fp_layer in fp_layers:
+        if fp_layer == "":
+            continue
+        if fp_layer in all_layer_names:
+            not_to_quantized_layers.append(fp_layer)
+            continue
+        if fp_layer[-1] != ".":
```
WeiweiZhang1, line 1077: judge  whether the last character is an integer value should be more in line with expectations

WeiweiZhang1 (approved):

auto_round/utils.py
wenhuach21: the code is to handle this scenario, model.layer.1 should only exclude model.layer.1,should not exclude model.layer.11

PR #369, opened 2024-12-04
quant_block_names enhancement

WeiweiZhang1:


auto_round/utils.py
```diff
@@ -318,6 +318,22 @@ def validate_modules(module_names, quant_vision=False, vison_blocks_names=None):
...
+    common_prefix = split_paths[0]
+    for path in split_paths[1:]:
+        common_prefix = [comp for comp, other in zip(common_prefix, path) if comp == other]
+    return '.'.join(common_prefix)
+
+def extract_block_names_to_str(quant_block_list):
+    if not isinstance(quant_block_list, (list,tuple)):
+        return None
+    # Extract common prefix for each list
+    prefixes = [get_common_prefix(blocks) for blocks in quant_block_list]
```
wenhuach21, line 333: should have a double check if the common perfix include some other layers, it have, should set it to 16 bits

auto_round/export/export_to_autogptq/export.py
```diff
@@ -130,8 +131,8 @@ def save_quantized_as_autogptq(output_dir, inplace=True, backend="auto_gptq:exll
         processor.save_pretrained(output_dir)
     ##check module quantized in block, this may have bug for mixed precision quantization
     quantization_config = kwargs["serialization_dict"]
-    if bool(to_quant_block_names):
-        all_blocks = to_quant_block_names
+    if bool(quant_block_list):
```
wenhuach21, line 134: gptq only supporst one string

WeiweiZhang1: The quant_block_list will not be saved, it is just passed from the autoround class for export reference.

wenhuach21 (approved):

PR #376, opened 2024-12-06
fix gpu ut

n1ck-guo:


auto_round/mllm/autoround_mllm.py
```diff
@@ -34,6 +35,7 @@
 def _only_text_test(model, tokenizer, device):
     """Test if the model whether can use text-only datasets."""
     try:
+        device = detect_device(device)
```
wenhuach21, line 38: why need to add this, the device should have been  detected in autoround initialization

n1ck-guo: this test func happened before autoround initialization

wenhuach21 (approved):

PR #387, opened 2024-12-16
Fix auto-round install & bump into 0.4.4

XuehaoSun:


requirements-gpu.txt
```diff
@@ -0,0 +1,5 @@
+torch
+transformers>=4.38
+threadpoolctl
+lm-eval>=0.4.2,<0.5
+auto-gptq>=0.7.1
```
wenhuach21, line 1: the above 4 should be in requirement

XuehaoSun: > the above 4 should be in requirement

All platforms will install requirement.txt, we can't add incompatible packages

wenhuach21: The basic requirement must support tuning on GPU, otherwise, the user experience will suffer. We may release a whl for HPU.

yiliu30:
@XuehaoSun Please help update the HPU-CI to use `auto-round-lib`.

yiliu30:
I'm testing the `auto-round-lib` on the INC side. https://github.com/intel/neural-compressor/pull/2102 cc @XuehaoSun

wenhuach21 (approved):

PR #419, opened 2025-01-17
fix evaluation device bug, relate to issue 413

n1ck-guo:


auto_round/script/llm.py
```diff
@@ -28,6 +28,7 @@
 import os
 import re
 import argparse
+from typing import Union, Optional
```
wenhuach21: if vlm has similar issue, we need to fix it too

auto_round/script/llm.py
```diff
@@ -209,20 +215,16 @@ def setup_parser():
...
+    parser.add_argument("--iters", "--iter", default=1000, type=int, help="iterations to tune each block")
 
-    parser.add_argument("--seqlen", "--seq_len", default=2048, type=int,
-                        help="sequence length of the calibration samples")
+    parser.add_argument(
+        "--seqlen", "--seq_len", default=2048, type=int, help="sequence length of the calibration samples")
 
-    parser.add_argument("--nsamples", default=512, type=int,
-                        help="number of samples")
+    parser.add_argument("--nsamples", default=512, type=int, help="number of samples")
```
WeiweiZhang1: also recommanded to add --nsample

WeiweiZhang1 (approved):

PR #444, opened 2025-02-17
fix bug of evaluate user model

n1ck-guo:


auto_round/script/llm.py
```diff
@@ -558,7 +558,8 @@ def tune(args):
                 dispatch_model(model, model.hf_device_map)
                 user_model = model
             else:
-                user_model = model  # .to(device_str)
+                device_str = detect_device(device_str)
```
wenhuach21, line 568: mllm?

auto_round/utils.py
```diff
@@ -1184,7 +1184,9 @@ def set_cuda_visible_devices(device):
             devices = device.replace(" ", "").split(',')
         if len(devices) > 1:  ##for 70B model on single card, use auto will cause some layer offload to cpu
             parallelism = True
-        device_str = None
+            device_str = None
+        else:
+            device_str = detect_device(devices[0])
```
wenhuach21: it  is quite confusing if device is set to 0,1 and detect_device return 0?

auto_round/utils.py
```diff
@@ -1184,7 +1184,9 @@ def set_cuda_visible_devices(device):
...
     elif device == "auto":
         device_str = None
```
wenhuach21: why set_cuda_visible_devices also set parallelism,  it's not a good coding style. Set cuda visible device first, then set parallelism.
Similar coding style issue in detect_devices

auto_round/utils.py
n1ck-guo: if multi card or auto set, will return None. If only one card set, will return like cuda:1.

auto_round/script/llm.py
n1ck-guo: Not affected

auto_round/utils.py
wenhuach21: then why using devices[0] not devices

wenhuach21: if the device is set to "cpu" or "hpu", is devices[0] still correct?

wenhuach21 (approved):

PR #448, opened 2025-02-25
GGUF support step2: add naive Q2_KS and Q4_KS

n1ck-guo:


auto_round/wrapper.py
```diff
@@ -167,24 +168,39 @@ def _qdq_weight(self, value, min_scale, max_scale):
...
-            min_scale=min_scale,
-            max_scale=max_scale,
-            scale_dtype=self.orig_layer.scale_dtype,
-            tensor_min=self.weight_min,
-            tensor_max=self.weight_max,
-            data_type=self.data_type,
-            q_scale_thresh=self.q_scale_thresh,
-
-        )
+        if self.dq:
```
wenhuach21: change the return values to weight_q, (scale,d_scale), (min,d_min), thus could call the same function

auto_round/autoround.py
```diff
@@ -162,6 +162,8 @@ def __init__(
             enable_norm_bias_tuning: bool = False,
             enable_torch_compile: bool = False,
             device_map: Union[str, dict] = None,
+            super_bits: int = None,
```
wenhuach21, line 165: adam and opt

auto_round/autoround.py
```diff
@@ -523,7 +528,7 @@ def set_layerwise_config(self, layer_config):
         """
         layers_in_blocks = get_layer_names_in_block(self.model, self.supported_types, self.quant_block_list)
         keys = ["data_type", "bits", "group_size", "sym", "scale_dtype", "act_bits", "act_group_size", "act_sym",
-                "act_dynamic", "act_data_type"]
+                "act_dynamic", "act_data_type", "super_bits", "super_group_size"]
```
wenhuach21, line 531: Better not dump too much information to the model. If double quant is not enabled, can we remove this info? The same applies to activation.

auto_round/data_type/int.py
```diff
@@ -128,7 +128,7 @@ def quant_tensor_asym_dq(tensor, bits=4, group_size=-1, v=0, min_scale=1.0, max_
     qdq_result = (scale * q - wmin_m).to(tensor.dtype)
     qdq_result = revert_tensor_by_pad(qdq_result, orig_shape=orig_shape, pad_len=pad_len)
     zp = round_ste(wmin_m / scale)  # remove this later
-    return qdq_result, scale, zp
+    return qdq_result, {"scale": scale, "d_scale": d_scale}, {"wmin_m": wmin_m, "d_wmin_m": d_wmin_m}
```
wenhuach21, line 131: support float zp later for q4_0,q4_1 later

auto_round/export/export_to_gguf/quant.py
```diff
@@ -13,59 +13,102 @@
...
+
+    assert new_data.dtype == np.uint8
+    assert new_data.shape[-1] == type_size
     new_data = new_data.reshape(*shape[:-1], shape[-1] // block_size * type_size)
     return new_data
 
+
+def np_roundf(n: np.ndarray) -> np.ndarray:
+    a = abs(n)
+    floored = np.floor(a)
```
wenhuach21, line 83: suppport cuda packing later

auto_round/export/export_to_gguf/config.py
```diff
@@ -0,0 +1,41 @@
...
+    "bits": 4,
+    "act_bits": 16,
+    "super_group_size": 8,
+    "super_bits": 6,
+    "group_size": 32,
+    "asym": True,
+    "data_type": "int_asym_dq"
+}
+
+GGUF_CONFIG["gguf:q2_k"] = GGUF_CONFIG["gguf:q2_k_s"] = {
```
wenhuach21: I think we should name this to q2_ks as it is different from q2_k which has mixed precision

wenhuach21 (approved):

auto_round/export/export_to_gguf/config.py
n1ck-guo: renamed, current only support q2_k_s, q4_k_s

auto_round/autoround.py
```diff
@@ -522,8 +527,12 @@ def set_layerwise_config(self, layer_config):
         None
         """
         layers_in_blocks = get_layer_names_in_block(self.model, self.supported_types, self.quant_block_list)
-        keys = ["data_type", "bits", "group_size", "sym", "scale_dtype", "act_bits", "act_group_size", "act_sym",
-                "act_dynamic", "act_data_type"]
+        breakpoint()
+        keys = ["data_type", "bits", "group_size", "sym", "scale_dtype"]
+        if self.act_bits <= 8:
+            keys.extend(["act_bits", "act_group_size", "act_sym","act_dynamic", "act_data_type"])
```
wenhuach21: this simple change is not correct

wenhuach21: please make sure this is correct, I remember changing like this is not correct

auto_round/script/llm.py
```diff
@@ -186,6 +186,12 @@ def __init__(self, *args, **kwargs):
 
         self.add_argument("--device_map", default=None, type=str, help="device_map for block in tuning phase")
 
+        self.add_argument(
+            "--super_group_size", default=-1, type=int, help="the number of super group size when use double quant.")
```
wenhuach21: better not to set the default value to -1, as -1 has a special meaning for group_size.

auto_round/script/llm.py
```diff
@@ -564,7 +568,10 @@ def tune(args):
         enable_torch_compile=enable_torch_compile,
         act_data_type=args.act_data_type,
         act_dynamic=not args.disable_act_dynamic,
-        device_map=args.device_map)
+        device_map=args.device_map,
+        super_group_size=args.super_group_size,
```
wenhuach21, line 572: better to add a check, other formats does not support dq

PR #481, opened 2025-03-31
support xpu in tuning and inference

wenhuach21:
TODO:  issue=xpu could not be automatically used when not using auto backend

### DIFF
--- a/auto_round/auto_quantizer.py
+++ b/auto_round/auto_quantizer.py
@@ -126,7 +126,7 @@ class AutoHfQuantizer:
             else:
                 quantization_config = AutoQuantizationConfig.from_dict(quantization_config)  # pylint: disable=E1101
         quant_method = quantization_config.quant_method
-    
+
         # Again, we need a special care for bnb as we have a single quantization config
         # class for both 4-bit and 8-bit quantization
         if quant_method == QuantizationMethod.BITS_AND_BYTES:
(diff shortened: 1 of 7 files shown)


auto_round/auto_quantizer.py
```diff
@@ -385,29 +392,29 @@ def detect_device(self, target_backend, orig_backend):
...
-                return "hpu"
-            else:
-                return "cpu"
+            return self.detect_auto_device()
 
         # Find the backend and determine the device type from BackendInfos
         backend = self.find_backend(target_backend)
         if backend is None:
-            raise ValueError("Backend not found, please set it to 'auto' to have a try ")
+            raise ValueError("Backend is not found, please set it to 'auto' to have a try ")
```
n1ck-guo, line 407: set it to 'auto' and try again?

n1ck-guo (approved):

auto_round/auto_quantizer.py
```diff
@@ -385,29 +392,29 @@ def detect_device(self, target_backend, orig_backend):
             return "cuda"
         elif "hpu" in target_backend:
             return "hpu"
+        elif "xpu" in target_backend:
+            return "xpu"
         elif "cpu" in target_backend:
             return "cpu"
```
n1ck-guo, line 398: How about:
for device_type in ["cuda", "hpu", "xpu", "cpu"]:
    if device_type in target_backend:
         return device_type

auto_round/auto_quantizer.py
n1ck-guo: I feel that this sentence should not add 'is', Backend not found.

PR #484, opened 2025-04-01
fix gemma3 cache data issue

WeiweiZhang1:


auto_round/autoround.py
```diff
@@ -997,7 +997,14 @@ def forward(m, hidden_states=None, *positional_inputs, **kwargs):
                             self.inputs[name][key] = [data]
                         else:
                             data = post_process_cache_data(self.batch_size, data, key)
-                            self.inputs[name][key] = list(torch.split(data, 1, dim=self.batch_dim))
+                            try:
+                                self.inputs[name][key] = list(torch.split(data, 1, dim=self.batch_dim))
+                            except:
+                                logger.error(
+                                f"this model has not been supported, "
+                                f"please raise an issue in https://github.com/intel/auto-round/issues"
```
wenhuach21, line 1005: could batch_size =1 automatically be set? we could throw a specific error and catch it in some function and try bs=1 again. if batch_size is already 1, just throw exception.

auto_round/special_model_handler.py
```diff
@@ -15,7 +15,7 @@
 import torch
 from collections import UserDict
 shareable_keywords = ("position_ids", "cache_position", "position_embeddings",
-                        "position_embeddings_global", "position_embeddings_local")
+                        "position_embeddings_global", "position_embeddings_local", "slope_rate")
```
wenhuach21, line 18: As we may release new version soon, to minimize potential risks, we  better temporarily relocate the newly introduced keys to the specialized model handler.

wenhuach21:
https://github.com/intel/auto-round/pull/515

PR #496, opened 2025-04-08
[WIP]enable llama4 quantization, only for bs=1 currently

WeiweiZhang1:


wenhuach21:
Not planed. An unconventional model file where the main parameters are not wrapped in linear layers, which is not supported on the kernel side

PR #522, opened 2025-04-16
enable llama4 int8 quantization baseline

WeiweiZhang1:
for Xeon SGlang INT8 llama4 model workround

pull this PR code
cd auto-round/
# build from source
pip install -e .
sh run_llama4_quant.sh

### DIFF
--- a/auto_round/utils.py
+++ b/auto_round/utils.py
@@ -40,7 +40,12 @@ supported_formats = (
 
 supported_formats = supported_formats + tuple(GGUF_CONFIG.keys())
 
-supported_layer_types = (torch.nn.Linear, transformers.modeling_utils.Conv1D)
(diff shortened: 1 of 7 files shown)


auto_round/utils.py
```diff
@@ -1136,6 +1141,130 @@ def get_device_and_parallelism(device):
...
+
+
+def pack_to_int8(model, output_dir):
+    import json
+    from safetensors.torch import save_file
+    with torch.no_grad():
+        state_dict = translate_2_sglang_int8(model)
+    max_shard_size = 40 * 1024**3  # 40GB
+    shards = {}
+    current_shard = {}
```
wenhuach21, line 1214: this part should be refined and supported in main branch.
Better follow the origin code style if possible

PR #567, opened 2025-05-21
Add script for cuda unit test

XuehaoSun:
Add script for cuda unit test
`bash .azure-pipelines/scripts/ut/run_ut_cuda.sh`


.azure-pipelines/scripts/ut/run_ut_cuda.sh
```diff
@@ -39,9 +39,10 @@ function run_unit_test() {
     create_conda_env
     cd ${REPO_PATH}/test/test_cuda
 
-    uv pip install -r requirements.txt
-    uv pip install -v git+https://github.com/ModelCloud/GPTQModel.git@v2.1.0 --no-build-isolation
     uv pip install -v git+https://github.com/casper-hansen/AutoAWQ.git --no-build-isolation
+    uv pip install -v git+https://github.com/ModelCloud/GPTQModel.git@v2.1.0 --no-build-isolation
```
wenhuach21: We should avoid restricting the version of the main dependency in the future.

XuehaoSun: I did not use the latest version of `GPTQModel` because it does not support python 3.10 and below

Issue #565, opened 2025-05-20
sentence transformers quantization error


wenhuach21:
example code
~~~python
from sentence_transformers import SentenceTransformer
import auto_round as ar

model = SentenceTransformer("all-MiniLM-L6-v2", model_kwargs={"torch_dtype": "bfloat16"})
for n,m in model.named_modules():
    print(n)
ar_model = ar.AutoRound(model[0].auto_model, model.tokenizer, bits=4, group_size=128, sym=True, nsamples=128, iters=200)
ar_model.quantize()
~~~

error

~~~bash
2025-05-20 03:17:49 WARNING _logger.py L72: When quantization encounters tensor shape mismatch error, you can try to avoid it with batch_size=1
2025-05-20 03:17:49 ERROR autoround.py L908: The expanded size of the tensor (2048) must match the existing size (512) at non-singleton dimension 1.  Target sizes: [8, 2048].  Tensor sizes: [1, 512]
2025-05-20 03:17:49 INFO autoround.py L596: caching done
  0%|                                                     | 0/6 [00:00<?, ?it/s]Traceback (most recent call last):
  File "/home/wenhuach/auto-round/main_test4.py", line 9, in <module>
    ar_model.quantize()
  File "/home/wenhuach/auto-round/auto_round/autoround.py", line 600, in quantize
    inputs = all_inputs[block_names[0]]
KeyError: 'encoder.layer.0'
  0%|                                                     | 0/6 [00:01<?, ?it/s]
ERROR conda.cli.main_run:execute(41): `conda run python /home/wenhuach/auto-round/main_test4.py` failed. (See above for error)
~~~bash


this should be related to batch tensor

PR #582, opened 2025-05-28
fix tensor shape mismatch error for API usage

WeiweiZhang1:
fix issue:
https://github.com/intel/auto-round/issues/565

### DIFF
--- a/auto_round/autoround.py
+++ b/auto_round/autoround.py
@@ -205,8 +205,8 @@ class AutoRound(object):
         self.minmax_lr = minmax_lr or self.lr
 
         self.data_type = data_type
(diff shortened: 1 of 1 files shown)


auto_round/script/mllm.py
```diff
@@ -328,6 +328,17 @@ def tune(args):
 
     from auto_round import AutoRoundMLLM
 
+    seqlen = args.seqlen
```
wenhuach21, line 331: Better move all the checkers  to API side, then the API usage and cmd usage share the same behavior

wenhuach21 (approved):

PR #595, opened 2025-06-05
fix lots of gguf critic bugs and support imatrix in rtn mode

wenhuach21:
- [x] 3/6 bits imatrix
- [x] tuning gguf quantizing embedding and lm-head
- [x] fix q2_k_s_pure gguf exporting
- [x] verify moe model
- [ ] fallback to block-wise way at rtn if oom
- [x] fix the same issue in cpu packing or leverage the same code of gpu packing
- [x] gguf exporting folder name needs to be changed
- [x] quantize embedding and lm in rtn mode
- [x] fix weight permute issue


n1ck-guo (approved):

wenhuach21:
merge first

PR #596, opened 2025-06-06
support for q4_k_m

n1ck-guo:


auto_round/utils.py
```diff
@@ -1554,3 +1557,87 @@ def check_seqlen_compatible(input_seqlen, tokenizer=None, model=None):
...
+    else:
+        return None
+
+def get_layer_config_by_gguf_format(layer_config, gguf_format, model):
+    # TODO: support for other format later
+    if "q4_k_m" not in gguf_format:
+        return layer_config
+
+    import gguf
+    from auto_round.export.export_to_gguf.convert import Model
```
wenhuach21, line 1577: It's better to return a dictionary to store the configuration (e.g., (q4_k_s, q5_k_s)) for each layer, which will help facilitate automatic mixed-bit quantization later.

wenhuach21 (approved):

PR #601, opened 2025-06-09
fix vlm uttest path error

WeiweiZhang1:


test/test_cuda/test_vlms.py
```diff
@@ -183,10 +183,10 @@ def test_quant_not_text(self):
                                   bits=bits, group_size=group_size, sym=sym, iters=1, nsamples=1,quant_nontext_module=True)
         autoround.quantize()
 
-        quantized_model_path = "./saved/Phi-3.5-vision-instruct-w4g128-auto_round"
+        quantized_model_path = "./saved/Phi-3.5-vision-instruct-w4g128"
```
wenhuach21, line 186: Why has the name of the quantization folder changed? The only PR that might have renamed it is yours—the one that changed _ to -.

WeiweiZhang1: I guess the PR that changed the name should have merged quite a long time ago, when there is only one pack format, save_dir does not add the format name.

 it's just that cuda's vlm test is often skipped because it doesn't meet the environment configuration requirements Everyone didn't test it.

PR #624, opened 2025-06-27
fix vlm hf config loading issue

WeiweiZhang1:


auto_round/utils.py
```diff
@@ -1334,13 +1334,21 @@ def mllm_load_model(
...
     from transformers import AutoProcessor, AutoTokenizer, AutoModelForCausalLM
-    from huggingface_hub import HfApi, HfFileSystem
+    from huggingface_hub import HfApi, hf_hub_download
 
     if os.path.isdir(pretrained_model_name_or_path):
         config = json.load(open(os.path.join(pretrained_model_name_or_path, "config.json")))
     else:
-        hf_file = HfFileSystem()
-        config = json.load(hf_file.open(pretrained_model_name_or_path + "/config.json"))
+        from huggingface_hub import hf_hub_download
```
wenhuach21: should we use different codse for different versions of hf_hub?

WeiweiZhang1: It's very stable, from v0.10.0 to v0.33.1

auto_round/utils.py
```diff
@@ -1334,13 +1334,21 @@ def mllm_load_model(
...
+        import gzip
+        try:
+            config_path = hf_hub_download(pretrained_model_name_or_path, "config.json")
+            with open(config_path, "r", encoding="utf-8") as f:
+                config = json.load(f)
+        except Exception:
+            config_path = hf_hub_download(pretrained_model_name_or_path, "config.json.gz")
```
wenhuach21, line 1352: move import gzip to here, and better have a check of whether having gzip  or not

PR #646, opened 2025-07-07
support llmcompressor format

xin3he:
For whisper quantization, INT8 W8A8 is enough.
To leverage vLLM, we save it in llmcompressor format, which is using compressed-tensor inside.

### DIFF
--- a/auto_round/export/export_to_llmcompressor/config.py
+++ b/auto_round/export/export_to_llmcompressor/config.py
@@ -1,0 +1,55 @@
+# Copyright (c) 2025 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+quantization_config = {
+"config_groups": {
+    "group_0": {
+    "input_activations": {
+        "actorder": None,
+        "block_structure": None,
+        "dynamic": True,
+        "group_size": None,
+        "num_bits": 8,
+        "observer": None,
+        "observer_kwargs": {},
+        "strategy": "token",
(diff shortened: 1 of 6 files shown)


auto_round/export/export_to_llmcompressor/config.py
```diff
@@ -0,0 +1,55 @@
...
+        "symmetric": True,
+        "type": "int"
+    }
+    }
+},
+"format": "int-quantized",
+"global_compression_ratio": 1.5,  # not exactly correct, but a placeholder
+"ignore": [],
+"kv_cache_scheme": None,
+"quant_method": "compressed-tensors",
```
wenhuach21, line 53: how about adding a `provider:"auto-round"`  to the model config,

xin3he: good suggestion

auto_round/export/export_to_llmcompressor/export.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Returns:
+        torch.Tensor: The recovered floating-point weight tensor.
+    """
+    return (qdq_weight / scale).to(torch.int8)
+
+
+
+@torch.no_grad()
+def save_fake_to_llmcompressor_format(output_dir, model=None, **kwargs):
```
wenhuach21: it's better to follow the name style save_quantized_to_xxx though the name is not good enough

auto_round/export/export_to_llmcompressor/export.py
```diff
@@ -0,0 +1,73 @@
...
+    if output_dir is not None and tokenizer is not None:
+        tokenizer.save_pretrained(output_dir)
+    if output_dir is not None and processor is not None:
+        processor.save_pretrained(output_dir)
+
+    # generate q_weight
+    for n, m in model.named_modules():
+        if isinstance(m, WrapperWALayer):
+            m = m.orig_layer
+            q_weight = recover_qweight(m.weight, m.scale)
```
wenhuach21: we could try a larger model, such as a 7B model. The packing time should less than 1 minute. If it's significantly longer, we can consider using CUDA or XPU to speed up the packing process.

xin3he: It's less than 20s for whisper-large-v3 on CPU. But I agree to use the detected device.

wenhuach21 (approved):
It is also recommended to support immediate packing later, as done in other formats, to reduce ram usage when handling extremely large models.

PR #667, opened 2025-07-17
handle zeros in imatrix

wenhuach21:


auto_round/data_type/gguf.py
```diff
@@ -338,13 +338,43 @@ def quant_tensor_gguf_asym_dq(
         wmin = wmin.view(-1, 1)
         scale = scale.view(-1, 1)
     else:
+        logger.info(f"{torch.min(imatrix)},{torch.max(imatrix)}")
```
wenhuach21: some logs need deleting later

PR #670, opened 2025-07-18
support to use lm_eval for vlm

n1ck-guo:


wenhuach21 (approved):

auto_round/script/llm.py
```diff
@@ -210,6 +210,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.add_argument(
             "--model", "--model_name", "--model_name_or_path", default="facebook/opt-125m", help="model name or path")
+        self.add_argument(
+            "--lm_eval_model", default="hf", help="Name of lm_eval model, e.g. 'hf' of 'hf-multimodal'"
```
wenhuach21: this arg name is confusing, and it would be better rename(or reuse --mllm )and  move to eval_args

auto_round/script/llm.py
```diff
@@ -210,6 +210,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.add_argument(
             "--model", "--model_name", "--model_name_or_path", default="facebook/opt-125m", help="model name or path")
+        self.add_argument(
+            "--mllm", default=False, help="whether to eval multi-model."
```
wenhuach21: multi-model is incorrect

PR #673, opened 2025-07-21
restrict accelerate version to reduce ram usage

wenhuach21:


.azure-pipelines/unit-test.yml
```diff
@@ -13,6 +13,7 @@ pr:
       - test
       - setup.py
       - requirements.txt
+      - requirements-cpu.txt
```
wenhuach21, line 16: @XuehaoSun It seems we haven't installed IPEX for the previous UTs, right? If that's the case, we're using the Torch backend for the CPU UTs.

PR #694, opened 2025-07-25
support to save image_processor

n1ck-guo:


auto_round/export/export_to_autogptq/export.py
```diff
@@ -137,12 +137,15 @@ def save_quantized_as_autogptq(output_dir, inplace=True, backend="auto_gptq:exll
     quant_block_list = kwargs.get("quant_block_list", get_block_names(model))
     tokenizer = kwargs.get("tokenizer", None)
     processor = kwargs.get("processor", None)
+    image_processor = kwargs.get("image_processor", None)
```
wenhuach21, line 138: where do we pass the image_processor? Please make sure it will be saved if the origin model have

n1ck-guo: in utils.py mllm_load_model

PR #699, opened 2025-07-31
 support fp8 model and str as input in llm quantization

wenhuach21:
- [x] RTN mode
- [x] lm-head check


wenhuach21:
**Important Notice**
the default value of tokenzier, low_cpu_mem_usage, device in API has been changed

PR #703, opened 2025-07-31
FIXBUG: CPU Offloading for Cache Blocks in Low-Memory GPU Systems or Single GPU on ROCM Configs

JartX:
Description
This Pull Request introduces a new feature designed to optimize memory usage, particularly beneficial for systems with integrated GPUs (like many Intel platforms) or discrete GPUs with limited memory. The new --force_cpu_caching option allows users to explicitly instruct the application to cache intermediate model block inputs directly on the CPU instead of the GPU.

This capability is especially valuable in scenarios where GPU memory (VRAM) is a bottleneck, as offloading this data to system RAM frees up crucial GPU resources. This can significantly improve the ability to run larger models or process more extensive datasets that might otherwise exceed the GPU's memory capacity.

Changes Introduced
--force_cpu_caching Argument: A new boolean argument has been added. When enabled, this flag temporarily shifts the model's device to the CPU specifically for the caching operation.

Intelligent Device Management: If --force_cpu_caching is active, the model is briefly moved to the CPU to cache intermediate data (all_inputs). Upon completion of the caching process, the model's device is automatically restored to its original configuration (e.g., the integrated Intel GPU or a discrete GPU). This ensures that subsequent model operations continue to execute on the intended hardware.

Logging: Informative log messages have been added to indicate when the force_cpu_caching flag is detected and activated, and when the original device is restored.

Impact
This enhancement provides greater flexibility and robustness to our application. It allows for better adaptability across various hardware configurations, notably benefiting systems leveraging Intel integrated graphics or other GPUs with restricted memory. Users encountering out-of-memory issues or performance degradation due to VRAM limitations now have a direct solution to alleviate pressure on GPU memory.

### DIFF
--- a/auto_round/utils.py
+++ b/auto_round/utils.py
@@ -2066,7 +2066,7 @@ def out_of_vram(error_msg):
     # XPU
     if "UR_RESULT_ERROR_OUT_OF_DEVICE_MEMORY" in error_msg:
         return True
-    #ROCM
+    # ROCM
     if "HIP out of memory. Tried to allocate" in error_msg:
         return True
     return False
--- a/auto_round/utils.py
+++ b/auto_round/utils.py
@@ -2066,4 +2066,7 @@ def out_of_vram(error_msg):
     # XPU
     if "UR_RESULT_ERROR_OUT_OF_DEVICE_MEMORY" in error_msg:
         return True
+    #ROCM
+    if "HIP out of memory. Tried to allocate" in error_msg:
+        return True
     return False


wenhuach21:
thanks for the pr. When` low_gpu_mem_usage` is enable, the data will move from gpu to cpu

wenhuach21:
> @wenhuach21 thanks for your answer Yes, however, when generating the cache blocks, this was not being applied. As a result, it would trigger a HIP memory error at that stage, since it attempted to load around 30–40 GB into the GPU's VRAM — even with low_gpu_mem_usage enabled — ultimately halting the process. Now, I’m able to quantize Qwen3-30B-A3B on my 7900XTX using ROCm. Another idea I considered was reusing the low_gpu_mem_usage flag for this purpose, though that decision is ultimately up to the Intel team. Thank you very much for your time!
sorry, I think the root cause it we have not catch the oom in AMD GPU. We will first move the model to gpu and try to calibration, if its out of vram, it will fallback to cpu. So if could catch the oom, the issue will be fixed. Please correct me if I am wrong
https://github.com/intel/auto-round/commit/32d2cd512bd1d984c90cc9b97b0a283c64cba415

JartX:
@wenhuach21 Works fine, thanks for your suggestions and guidance! I've pushed the changes. :)

wenhuach21:
> @wenhuach21 Works fine, thanks for your suggestions and guidance! I've pushed the changes. :)

Thanks for your contribution to our repo. Looking forward to more prs from you

JartX:
@wenhuach21 
This seems strange to me, because it used to fail consistently before, but now it sometimes fails and sometimes doesn’t, and the process gets interrupted before falling back to the CPU. Only un cache block

The only way it always works correctly on ROCm is with the previous version of the code, which avoids hitting that failure point altogether.

What has been correctly fixed is that when the model doesn't fit in VRAM, it now throws the appropriate error.

Would you mind if I include the previous code, but controlled through the low_gpu_mem_usage flag?

wenhuach21:
> @wenhuach21 This seems strange to me, because it used to fail consistently before, but now it sometimes fails and sometimes doesn’t, and the process gets interrupted before falling back to the CPU. Only at cache block
> 
> The only way it always works correctly on ROCm is with the previous version of the code, which avoids hitting that failure point altogether.
> 
> What has been correctly fixed is that when the model doesn't fit in VRAM, it now not generate the appropriate error.
> 
> Would you mind if I include the previous code, but controlled through the low_gpu_mem_usage flag?

could you try this pr first?
https://github.com/intel/auto-round/pull/708

PR #705, opened 2025-08-01
fix xpu oom checker

n1ck-guo:


auto_round/utils.py
```diff
@@ -2053,3 +2053,17 @@ def check_need_act_calibration(is_act_dynamic, act_data_type=None):
     if act_data_type is not None and "static" in act_data_type:
         return True
     return False
+
+
+def check_oom(error_msg):
```
wenhuach21: check_gpu_oom?

wenhuach21 (approved):

auto_round/utils.py
wenhuach21: or vram?

auto_round/utils.py
```diff
@@ -36,6 +36,7 @@
 
 
 class SupportedFormats:
+
```
wenhuach21, line 39: delete

PR #708, opened 2025-08-01
catch oom error and move to cpu directly

n1ck-guo:


auto_round/autoround.py
```diff
@@ -794,7 +795,8 @@ def quantize_embedding_layer(self):
                             for k in ["bits", "group_size", "super_bits", "super_group_size", "scale_dtype"]
                         },
                     )
-                else:
+                except Exception as e:
+                    logger.error(cuda_error_msg)
```
wenhuach21, line 799: if cpu still throws exception, logger its exception
if cpu works fine, then check gpu error, if gpu error not in out of vram error, logger a warning

JartX:
@wenhuach21 
This implementation works and no longer triggers the HIP memory error. However, the GPU becomes unresponsive once VRAM usage reaches around 99%. Although it eventually drops to ~2% after the "switch to CPU to cache block inputs" step, the GPU remains frozen and cannot be interacted with.

Additionally, I see memory-related errors in dmesg, which suggests that this is likely not an issue with AutoRound, CUDA, or Intel XPU support. It seems to occur specifically with AMD GPUs under ROCm.

From what I can tell, the actual freeze occurs when falling back to the CPU after the GPU driver has already failed due to memory exhaustion. So the fallback logic executes too late to prevent the system instability.

wenhuach21:
> @wenhuach21 You're right — it is indeed slower, but only during the block cache stage.
> 
> During quantization, the model actually fits within the VRAM. For instance, Qwen3-30B-A3B uses around 20GB of VRAM on my 7900XTX. However, during the block cache phase, it tries to allocate close to 50GB of VRAM.
> 
> So I'd say this issue only occurs at that specific point and under that specific condition.
> 
> Are you sure we couldn't introduce a force_block_cache_cpu flag?
> 
> That would avoid having to use low_gpu_mem_usage, since the model does fit in VRAM during quantization. This way, the process wouldn’t run slower overall, and we could still prevent the driver from crashing — as the 50GB peak would be handled by the CPU instead.
> 
> If that's the case and you're open to it, please let me help with the implementation.

If "low_gpu_mem_usage" is enabled,  you'll need about 10GB, I don't know, but it's about 30% during quantization.

Please try again. In most cases, using the CPU for calibration is sufficient, since only the embedding output needs to be cached. However, for cases like LM head quantization or VLMs, CPU-based calibration can be very slow, users could leverage rtn mode "--iters 0 " which is extremely fast and low cost

JartX:
@wenhuach21 works fine

JartX:
Outside of this PR, would you mind if I open another one to add support for quantization with ROCm? It would include a Dockerfile and a markdown file with an explanation.

wenhuach21:
> Outside of this PR, would you mind if I open another one to add support for quantization with ROCm? It would include a Dockerfile and a markdown file with an explanation.

Sounds great! We welcome and truly appreciate all contributions.  Since we don’t have access to such devices, we’re relying entirely on you.

JartX:
@wenhuach21  With your permission, I’m documenting the motivation behind this PR, referencing the previous one and the commit from which this improvement originates, to maintain clear traceability of the changes.

This PR improves the CPU-based block caching mechanism introduced in the previous PR:
https://github.com/intel/auto-round/pull/703
Specifically, it builds upon the following commit:
https://github.com/JartX/auto-round/commit/58e709d28cf61003eae4c5e2f448b4ca5a289c4a

Unlike PR #703, which required the user to explicitly enable CPU block caching via a flag, this PR handles it automatically—enhancing the user experience by removing the need for manual configuration.

This implementation avoids both the HIP memory error under ROCm and the subsequent GPU freeze or instability caused when the AMDGPU kernel driver runs out of memory. Even when proper fallback mechanisms were in place after an OOM, the driver would often leave the GPU in an unusable state.

Together, PRs #703 and #708 address OOM-related failures across different ROCm environments and platforms—both during quantization (when the full model doesn't fit in VRAM) and during the block cache stage, which attempts to load large amounts of data into VRAM, as described by @wenhuach21.

Thanks to this improvement, the entry barrier for quantizing large models using auto-round is now significantly lower. By offloading the block cache to the CPU when needed, it's possible to quantize large models such as:
https://huggingface.co/Qwen/Qwen3-Coder-30B-A3B-Instruct

On a machine equipped with a Radeon 7900XTX (24GB VRAM) and 64GB of RAM.

During quantization, if lm_head layers are skipped, VRAM usage can remain around ~19GB. This allows users with more modest LLM setups to quantize large models directly on the GPU, without having to fall back to full CPU quantization—greatly speeding up the process.

I would say that, at this point, auto-round has become one of the most efficient quantization libraries available.

Special thanks to @wenhuach21 for accepting and supporting the implementation of this fix, and for the collaboration, support, and improvement and refinement of the initial PR code and rework of the out_of_vram mechanism to make it more solid and robust.

PR #721, opened 2025-08-07
update gguf convert.py and support for gpt-oss

n1ck-guo:
* support for gtp-oss
* extract convert and auto download dependence file convert_hf_to_gguf.py


auto_round/autoround.py
```diff
@@ -582,6 +582,10 @@ def parse_format_to_list(self, format: str) -> list:
         from auto_round.utils import SUPPORTED_FORMATS
 
         for format_ in formats:
+            if "gguf" in format_:
+                from auto_round.export.export_to_gguf.convert import download_convert_file
+
+                download_convert_file()
```
wenhuach21: There's no need to download this file every time, especially since it requires the latest GGUF, which may involve reinstalling  gguf. How about adding a check first, if the model has been  already supported, we can skip the download?

wenhuach21 (approved):

auto_round/autoround.py
wenhuach21: another issue is the downloading may be failed due to network issue or some other issues

PR #727, opened 2025-08-11
new cast_to_nvfp4 with high performance

xin3he:


wenhuach21 (approved):

auto_round/data_type/nvfp.py
```diff
@@ -19,19 +19,19 @@
...
-    x[(x >= 0.0) & (x <= 0.25)] = 0.0
-    x[(x > 0.25) & (x < 0.75)] = 0.5
-    x[(x >= 0.75) & (x <= 1.25)] = 1.0
-    x[(x > 1.25) & (x < 1.75)] = 1.5
-    x[(x >= 1.75) & (x <= 2.5)] = 2.0
-    x[(x > 2.5) & (x < 3.5)] = 3.0
-    x[(x >= 3.5) & (x <= 5.0)] = 4.0
-    x[x > 5.0] = 6.0
+
+    step1 = torch.round(2.0 * x) / 2.0
```
wenhuach21, line 28: change torch.round to round_ste in autoround side, as we need tuning sometimes

wenhuach21: if round_ste is used,  I think cast_to_fp4_ste could be deleted

wenhuach21: if this one is better, we could change the implementation of mxfp4 too

xin3he: I think keeping the scast_to_fp4_ste might provide better performance. since the torch.round is called three times. round_ste is complex than torch.round. we can only prost-process it with cast_to_fp4_ste once.

auto_round/data_type/nvfp.py
```diff
@@ -222,6 +224,14 @@ def fp4_v2(tensor, bits=4, group_size=32, v=0, max_scale=1.0, **kwargs):
 
 
 if __name__ == "__main__":
```
yiliu30, line 226: Please help move tests into the test folder, others LGTM.

yiliu30 (approved):

auto_round/data_type/nvfp.py
xin3he: Will be later. Currently many tests exists in code, we need to handle them in batch.
@wenhuach21 I will track it as our P2 task. let me know if you have any comments.

PR #729, opened 2025-08-12
add original convert file and support for the newest llama.cpp

n1ck-guo:
llama.cpp add new logic for mistral, modify the affected code


wenhuach21:
we now support FP8 model as input, please have a check whether exporting to gguf is ok  or not

PR #731, opened 2025-08-12
fix bug for exporting afp8 fake format

n1ck-guo:


auto_round/autoround.py
```diff
@@ -671,9 +671,9 @@ def _check_supported_format(self, format: str) -> bool:
                     )
                     format = "fake"
             else:
-                if not (format == "auto_round" or format == "auto_round:fp8"):
+                if not (format == "auto_round" or format == "auto_round:fp8" or format == "fake"):
                     logger.warning(
-                        f"Currently only support to export auto_round format for static W{self.bits}AFP8 model,"
+                        f"Currently only support to export auto_round or fake format for static W{self.bits}AFP8 model,"
```
wenhuach21, line 678: how about if format==fake: return fake. and move this to the beginning of this function

PR #735, opened 2025-08-14
support for glm4.5

n1ck-guo:


wenhuach21 (approved):

auto_round/autoround.py
```diff
@@ -899,7 +899,7 @@ def get_imatrix_hook(module, input, output):
                     module.imatrix = squared
                     module.imatrix_cnt = input.shape[0]
                 else:
-                    module.imatrix += squared
+                    module.imatrix += squared.to(module.imatrix.device)
```
wenhuach21, line 914: why need this

n1ck-guo: Bugs have occurred. module.imatrix and squared on different device, cpu and cuda

wenhuach21: Could you double-check?  I’m not sure how or why that happens.

PR #737, opened 2025-08-15
Add Static FP8 KV Support

yiliu30:


wenhuach21:
Please note that kv cache does not affect tuning as we only use forward in tuning. So the first change is moving the arg in init to save_quantized?

yiliu30:
> Please note that kv cache does not affect tuning as we only use forward in tuning. So the first change is moving the arg from init to save_quantized?

Since the other non-tuning args, such as `enable_torch_compile`, are defined in `__init__`, how about keeping this one in `__init__` as well?

wenhuach21:
> > Please note that kv cache does not affect tuning as we only use forward in tuning. So the first change is moving the arg from init to save_quantized?
> 
> Since the other non-tuning args, such as `enable_torch_compile`, are defined in `__init__`, how about keeping this one in `__init__` as well?

Due to the large number of arguments in __init__, only the arguments likely to be used are explicitly listed, while the others are hidden in **kwargs. I assume that the static KV cache is not frequently used, so we can keep it hidden in **kwargs for now.

wenhuach21:
> > the code could be improved by providing another arg like static_kv_dtype to replace the exstiing enable_static_kv to support more dtype, fp4, int4/int8, etc.
> 
> Since only FP8 KV is supported in the vLLM end, I’d prefer to keep it as is in this PR. I’m open to adding support for more data types if needed.

Ok, then how about changing the arg now to make the API more backward compatibility in case we need to support other dytpes in the future. only support bf16/fp16/fp8 for now, while for 16 bits, nothing needs to do in the code

yiliu30:
> > > the code could be improved by providing another arg like static_kv_dtype to replace the exstiing enable_static_kv to support more dtype, fp4, int4/int8, etc.
> > 
> > 
> > Since only FP8 KV is supported in the vLLM end, I’d prefer to keep it as is in this PR. I’m open to adding support for more data types if needed.
> 
> Ok, then how about changing the arg now to make the API more backward compatibility in case we need to support other dytpes in the future. only support bf16/fp16/fp8 for now, while for 16 bits, nothing needs to do in the code

Sure, updated.

wenhuach21 (approved):

PR #739, opened 2025-08-15
support bf16 scale and bias

wenhuach21:
autogptq add torch version restriction
int_sym code refine
zp sym packing optimization


wenhuach21:
transformers is ok, as we have kernels to backup autogptq kernels. For vllm, only 4btis and 8 bits if fine.

wenhuach21:
for bits other than 4,8, we use the  fp16

PR #750, opened 2025-08-21
mllm supports loading fp8 model and fix bug of loading fp8 model

n1ck-guo:


auto_round/autoround.py
```diff
@@ -2331,7 +2331,7 @@ def quantize_block(self, block, input_ids, input_others, q_input=None, device=to
         """
         if hasattr(self.model, "is_fp8"):
             for n, m in block.named_modules():
-                if m.__class__.__name__ == "FP8Linear":
+                if hasattr(m, "is_fp8"):
```
wenhuach21, line 2334: why make this change, convert_fp8_layer_to_linear could not skip non-linear layer

auto_round/utils.py
```diff
@@ -1343,6 +1343,20 @@ def fake_cuda():
     return orig_func
 
 
+def is_fp8_model(model: torch.nn.Module) -> bool:
```
wenhuach21: it would be better to change the name as you attach some infos to module

auto_round/utils.py
```diff
@@ -2223,6 +2252,7 @@ def convert_fp8_model_to_16b_model(model, dtype=torch.bfloat16):
     Convert a model with FP8 quantized layers to a model with 16-bit linear layers.
     This is useful for compatibility with other frameworks or for further processing.
     """
+    breakpoint()
```
wenhuach21: delete

PR #752, opened 2025-08-22
api support for fp8 model and mllm api support load from str 

n1ck-guo:


auto_round/mllm/autoround_mllm.py
```diff
@@ -130,7 +129,7 @@ class AutoRoundMLLM(AutoRound):
     def __init__(
         self,
         model: torch.nn.Module,
```
wenhuach21: add str

wenhuach21:
merge first, use other pr for ut

PR #756, opened 2025-08-22
fix_script_fp_layer_config_for_bits_checking

WeiweiZhang1:


auto_round/script/llm.py
```diff
@@ -485,11 +485,16 @@ def tune(args):
     layer_config = {}
     not_quantize_layer_names = get_fp_layer_names(model, args.fp_layers)
     for name in not_quantize_layer_names:
-        layer_config[name] = {"bits": 16, "act_bits": 16}
+        layer_config[name] = {"bits": 16, "act_bits": 16, "data_type": "float", "act_data_type": "float"}
```
wenhuach21, line 488: why this is required?  there should be same issue in API side, we'd better to fix it in auto_round.py code

test/test_cpu/test_export.py
```diff
@@ -269,9 +269,16 @@ def test_mxfp4_llmcompressor_format(self):
         from transformers import AutoConfig
 
         bits = 4
-        data_type = "mx_fp4e2m1"
+        data_type = "mx_fp"
```
wenhuach21, line 272: mx_fp4e2m1 should be supported,  better to fix the code if not

test/test_cpu/test_export.py
```diff
@@ -311,7 +323,7 @@ def test_mxfp8_llmcompressor_format(self):
         from transformers import AutoConfig
 
         bits = 8
-        data_type = "mx_fp8e4m3_rceil"
+        data_type = "mx_fp_rceil"
```
wenhuach21, line 326: fix the code here https://github.com/intel/auto-round/blob/main/auto_round/data_type/mxfp.py#L186
then mx_fp8e4m3_rceil could be suppported

auto_round/script/llm.py
wenhuach21: fix the bug at https://github.com/intel/auto-round/blob/main/auto_round/autoround.py#L1734. If the user set the bits/act bits for the specific layer, we should change the dtype/act_dtype

wenhuach21: if dytpe is supported, raise the error

test/test_cpu/test_export.py
WeiweiZhang1: mx_fp8e4m3_rceil  is enabled already

auto_round/script/llm.py
WeiweiZhang1: will fix API config on other PR

PR #758, opened 2025-08-24
add autoround._generate_recipe()

xin3he:
```python
layer_config, avg_bits = autoround._generate_recipe(
        # same data type config as before
        mp_dtype={
            "data_type": "mx_fp8",
            "act_data_type": "mx_fp8",
        },
        # special mix-precision configuration
        mp_config={
            "mp_ratio": 1/3,
            "loss_weight": 2.0,
            "numel_weight": 1.0,
        },
)
autoround.layer_config = layer_config
autoround.quantize()
```
This code is used for INC accuracy tuning, currently only `mx_fp8` is supported for mixing with `mx_fp4` and `nv_fp4`.
image

### DIFF
--- a/auto_round/autoround.py
+++ b/auto_round/autoround.py
@@ -38,6 +38,8 @@ from auto_round.utils import (
     SUPPORTED_LAYER_TYPES,
     TORCH_VERSION_AT_LEAST_2_6,
     CpuInfo,
+    _generate_block_recipe,
+    _generate_recipe,
     _gguf_args_check,
     block_forward,
     check_and_mark_fp8_model,
@@ -67,9 +69,9 @@ from auto_round.utils import (
     infer_bits_by_data_type,
     init_cache,
     is_debug_mode,
+    is_hpex_available,
     is_mx_fp,
     is_nv_fp,
-    is_optimum_habana_available,
     is_standard_fp,
     llm_load_model,
     logger,
@@ -100,6 +102,11 @@ class AutoRound(object):
         nsamples (int): Number of calibration samples.
         enable_torch_compile (bool): Whether to enable torch.compile for quant blocks/layers.
     """
+
+    # If function is not necessary for AutoRound, putting it in other place and
+    # assembling the function here, can improve code readability and maintainability
+    _generate_recipe = _generate_recipe
+    _generate_block_recipe = _generate_block_recipe
 
     def __init__(
         self,
@@ -364,7 +371,7 @@ class AutoRound(object):
 
         torch.set_printoptions(precision=3, sci_mode=True)
 
-        if is_optimum_habana_available():
+        if is_hpex_available():
             logger.info("Optimum Habana is available, import htcore explicitly.")
             import habana_frameworks.torch.core as htcore  # pylint: disable=E0401
             import habana_frameworks.torch.hpu as hthpu  # pylint: disable=E0401]
@@ -2453,7 +2460,11 @@ class AutoRound(object):
                 try:
                     update_fused_layer_global_scales(module)
                 except:
-                    pass  # mix-precision may cause error, since q,k,v are not the same dtype.
+                    # mix-precision may cause error, since q,k,v are not the same dtype.
+                    logger.warning_once(
+                        "Cannot keep the same global scale for fused layers, "
+                        + "so the model may not work with vLLM with fused QKV or else."
+                    )
         round_params = []
         minmax_params = []
         for n, m in block.named_modules():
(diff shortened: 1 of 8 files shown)


auto_round/autoround.py
```diff
@@ -2561,7 +2571,7 @@ def quantize_block(self, block, input_ids, input_others, q_input=None, device=to
             logger.info(f"{unquantized_layer_names} have not been quantized")
         with torch.no_grad():
             unwrapper_block(block, best_params)
-        if self.enable_quanted_input:
+        if self.enable_quanted_input and hasattr(self, "formats"):
```
xin3he, line 2588: I think autoround.quantize is still necessary for debugging, and in this case, formats is not existing.

auto_round/data_type/mxfp.py
```diff
@@ -77,7 +77,6 @@ def quant_element(tensor, ebits, mbits, max_norm, mantissa_rounding="even"):
     return tensor
 
 
-@torch.compile()
 def quant_mx(tensor, bits=4, group_size=-1, v=0, max_scale=1.0, mantissa_rounding="even", data_type="mx_fp", **kwargs):
```
xin3he, line 81: this torch.compile will cause error on HPU.

auto_round/data_type/nvfp.py
```diff
@@ -89,6 +89,7 @@ def nv_fp4(tensor, bits=4, group_size=16, v=0, global_scale=None, **kwargs):
         tensor_max = tensor.abs().max().to(torch.float32)
         global_scale = FLOAT8_E4M3_MAX * FLOAT4_E2M1_MAX * get_reciprocal(tensor_max)
     global_scale = global_scale.to(tensor.device)
+    global_scale = global_scale.to(torch.float32)
```
xin3he, line 93: For opt-125m, the dtype of tensor is float16 so the global_scale cannot be fp32 according to the code logic.

auto_round/data_type/mxfp.py
wenhuach21: you could use torch.compile(func) to control it . without torch compile,  mx tuning is 2x slow on cuda

auto_round/autoround.py
```diff
@@ -2954,6 +2984,142 @@ def sampling_inputs(cls, input_ids, input_others, indices, seqlen, batch_dim=0,
 
         return current_input_ids, current_input_others
 
+    def _generate_recipe(
+        self,
+        # same data type config as before
+        mp_dtype={
```
wenhuach21: Is it possible to either inherit this class or move these functions to other folders? The class has become too long, and we’re working on decoupling its components

auto_round/utils.py
```diff
@@ -2502,3 +2508,78 @@ def is_mx_fp(backend):
 
 def is_nv_fp(backend):
     return BackendDataType.NV_FP in backend
+
+
+######################################### Recipe related codes ####################################
+def create_mp_block(block, mp_layers, mp_dtype):
+    from auto_round.wrapper import WrapperLinear
```
wenhuach21, line 2549: help add types for each args and return values, we are trying to refine our code to follow google style

auto_round/data_type/nvfp.py
wenhuach21: it's better to add this comment to  L92

auto_round/autoround.py
```diff
@@ -2439,7 +2446,10 @@ def quantize_block(self, block, input_ids, input_others, q_input=None, device=to
 
             modules = block.modules()
             for module in modules:
-                update_fused_layer_global_scales(module)
+                try:
+                    update_fused_layer_global_scales(module)
+                except:
+                    pass  # mix-precision may cause error, since q,k,v are not the same dtype.
```
wenhuach21: why not raising the error?

wenhuach21:
Have you tested it on an MoE model? It might require some special handling.

auto_round/utils.py
```diff
@@ -2502,3 +2532,263 @@ def is_mx_fp(backend):
...
+            )
+            total_loss += loss
+        if is_hpex_available():
+            htcore.mark_step()
+        return loss
+
+    combination_list = []
+    numel_list = []
+    loss_list = []
+    for hp_layers in combinations(quantizable_layers, quantizable_num):
```
wenhuach21, line 2759: How about creating a new file in the experimental folder and moving these there?

auto_round/autoround.py
```diff
@@ -2616,6 +2641,29 @@ def quantize_blocks(
         clear_memory()
         input_ids = to_device(input_ids, self.cache_device)
         input_others = to_device(input_others, self.cache_device)
+        if self.recipe_mode:
```
wenhuach21, line 2643: It would be better to wrap this new code into a function and call it as early as possible.

auto_round/autoround.py
```diff
@@ -2954,6 +3002,21 @@ def sampling_inputs(cls, input_ids, input_others, indices, seqlen, batch_dim=0,
 
         return current_input_ids, current_input_others
 
+    def _dump_average_bits(self, layer_config=None):
```
wenhuach21: This function cannot be used by AutoRound, since layers are converted to QuantizedLinear after quantization. If the function can correctly dump average bits in typical scenarios such as INT4, I’d prefer to keep it in the class. Otherwise, it would be better to move it elsewhere for now.

PR #771, opened 2025-08-27
fix gguf bug by int zp

n1ck-guo:


wenhuach21:
we should have q40 ut in cpu, if not, please add an ut later to test this funciton

PR #772, opened 2025-08-27
fix bug of auto_round exporting

n1ck-guo:
fix bug of auto_round exporting when layer_config inclue unsupport layers

### DIFF
--- a/test/test_cpu/test_autoround.py
+++ b/test/test_cpu/test_autoround.py
@@ -376,9 +376,7 @@ class TestAutoRound(unittest.TestCase):
             dataset=self.llm_dataloader,
             layer_config=layer_config,
         )
-        quantized_model_path = self.save_folder
-        autoround.quantize_and_save(output_dir=quantized_model_path, format="auto_round", inplace=True)
-        shutil.rmtree(quantized_model_path, ignore_errors=True)
+        autoround.quantize()
 
     def test_fallback_layers(self):
         bits, group_size, sym = 4, 128, True


auto_round/export/export_to_autoround/export.py
```diff
@@ -301,6 +301,8 @@ def save_quantized_as_autoround(output_dir, inplace=True, backend="auto_round:ex
             block_name_to_quantize[i] = os.path.commonprefix(block_name_to_quantize[i]).rstrip(".")
 
     for layer_name in layer_config:
+        if "in_blocks" not in layer_config[layer_name]:
```
wenhuach21: Could you provide a bit more detail about the issue? Is there an alternative way to fix it, for example by setting the in_blocks attribute for these layers? Otherwise, it may confuse developers if we always need to check each key before using it.

n1ck-guo: Setting a unsupported types layer in the layer_config will trigger this bug, for example layer_config = {"model.embed_tokens": {"bits": 16}}. Another way to fix this is to modify the _set_layerwise_config func. Add attr 'in_blocks'  for layers which is not in supported_types but in layer_config.

wenhuach21: if the layer is unsupported, we could log a warning and  remove it in layer_config. This is an embedding layer,  only gguf could support it

auto_round/autoround.py
```diff
@@ -1727,7 +1727,21 @@ def _set_layerwise_config(self, layer_config: dict) -> bool:
         for n, m in self.model.named_modules():
             # Skip unsupported types
             if not isinstance(m, supported_types) and m.__class__.__name__ not in self.inner_supported_types:
-                continue
+                if n in self.layer_config:
+                    if not check_to_quantized(layer_config[n]):
+                        if isinstance(m, torch.nn.Embedding):
+                            if hasattr(self, "formats") and any("gguf" in format_ for format_ in self.formats):
+                                self.layer_config.pip(n)
```
wenhuach21: pop?

auto_round/autoround.py
```diff
@@ -1727,7 +1727,21 @@ def _set_layerwise_config(self, layer_config: dict) -> bool:
...
+                                continue
+                        else:
+                            self.layer_config.pop(n)
+                            continue
+                    elif not isinstance(m, torch.nn.Embedding):
```
wenhuach21, line 1739: check bits, if bist/act_bits are >=16, no warning is needed

test/test_cpu/test_autoround.py
```diff
@@ -358,6 +358,28 @@ def test_rtn(self):
...
+    
+    def test_embed_quant(self):
+        bits, group_size, sym = 4, 128, True
+        model_name = "facebook/opt-125m"
+        layer_config = {
+            "model.decoder.embed_tokens": {"bits": 4},
+        }
+        autoround = AutoRound(
+            model_name,
+            self.tokenizer,
```
wenhuach21: no need to pass tokenizer

auto_round/autoround.py
wenhuach21: I think you still need to set the in_block attribute as users may use the `quantize` API then there is no formats

PR #774, opened 2025-08-29
[API CHANGE] Stage 1 add quant scheme and consolidate device and device_map

wenhuach21:
- [x] wrapper config to class
- [x] layer_config support config class
- [x] backward compatibility, users could still run AutoRound(bits=xxx,...)
- [x] add ut
- [x] modify adam
- [x] support gguf


wenhuach21:
In the next PR, I’ll consolidate `device` and `device_map` in the API. For now, I’d appreciate a thorough review of this PR.

auto_round/schemes.py
```diff
@@ -0,0 +1,176 @@
...
+from copy import deepcopy
+from dataclasses import dataclass
+from typing import Optional
+
+__all__ = ["QuantizationScheme", "preset_name_to_scheme"]
+
+
+@dataclass
+class QuantizationScheme:
+    bits: int = 4
```
hshen14, line 24: shall we add wei_ as prefix?

auto_round/schemes.py
```diff
@@ -0,0 +1,176 @@
...
+    group_size: int = 128
+    sym: bool = True
+    data_type: str = "int"
+    act_bits: Optional[int] = None
+    act_group_size: Optional[int] = None
+    act_sym: Optional[bool] = None
+    act_data_type: Optional[str] = None
+    act_dynamic: Optional[bool] = None
+    super_bits: Optional[int] = None
```
hshen14, line 33: what is super_bits and how is used?

auto_round/schemes.py
```diff
@@ -0,0 +1,176 @@
...
+    }
+)
+
+MXFP8 = QuantizationScheme.from_dict(
+    {
+        "bits": 8,
+        "group_size": 32,
+        "data_type": "mx_fp",
+        "act_bits": 8,
+        "act_data_type": "mx_fp_rceil",
```
hshen14, line 113: rceil is rounding/scaling, which should be separate from data type?

auto_round/autoround.py
```diff
@@ -107,44 +109,43 @@ class AutoRound(object):
...
         tokenizer=None,
-        bits: int = 4,
-        group_size: int = 128,
-        sym: bool = True,
-        layer_config: dict = None,
-        batch_size: int = 8,
-        amp: bool = True,
-        device: Union[str, torch.device, int] = 0,
+        scheme: Union[str, dict, QuantizationScheme] = "W4A16",
+        layer_config: Union[dict, QuantizationScheme] = None,
```
n1ck-guo: Need ut for layer_config is QuantizationScheme, it seems to cause some bugs

auto_round/schemes.py
```diff
@@ -0,0 +1,176 @@
...
+    }
+)
+
+NVFP4 = QuantizationScheme.from_dict(
+    {
+        "bits": 4,
+        "group_size": 16,
+        "data_type": "nv_fp",
+        "act_bits": 4,
+        "act_data_type": "nv_fp_with_static_gs",
```
hshen14, line 123: similar as above. static_gs does not belong to data type.

how about adding an act_scheme field? it can be static_gs, rceil, etc

wenhuach21: Theoretically, you are correct. However, each dtype has its own scheme, for example, FP8 may use  unit scale. As the number of supported dtypes grows, the number of keys in act_scheme also increases. Therefore, combining certain schemes with the data_type is a more practical approach for coding.

auto_round/schemes.py
wenhuach21: same issue as below, mark duplicate

auto_round/schemes.py
wenhuach21: This is for gguf double quantization or NF4, bits for scale and zp

auto_round/autoround.py
n1ck-guo: Does layer_config support multiple types? For example:
layer_config = {
layer1: {bits: 16},
layer2: "W4A16",
layer3: QuantizationScheme.from_dirc({})
}

auto_round/schemes.py
wenhuach21: Yes, weight_bits is more clear. But for backward compatibility, I’d prefer to keep it for now.

auto_round/autoround.py
wenhuach21: nice catch, it has bug for now.  I will fix it.

auto_round/autoround.py
```diff
@@ -1099,20 +1135,26 @@ def _check_need_to_quantize_lm_head_embedding(self) -> bool:
...
+            formats: list[str] = [fmt for fmt in self.formats if "fake" not in fmt]
+            if not (len(formats) == 1 and "gguf" in formats[0]):
+                raise NotImplementedError("Only one GGUF format can be set at a time.")
+            target_format: str = formats[0]
 
-        formats: list[str] = [fmt for fmt in self.formats if "fake" not in fmt]
-        if not (len(formats) == 1 and "gguf" in formats[0]):
-            raise NotImplementedError("Only one GGUF format can be set at a time.")
+        else:
+            target_format = self.scheme.lower()
```
n1ck-guo, line 1202: self.scheme may be None

wenhuach21: if not hasattr(self, "formats") and not gguf_scheme:
            return False

wenhuach21:
@n1ck-guo  could you have a test for vlms, specially for large model and lm-head quantization

WeiweiZhang1 (approved):

PR #778, opened 2025-09-01
gguf format supports for fp8 model

n1ck-guo:


auto_round/export/export_to_gguf/convert.py
```diff
@@ -144,6 +144,13 @@ def get_tensors(cls) -> Iterator[tuple[str, Tensor]]:
                     break
         yield name, tensor
 
+    def is_extra_tensor(tensor_name):
+        if hasattr(cls.model, "is_fp8") and "scale" in tensor_name.split(".")[-1]:
```
wenhuach21: could you  wrap "is_fp8" as a function? as it is too tricky, we may change it later.

wenhuach21: and change to  this function calling for the similar codes in other places

wenhuach21 (approved):

PR #794, opened 2025-09-05
fix bug and add nvfp in alg-ext with slight improvement

wenhuach21:


auto_round/autoround.py
```diff
@@ -2815,7 +2819,7 @@ def _quantize_blocks(
                 quantize_block = self.quantize_block_ext  # must use self.quantize_block_ext
                 if self.bits > 2 and not self.data_type.startswith("mx"):
                     logger.warning(
-                        "algorithm extension has only undergone limited validation on INT2 and mxfp4; use with caution."
+                        "algorithm extension has only undergone limited validation on INT2,mxfp4 and nvfp4; use with caution."
```
thuang6: since we try to give warning msg for using mxfp4,  why do we check data type is not started with "mx"?

wenhuach21: The intention is to suppress warnings in already validated scenarios, including int2, mxfp4, and nvfp4.

auto_round/autoround.py
```diff
@@ -2806,16 +2806,21 @@ def _quantize_blocks(
             self.sym
             and self.enable_alg_ext
             and self.super_group_size is None
-            and ((self.data_type.startswith("int") and self.act_bits >= 8) or self.data_type.startswith("mx"))
+            and (
+                (self.data_type.startswith("int") and self.act_bits >= 8)
+                or self.data_type.startswith("mx")
+                or self.data_type.startswith("nv")
```
WeiweiZhang1, line 2812: how about is_nv_fp(self.data_type)

WeiweiZhang1 (approved):

PR #803, opened 2025-09-09
improve the robustness of scheme

wenhuach21:


auto_round/schemes.py
```diff
@@ -111,6 +112,7 @@ def is_preset_scheme(name: str) -> bool:
         "data_type": "mx_fp",
         "act_bits": 8,
         "act_data_type": "mx_fp_rceil",
+        "act_group_size": 32,
```
WeiweiZhang1, line 115: sym=True is also needed for mxfp/nvfp

wenhuach21: Why? Sym vs. asym doesn’t change the quantization part. Does it matter for packing?

PR #805, opened 2025-09-09
fix cuda ut bug of use_deterministic_algorithms

n1ck-guo:


auto_round/script/llm.py
```diff
@@ -543,6 +546,12 @@ def tune(args):
     scheme = args.scheme.upper()
     if scheme not in PRESET_SCHEMES:
         raise ValueError(f"{scheme} is not supported. only {PRESET_SCHEMES.keys()} are supported ")
+    if args.disable_deterministic_algorithms:
+        logger.warning(
+            "default not use deterministic_algorithms. disable_deterministic_algorithms is deprecated,"
+            " please use enable_deterministic_algorithms instead. "
+        )
+    enable_deterministic_algorithms = args.enable_deterministic_algorithms and not args.disable_deterministic_algorithms
```
wenhuach21: use enable_deterministic_algorithms only , it is quite confusing that using them both

auto_round/script/llm.py
```diff
@@ -208,6 +208,9 @@ def __init__(self, *args, **kwargs):
         self.add_argument(
             "--disable_deterministic_algorithms", action="store_true", help="disable torch deterministic algorithms."
```
wenhuach21: add deprecated in help

auto_round/autoround.py
```diff
@@ -235,7 +236,13 @@ def __init__(
         if kwargs:
             logger.warning(f"unrecognized keys {list(kwargs.keys())} were passed. Please check them.")
 
-        if not disable_deterministic_algorithms:
+        # deprecated, default not to use torch.use_deterministic_algorithms
```
wenhuach21, line 240: upper case

auto_round/autoround.py
```diff
@@ -235,7 +236,13 @@ def __init__(
...
+        if not disable_deterministic_algorithms or enable_deterministic_algorithms:
+            if not disable_deterministic_algorithms:
+                logger.warning(
+                    "default not use deterministic_algorithms. disable_deterministic_algorithms is deprecated,"
+                    " please use enable_deterministic_algorithms instead. "
+                )
             if "CUBLAS_WORKSPACE_CONFIG" not in os.environ:
                 os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
             torch.use_deterministic_algorithms(True, warn_only=False)
```
wenhuach21, line 248: if enable_deterministic_algorithms is not set, we use warn_only=Ture, which is the default behavior in our previous version, though I admit it's a quite confusing.

PR #815, opened 2025-09-11
enhance auto-round eval with vllm backend

xin3he:
It was tested on HPU and CUDA. GGUF is not supported since vllm doesn't allow passing model object, only model string is allowed.

Command: `VLLM_WORKER_MULTIPROC_METHOD=spawn VLLM_SKIP_WARMUP=true auto-round --model facebook/opt-125m  --eval --vllm --tasks lambada_openai --limit 10`

### DIFF
--- a/auto_round/script/llm.py
+++ b/auto_round/script/llm.py
@@ -341,7 +341,6 @@ class EvalArgumentParser(argparse.ArgumentParser):
         self.add_argument("--max_batch_size", default=None, type=int, help="max batch size for vllm")
         self.add_argument("--max_length", default=None, type=int, help="max generation length for vllm")
         self.add_argument("--max_model_len", default=None, type=int, help="maximum model sequence length for vllm")
-        self.add_argument("--seed", default=1234, type=int, help="random seed")
         self.add_argument(
             "--gpu_memory_utilization", default=0.9, type=float, help="target GPU memory utilization for vllm"
         )


auto_round/script/llm.py
```diff
@@ -311,9 +313,7 @@ def __init__(self, *args, **kwargs):
         )
         self.add_argument("--eval_bs", "--bs", "--batch_size", default=None, type=int, help="batch size in evaluation")
         self.add_argument("--eval_task_by_task", action="store_true", help="whether to eval task by task.")
-        self.add_argument(
-            "--eval_model_dtype", default=None, type=str, help="the torch_dytpe to load the model for evaluation."
-        )
+        self.add_argument("--dtype", default="auto", type=str, help="the dtype to load the model for evaluation.")
```
WeiweiZhang1: please do not change the argument name eval_model_dtype, since using dtype could be confused with the quantization process data type

xin3he: done

auto_round/script/llm.py
```diff
@@ -322,6 +324,27 @@ def __init__(self, *args, **kwargs):
             help="Limit the number of examples per task. "
             "If <1, limit is a percentage of the total number of examples.",
         )
+        # vllm related arguments
```
yiliu30, line 327: How can we better manage the numerous vLLM arguments in AutoRound to accommodate future changes? Could we explore a cleaner way to propagate them?

auto_round/script/llm.py
```diff
@@ -949,3 +973,51 @@ def eval_task_by_task(
...
+    from lm_eval import evaluator  # pylint: disable=E0401
+    from lm_eval.models.vllm_causallms import VLLM  # pylint: disable=E0401
+    from lm_eval.utils import make_table  # pylint: disable=E0401
+
+    device_str, _ = get_device_and_parallelism(args.device)
+    eval_model_dtype = get_model_dtype(args.eval_model_dtype, "auto")
+    if (batch_size := args.eval_bs) is None:
+        batch_size = "auto:8"
+
+    vllm_lm = VLLM(
```
yiliu30, line 992: I assume this backend is faster than the HF backend. Do we pass the HF model directly to vLLM? Could you explain the philosophy behind this approach? Thanks!

docs/step_by_step.md
```diff
@@ -617,6 +617,8 @@ If not explicitly specify '--task', the default value will be used (typically co
   ~~~
   The last format will be used in evaluation if multiple formats have been exported.
 
+Note: To use the vllm backend, please add `--vllm` into the upper command.
+
```
yiliu30, line 621: Could you please add the usage and limitations to the doc?

auto_round/script/llm.py
xin3he: Here I collected all arguments in VLLM class in lm_eval, and we can expect contiguously development to suit future changes.
If you saw any solution that can handle future changes better, let me know it. Thanks.

auto_round/script/llm.py
xin3he: This class comes from lm_eval which only supports model string that can be loaded by vLLM itself.

docs/step_by_step.md
xin3he: The usage is quite direct, just append the `--vllm` as I noted.
The limitation is not limited by us, it's from lm_eval itself. It's not suggested to maintain the limitation here since all argument will be passed to lm_eval and we need to update the limitation once it upgrade. Not a good solution to maintain I think

PR #820, opened 2025-09-11
Fix llm-compressor export

Kaihui-intel:
add inplace & return model for llmcompressor export

### DIFF
--- a/auto_round/export/export_to_llmcompressor/export.py
+++ b/auto_round/export/export_to_llmcompressor/export.py
@@ -80,9 +80,9 @@ def save_quantized_as_llmcompressor(output_dir: str, inplace: bool = True, **kwa
     """
     Save a quantized model in the LLM-Compressor format.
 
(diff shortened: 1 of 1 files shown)


auto_round/export/export_to_llmcompressor/export.py
```diff
@@ -75,7 +76,31 @@ def pack_layer(layer_name, model, backend, device=None):
...
+            - safe_serialization (bool): Whether to use safe serialization when saving
+                                         the model. Default is True.
+
+    Returns:
+        torch.nn.Module: The quantized model that was saved.
+    """
+
     backend = kwargs.get("backend", None)
     if is_nv_fp(backend) or is_mx_fp(backend):
         return save_quantized_as_fp(output_dir, **kwargs)
```
WeiweiZhang1: Previously, inplace was included within kwargs. Now it needs to be explicitly passed through to other save sub-functions. Please help align it accordingly.

PR #825, opened 2025-09-16
enable dynamic quantization config saving

WeiweiZhang1:


auto_round/autoround.py
```diff
@@ -1916,6 +1916,7 @@ def _set_layerwise_config(self, layer_config: dict) -> bool:
         """
         # Get the names of layers in quantization blocks
         supported_types = self.supported_types
+        dynamic_config = {}
```
wenhuach21: I suggest avoiding the term 'dynamic config,' which is not used in academic literature and was coined by Unlosh, and personally, I think it's a little confusing, as there are static activation quantization and dynamic activation quantization

wenhuach21:
TODO:
1 Supported the inference for AutoRound  format in Transformers
2 Validated the inference for AutoRound  format in vLLMs/Sglangs
3 ADD UTs

auto_round/export/export_to_awq/export.py
```diff
@@ -137,6 +137,7 @@ def wrapper(name):
         return model
 
     quantization_config = kwargs["serialization_dict"]
+    quantization_config.pop("regex_config") #as awq do not support mixed bits config saving
```
wenhuach21: if its 16 bits, we could convert it to not_convert_module, I forget the name

test/test_cuda/test_mix_bits.py
```diff
@@ -58,17 +57,16 @@ def test_mixed_gptqmodel(self):
...
         autoround.quantize_and_save(output_dir=quantized_model_path, format="auto_gptq")
         from gptqmodel import GPTQModel
-
         model = GPTQModel.load(quantized_model_path)
-        assert model.model.model.decoder.layers[0].self_attn.k_proj.bits == 8
-        assert model.model.model.decoder.layers[0].self_attn.q_proj.bits == 4
-        result = model.generate("Uncovering deep insights begins with")[0]  # tokens
-        assert "!!!" not in model.tokenizer.decode(result)  # string output
+        assert (model.model.model.decoder.layers[0].self_attn.k_proj.bits == 8)
+        assert (model.model.model.decoder.layers[0].self_attn.q_proj.bits == 4)
```
wenhuach21: test the regex , full name, part name and make sure the model could inference correctly.

test/test_cuda/test_mix_bits.py
```diff
@@ -181,5 +172,7 @@ def test_mixed_llmcompressor_format_vllm(self):
         shutil.rmtree(quantized_model_path, ignore_errors=True)
 
 
+
 if __name__ == "__main__":
     unittest.main()
```
wenhuach21, line 329: For AutoRound format, please make sure inference is ready first then support it on exporting side

wenhuach21:
Have you validated the GPTQ format? Even without regex in the layer_config , the configuration should still be set under the dynamic key in some way.

WeiweiZhang1:
> Have you validated the GPTQ format? Even without regex in the layer_config , the configuration should still be set under the dynamic key in some way.

sure, also added UT for auto_gptq format infer & convert to autoround format infer

wenhuach21:
Qwen3-VL-30B-A3B-Instruc or smaller model is  ok. And transformers is enough, no need to verify on vllm ,as vllm should not be able to support our converted model.

PR #834, opened 2025-09-19
dump avg_bits

xin3he:
Dump the average bits info at the beginning before AutoRound tuning.

```bash
2025-09-19 06:43:37 INFO autoround.py L1684: The target average bits of blocks in the model (without lm_head): 4.250 bits
2025-09-19 06:43:37 INFO autoround.py L1685: The target average bits of the entire model (with lm_head): 8.000 bits
```


auto_round/utils.py
```diff
@@ -2730,3 +2730,48 @@ def is_mllm_model(model_or_path: Union[str, torch.nn.Module]):
...
+        with_lm_head = False
+    for n, m in module.named_modules():
+        if n == lm_head_name and not with_lm_head:
+            continue
+        if isinstance(m, SUPPORTED_LAYER_TYPES):
+            m_numel = m.weight.numel()
+            all_numel += m_numel
+            w_bits = m.bits * m_numel
+            all_bits += w_bits
+            if m.data_type in ("fp4_v2", "nv_fp", "mx_fp", "nv_fp4", "mx_fp4", "mx_fp8"):
```
yiliu30: Some supported layers may be skipped by the user; please handle this case as well.

yiliu30: It would be better to define a constant list for these dtypes.

xin3he: Agreed

xin3he: Sorry, I don't understand. 
If layers are skipped, It will definitely increase the average bits.

auto_round/compressors/base.py
```diff
@@ -1690,6 +1691,15 @@ def quantize(self) -> tuple[torch.nn.Module, dict[str, Any]]:
         # because it may cause the gguf format to not be exported normally.
         self.model = _handle_moe_model(self.model, formats=formats)
         self.has_qlayer_outside_block = self._set_layerwise_config(self.layer_config)
+        average_bits = get_avg_bits(self.model)
+        average_bits_w_lm_head = get_avg_bits(self.model, with_lm_head=True)
+        if average_bits_w_lm_head != average_bits:
+            logger.info(f"The target average bits of blocks in the model (without lm_head): {average_bits:.3f} bits")
```
wenhuach21: No need to expose too many details, as users may not understand them. Please ensure the meaning of target_bits is consistent with automatic mixed bits. I suggest checking the target_bits setting in TRT, perhaps the scale or zero-point count is not being included.

wenhuach21: https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/74061f5bc91f880fa1e6abb339906834db1c54ab/examples/llm_autodeploy/run_auto_quantize.py#L71C21-L71C39

auto_round/utils.py
yiliu30: The skipped layers don’t have `data_type`, right?
Please add a check to handle this case?

auto_round/compressors/base.py
xin3he: It's fine to put one of them into debug mode.
image
upper image about TRT comes from copilot, I think it's aligned with TRT.

auto_round/utils.py
xin3he: I see, I can check, thanks.

n1ck-guo: @yiliu30 skipped layers also have data_type

auto_round/utils.py
```diff
@@ -2755,3 +2755,59 @@ def is_mllm_model(model_or_path: Union[str, torch.nn.Module]):
...
+            return 1
+        elif group_size == -1:
+            return input_features
+        else:
+            return weight_numel // group_size
+
+    all_numel = 0
+    all_bits = 0
+
+    lm_head_name = get_lm_head_name(module)
```
wenhuach21, line 2792: better provide an extra function for auto scheme,  get_bits_for_layer(layer, scheme: Union(str, dict, quantizations_cheme))

auto_round/utils.py
```diff
@@ -2755,3 +2755,59 @@ def is_mllm_model(model_or_path: Union[str, torch.nn.Module]):
...
+    if lm_head_name is None:
+        with_lm_head = False
+    for n, m in module.named_modules():
+        if n == lm_head_name and not with_lm_head:
+            continue
+        if isinstance(m, SUPPORTED_LAYER_TYPES):
```
n1ck-guo: change to type(m) in SUPPORTED_LAYER_TYPES，isinstance is inaccurate in some scenarios

yiliu30 (approved):

PR #859, opened 2025-09-28
add missing batch_size consideration

xin3he:
Batch_size is not considered before, quantizing llama3.3 70b will only use 1 card and got OOM.


xin3he:
Okay, I will share the reproduce command to you so that we can find out a correct solution.

PR #871, opened 2025-10-09
fix seqlen bug and calib slow of mllm tuning

n1ck-guo:


auto_round/compressors/mllm/compressor.py
```diff
@@ -224,6 +230,11 @@ def __init__(
             if dataset in MLLM_DATASET.keys():
                 truncation = False
                 seqlen = 512 if seqlen is None else seqlen
+                if seqlen > 512:
```
wenhuach21: As we may support other datasets  in the futhure, we'd better refine the code.
option 1: move this checker to the specific dataset
option 2: check the dataset name in line 233 as well
option 1 is better

auto_round/compressors/mllm/compressor.py
```diff
@@ -270,6 +281,76 @@ def __init__(
...
+            layer_names (list, optional): List of layer names to cache data for. Defaults to [].
+            last_cache_name (str, optional): Name of the last cache. Defaults to None.
+
+        Returns:
+            all_inputs: Cached intermediate data.
+
+        Raises:
+            Exception: If caching on GPU fails, switches to CPU and caches there.
+        """
+        if _is_fp8_model(self.model):
```
wenhuach21: what's the differences with the original function

n1ck-guo: mllm better to try cuda calib first. If they are mixed together, it may be troublesome to refactor the code later, so I separate them.

auto_round/compressors/mllm/dataset.py
```diff
@@ -233,6 +234,12 @@ def get_mllm_dataloader(
         )
 
     if os.path.isfile(dataset) or dataset in MLLM_DATASET.keys():
+        if seqlen > MLLM_DATASET[dataset].MAX_SUPPORT_SEQLEN:
+            logger.warning(
+                f"seqlen({seqlen}) is greater than the maximum length supported by the {dataset},"
+                f" reset to {MLLM_DATASET[dataset].MAX_SUPPORT_SEQLEN}"
+            )
+            seqlen = 512
```
wenhuach21, line 242: shall we return the seqlen to  auto_round object thus it could change its seqlen member to avoid any potential bug?

auto_round/compressors/mllm/compressor.py
wenhuach21: The original function includes several checks to determine whether to use the CPU first https://github.com/intel/auto-round/blob/main/auto_round/compressors/base.py#L2224. 

How about adding another condition to verify that the current block_name is not the first one, such as the one near the embedding layer
 or `self.mllm` to force it  try gpu first

PR #875, opened 2025-10-09
fix save_quantized format checker for nvfp4

WeiweiZhang1:
backport https://github.com/intel/auto-round/pull/871 into v0.8.0rc

### DIFF
--- a/test/test_cpu/test_export.py
+++ b/test/test_cpu/test_export.py
@@ -504,7 +504,6 @@ class TestAutoRound(unittest.TestCase):
         ), "Illegal NVFP4 packing name or data_type or shape"
         shutil.rmtree("./saved", ignore_errors=True)
 
-
     def test_nvfp4_autoround_save_quantized(self):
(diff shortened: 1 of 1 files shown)


auto_round/compressors/base.py
```diff
@@ -967,7 +967,7 @@ def _check_supported_format(self, format: str) -> bool:
                         f" act_bits={self.act_bits}"
                     )
                 elif "auto_round" in format and (
-                    is_mx_fp(self.act_data_type) or (is_nv_fp(format) and "static_gs" in self.act_data_type)
+                    is_mx_fp(self.act_data_type) or (is_nv_fp(self.act_data_type) and "static_gs" in self.act_data_type)
```
wenhuach21, line 970: why not checking both of data_type and act_type

WeiweiZhang1: The check happens when act_bits <= 8. If needed, a data_type checker can be added in a separate PR targeting main.

PR #880, opened 2025-10-10
Fix act calibration bug

mengniwang95:
Bug fix


test/test_cpu/test_autoround.py
```diff
@@ -49,6 +49,22 @@ def test_bits_setting(self):
         if module.bits != 8:
             raise ValueError(f"Expected bits to be 8, but got {module.bits}")
 
+    def test_layer_config(self):
+        layer_config = {
+            "self_attn": {"bits": 4, "data_type": "mx_fp", "act_bits": 4, "act_data_type": "mx_fp", "group_size": 32}
+        }
+        autoround = AutoRound(
+            self.model,
```
wenhuach21: please just pass the self.model_name in the future, better not use the same model

test/test_cpu/test_autoround.py
```diff
@@ -49,6 +49,22 @@ def test_bits_setting(self):
...
+            self.tokenizer,
+            scheme="NVFP4",
+            iters=0,
+            seqlen=2,
+            dataset=self.llm_dataloader,
+            layer_config=layer_config,
+            amp=False,
+        )
+        autoround.quantize_and_save(self.save_folder, inplace=False)
```
wenhuach21: if call save, then delete the folder after the ut

test/test_cpu/test_autoround.py
mengniwang95: updated

test/test_cpu/test_autoround.py
mengniwang95: done

PR #887, opened 2025-10-13
Add GPT-OSS quant support

yiliu30:
.

### DIFF
--- a/requirements-lib.txt
+++ b/requirements-lib.txt
@@ -1,4 +1,4 @@
-accelerate <= 1.5.1
+accelerate>=1.10.0
 datasets
 py-cpuinfo
 sentencepiece
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
-accelerate<=1.5.1
+# accelerate<=1.5.1
+accelerate>=1.10.0
 datasets
 py-cpuinfo
 sentencepiece


auto_round/special_model_handler.py
```diff
@@ -36,71 +37,24 @@
 }
 SPECIAL_SHARED_CACHE_KEYS["MiniMaxText01ForCausalLM"] = ("slope_rate",)
 
-CONVERT_EXPERT_TO_LINEAR_MODELS = ["llama4"]
+CONVERT_EXPERT_TO_LINEAR_MODELS = ["llama4", "gpt_oss"]
```
wenhuach21, line 40: It would be better not to categorize it into too many detailed types. A single flag like model_need_to_convert, or a similar name, should be sufficient, since some models may require conversion even if they don’t have expert layers. We provide a converter function for each model if needed, regardless of which parts need to be converted.

yiliu30: Yeah, I agree, the replacement code could be organized better. Once we support more model replacements, we can refactor that part as needed. For now, how about leaving it as is, since we have some higher-priority tasks to focus on?

wenhuach21: I don’t think it will take much effort to change. You could also finish the higher-priority tasks first.

yiliu30: Opened an issue to track https://github.com/intel/auto-round/issues/899

PR #891, opened 2025-10-13
fix nvfp act quantization bug

WeiweiZhang1:


test/test_cuda/test_export.py
```diff
@@ -402,6 +402,42 @@ def test_nvfp4_llmcompressor_format(self):
...
+        autoround = AutoRound(
+            model,
+            tokenizer,
+            scheme=scheme,
+            iters=0,
+            seqlen=2,
+            nsamples=2,
+            dataset=self.llm_dataloader,
+        )
+        autoround.quantize()
```
wenhuach21, line 416: add cpu ut as well, no need save in cpu ut

test/test_cuda/test_export.py
```diff
@@ -402,6 +402,42 @@ def test_nvfp4_llmcompressor_format(self):
         #     if "France" in prompt:
         #         assert "Paris" in generated_text
 
+    def test_nvfp4_moe_actmax_rtn(self):
+        model_name = "/data0/deepseek-ai/DeepSeek-V2-Lite"
+        model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype="auto", trust_remote_code=True)
```
wenhuach21: why still use the legacy way? just pass the model name is fine

test/test_cuda/test_export.py
```diff
@@ -402,6 +402,42 @@ def test_nvfp4_llmcompressor_format(self):
...
+            nsamples=2,
+            dataset=self.llm_dataloader,
+        )
+        autoround.quantize()
+        quantized_model_path = self.save_dir
+        autoround.save_quantized(output_dir=quantized_model_path, inplace=False, format="auto_round")
+
+    def test_nvfp4_moe_actmax_ar(self):
+        model_name = "/data0/deepseek-ai/DeepSeek-V2-Lite"
+        model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype="auto", trust_remote_code=True)
```
wenhuach21: better not use trust_remote_code if it is not required in the future

test/test_cuda/test_export.py
WeiweiZhang1: > add cpu ut as well, no need save in cpu ut

The loading and quantization of this model(30G) are heavy for CPU testing

wenhuach21: then use a small model

test/test_cpu/test_export.py
```diff
@@ -558,9 +558,7 @@ def test_nvfp4_moe_actmax_ar(self):
             nsamples=2,
             dataset=self.llm_dataloader,
         )
-        autoround.quantize()
-        quantized_model_path = self.save_dir
-        autoround.save_quantized(output_dir=quantized_model_path, inplace=False, format="auto_round")
+        autoround.quantize_and_save(output_dir=self.save_dir, inplace=True, format="auto_round")
```
wenhuach21, line 572: change one ut to quantize, as only quantize could expose the issue you are triyng to fix

PR #896, opened 2025-10-14
fix llama 4 ut

n1ck-guo:


test/test_cuda/test_gguf.py
```diff
@@ -252,7 +252,7 @@ def test_llama_4(self):
         )
         self.assertAlmostEqual(file_size, 58093.62, delta=1.0)
         file_size = os.path.getsize(os.path.join(quantized_model_path, "mmproj-model.gguf")) / 1024**2
-        self.assertAlmostEqual(file_size, 3323.52, delta=1.0)
+        self.assertAlmostEqual(file_size, 3326.18, delta=1.0)
```
wenhuach21: Why did this value change? If the reason is the GGUF version, we’d better provide a range instead of an exact value.

n1ck-guo: Yes, because of version update

PR #900, opened 2025-10-15
gguf weight type align with original, output.weight, token_embed

n1ck-guo:


auto_round/utils.py
```diff
@@ -1999,7 +1999,7 @@ def _set_config(config, target_config):
             elif new_type != "gguf:q8_0":
                 new_type = "gguf:q6_k"
         elif lm_head_name is not None and layer_name == lm_head_name and tie_word_embeddings:
-            pass
+            new_type = GGUF_CONFIG[target_gguf_format]["lm_head"]
```
wenhuach21: add ut to check bits
add another ut to check embedding and lm-head bits if customized layer_config in API is provided

auto_round/compressors/base.py
```diff
@@ -1333,7 +1333,11 @@ def _check_need_to_quantize_lm_head_embedding(self) -> bool:
         tie_word_embeddings: bool = getattr(getattr(self.model, "config", None), "tie_word_embeddings", True)
         for name, module in self.model.named_modules():
             if isinstance(module, torch.nn.Embedding):
-                key: str = "lm_head" if tie_word_embeddings else "embedding"
+                if tie_word_embeddings:
+                    config: dict[str, Any] = GGUF_INNER_CONFIG[GGUF_CONFIG[target_format]["lm_head"]]
+                    self._apply_config_to_layer("lm_head", config, False)
+                # key: str = "lm_head" if tie_word_embeddings else "embedding"
+                key: str = "embedding"
                 config: dict[str, Any] = GGUF_INNER_CONFIG[GGUF_CONFIG[target_format][key]]
```
wenhuach21: check more small models, I remember if tiled word embedding is True then the embedding layer is q6k in most scenario.
Qwen3-0.6B is q6k.
check opt-125m, llama-3B,etc

auto_round/compressors/base.py
```diff
@@ -502,65 +474,102 @@ def _set_device(self, device_map: Union[str, torch.device, int, dict]) -> None:
...
-        self.act_bits = self.act_bits if self.act_bits is not None else 16
-        self.act_sym = self.act_sym if self.act_sym is not None else self.sym
+        def _parse_and_set(scheme, kwargs):
+            res = ""
+            if isinstance(scheme, QuantizationScheme):
+                scheme = asdict(scheme)
+            elif isinstance(scheme, dict):
+                scheme = scheme
+            elif isinstance(scheme, str):
+                res = scheme  # gguf:q4_k_s and gguf_q4_k_m has the same dict scheme, but the result is different
```
wenhuach21: change the comment to
We’d better keep the string scheme instead of the dict config, since GGUF uses different mixed-bit strategies for q4_k_s and q4_k_m even though they share the same scheme dict.

auto_round/utils.py
```diff
@@ -3042,3 +3044,27 @@ def is_diffusion_model(model_or_path: Union[str, object]) -> bool:
         return isinstance(model_or_path, pipeline_utils.DiffusionPipeline)
     else:
         return False
+
+
+def is_separate_lm_head(model: torch.nn.Module) -> bool:
+    dir_path = model.name_or_path
+    if not os.path.isdir(dir_path):
+        dir_path = download_hf_model(dir_path)
```
wenhuach21, line 3032: Better not download the model again. Since returning a wrong answer here doesn’t cause major issues, we can simply return a placeholder answer instead.

auto_round/utils.py
```diff
@@ -3042,3 +3044,27 @@ def is_diffusion_model(model_or_path: Union[str, object]) -> bool:
         return isinstance(model_or_path, pipeline_utils.DiffusionPipeline)
     else:
         return False
+
+
+def is_separate_lm_head(model: torch.nn.Module) -> bool:
+    dir_path = model.name_or_path
```
wenhuach21, line 3030: dose each model have a name_or_path?

n1ck-guo: yes this is an arg of hf

PR #901, opened 2025-10-15
add UT for regression

xin3he:


test/test_cuda/test_transformers.py
```diff
@@ -204,6 +206,45 @@ def test_mixed_bits(self):
             inputs = tokenizer(text, return_tensors="pt").to(model.device)
             tokenizer.decode(model.generate(**inputs, max_new_tokens=5)[0])
 
+    def test_mx(self):
```
wenhuach21, line 209: this file is only for testing same uts  in huggingface transformers, need to move the new ut to some other place

xin3he: I will close it since weiwei had fixed this issue and added UT in her commits

PR #930, opened 2025-10-22
support attention mask in user's dataset

wenhuach21:


wenhuach21:
This PR is not very robust since the `attention_mask `is not incorporated into the forward pass since different models may handle it differently. Therefore, we should warn users when their input data includes padding. As a workaround, they can use a batch size of 1 instead.

WeiweiZhang1 (approved):

PR #936, opened 2025-10-23
Fix Flux tuning issue

mengniwang95:
In case the input of Flux is dict, need to fix the way of getting nsamples from input_ids for Flux

### DIFF
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -2646,7 +2646,7 @@ class BaseCompressor(object):
         else:
             lr_schedule = copy.deepcopy(self.lr_scheduler)
 
-        if isinstance(input_ids, dict): # input_ids of Flux is dict
+        if isinstance(input_ids, dict):  # input_ids of Flux is dict
             nsamples = len(input_ids["hidden_states"])
         else:
             nsamples = len(input_ids)


auto_round/compressors/base.py
```diff
@@ -2365,7 +2369,12 @@ def _quantize_layer(
             )
         else:
             lr_schedule = copy.deepcopy(self.lr_scheduler)
-        nsamples = len(inputs)
+
+        if isinstance(input_ids, dict):
+            nsamples = len(input_ids["hidden_states"])
```
wenhuach21: in LLM, inputs_id = hidden_states, and move other keys to input_others. Is it possible to keep this style? The input_others are assumed to remain unchanged across blocks.

auto_round/compressors/base.py
```diff
@@ -2306,14 +2306,18 @@ def _replace_forward(self):
...
+        q_inputs: Union[torch.Tensor, dict] = None,
+        device: str = "cpu",
     ):
         """Quantize a specific layer of the model using the provided inputs.
 
         Args:
             layer_name (str): The name of the layer to quantize.
-            inputs (torch.Tensor): Input data for quantization.
-            q_inputs (torch.Tensor, optional): Quantized input data. Defaults to None.
+            inputs (torch.Tensor | dict ): Input data for quantization.
```
wenhuach21: why linear layer still need a dict inputs?

mengniwang95: thx, it is not required. Have been removed

auto_round/compressors/base.py
mengniwang95: For Flux, there are hidden_states and encoder_hidden_states that needs update for next block tuning, so https://github.com/intel/auto-round/blob/873114a0ff5ea7496918ac71df3899122b8e02ca/auto_round/compressors/base.py#L2850 q_inputs and inputs should contain hidden_states and encoder_hidden_states.
If we want to keep style and reuse the _quantize_block between LLM and diffusion , the solution I can think of is adding a _get_nsamples function, base.py and diffusion compressor use different implementation.
Do you have other better suggestions?

auto_round/compressors/base.py
```diff
@@ -2646,7 +2646,11 @@ def _quantize_block(
         else:
             lr_schedule = copy.deepcopy(self.lr_scheduler)
 
-        nsamples = len(input_ids)
+        if isinstance(input_ids, dict):
```
wenhuach21: add comments this is for flux

auto_round/compressors/base.py
wenhuach21: ok, then keeps your style

auto_round/compressors/base.py
```diff
@@ -2527,9 +2527,9 @@ def _get_current_num_elm(
     def _quantize_block(
         self,
         block: torch.nn.Module,
-        input_ids: list[torch.Tensor],
+        input_ids: Union[list[torch.Tensor], dict],
         input_others: dict,
-        q_input: Union[None, torch.Tensor] = None,
+        q_input: Union[torch.Tensor, dict] = None,
```
wenhuach21: should include None in Union or use optional

wenhuach21 (approved):

PR #940, opened 2025-10-23
gguf support for inclusionAI/Ling-flash-2.0

n1ck-guo:


auto_round/export/export_to_gguf/export.py
```diff
@@ -73,7 +74,9 @@ def create_model_class(
     low_cpu_mem_usage=False,
     model_type=convert_hf_to_gguf.ModelType.TEXT,
 ):
-    tmp_work_dir = Path(os.path.join(output_dir, TMP_DIR_NAME))
+    tmp_work_dir = model.name_or_path
+    if not os.path.isdir(tmp_work_dir):
+        tmp_work_dir = download_hf_model(tmp_work_dir)
```
wenhuach21, line 80: In which cases do we need to redownload the model? It’s a bit confusing.

auto_round/export/export_to_gguf/export.py
```diff
@@ -95,7 +98,7 @@ def create_model_class(
         output_type = FTYPE_MAP.get(output_type.lower())
 
         model_instance = model_class(
-            dir_model=tmp_work_dir,
+            dir_model=Path(tmp_work_dir),
```
wenhuach21, line 102: Could we change this function to support module input in other pr?

auto_round/export/export_to_gguf/export.py
n1ck-guo: download_hf_model is not a redownload func, in fact, if the model is already download, it will returen the save_path

auto_round/export/export_to_gguf/export.py
n1ck-guo: Wil try in other pr

wenhuach21: no need if the model is not redownloaded

PR #943, opened 2025-10-24
refactor utils file

n1ck-guo:


auto_round/utils/constants.py
```diff
@@ -0,0 +1,82 @@
...
+
+def compare_versions(v1, v2):
+    return version.parse(v1) >= version.parse(v2)
+
+
+def torch_version_at_least(version_string):
+    return compare_versions(torch.__version__, version_string)
+
+
+TORCH_VERSION_AT_LEAST_2_6_PRE_RELEASE = torch_version_at_least("2.5.99")
```
wenhuach21: too many utils files. Could we move some to folders, like compressor, export and leave the common ones to utils.py ?

n1ck-guo: I will try to reduce the number of utils files, but I think it is necessary to split it into multiple files because there are still many common utils. Split it into multiple files will improves readability and eases maintenance

auto_round/utils/constants.py
```diff
@@ -0,0 +1,82 @@
...
+TORCH_VERSION_AT_LEAST_2_6 = torch_version_at_least("2.6.0")
+TORCH_VERSION_AT_LEAST_2_5 = torch_version_at_least("2.5.0")
+TORCH_VERSION_AT_LEAST_2_4 = torch_version_at_least("2.4.0")
+
+
+class SupportedFormats:
```
wenhuach21: This could be placed in the auto_round foder

auto_round/utils/constants.py
wenhuach21: Okay, having too many utils can confuse developers about where each file should go.

wenhuach21:
the two so files are also need updates

aur61 (approved):

PR #955, opened 2025-10-28
fix bug of imatrix contains 0

n1ck-guo:


auto_round/data_type/int.py
```diff
@@ -71,9 +72,34 @@ def quant_tensor_rnt_sym(tensor, bits=4, group_size=-1, v=0, q_scale_thresh=1e-5
         imatrix = imatrix.reshape(1, -1)
 
         imatrix = imatrix.expand(tensor.numel() // imatrix.numel(), -1)
-        imatrix = imatrix.reshape(tensor.shape)
-
-    scale = search_scales(tensor, bits, qw=imatrix)
+        quant_weights = imatrix.reshape(tensor.shape)
+
+        if torch.min(quant_weights) == 0:
+            logger.warning_once(
```
wenhuach21: is  it possible to wrap it as a function

auto_round/data_type/gguf.py
```diff
@@ -285,6 +285,39 @@ def quant_tensor_asym_dq(
     return qdq_result, {"scale": scale, "d_scale": d_scale}, {"wmin": wmin, "d_wmin": d_wmin}
 
 
+def _imatrix_handle_zero(imatrix, weight, bits):
```
wenhuach21: type annotation

PR #956, opened 2025-10-29
add self attribution and fix avg_bits error

xin3he:


auto_round/auto_scheme/utils.py
```diff
@@ -185,9 +187,12 @@ def compute_layer_bits(
 
     # Compute auxiliary bits (scales, zero-points, or double quantization)
     aux_total_bits = 0
-    if not super_group_size:
+    if "mx_fp" in data_type or "nv_fp" in data_type or "fp4" in data_type:
+        scale_bits = 8
+    else:
         scale_bits = 16
-        zp_bits = weight_bits
+    zp_bits = weight_bits if not is_sym else 0
```
wenhuach21: In GPTQ/AutoRound, although the zero points are identical in symmetric mode, the bit width remains the same as in the asymmetric mode.

xin3he: It's true, maybe we could add a log to indicate user this special case, since the zp is truly useless in the model.

auto_round/auto_scheme/utils.py
```diff
@@ -175,7 +177,7 @@ def compute_layer_bits(
 
     # Determine number of groups based on group size
     if group_size > 0:
-        n_group = out_features * (in_features + group_size - 1) // group_size
+        n_group = out_features * ((in_features + group_size - 1) // group_size)
```
wenhuach21, line 180: nice catch

auto_round/auto_scheme/utils.py
xin3he: add zp bits when "int" in data_type

test/test_cuda/test_auto_scheme.py
```diff
@@ -146,7 +146,7 @@ def test_min_target_bits(self):
     #
     def test_max_target_bits(self):
         model_name = "/models/opt-125m"
-        target_bits = 8.211
+        target_bits = 8.025
```
wenhuach21, line 149: Are you sure  there is only one place needs change?

xin3he: It is. CI had passed, will merge it.

PR #957, opened 2025-10-29
support for model scope 

n1ck-guo:


auto_round/compressors/base.py
```diff
@@ -234,6 +234,9 @@ def __init__(
         device = kwargs.pop("device", None)
         # Scale factor for RAM usage per parameter.
         mem_per_param_scale = kwargs.pop("mem_per_param_scale", None)
+
+        if os.getenv("MODEL_PLATFORM"):
```
hshen14: Suggest follow VLLM usage, which is a bool flag VLLM_USE_MODELSCOPE. By default model platform is HF hub. You may use AUTOROUND_USE_MODELSCOPE or just USE_MODELSCOPE

yiliu30: We have added the `envs.py` for auto-round. 
How about simplifying the name to start with `AR_`?

https://github.com/intel/auto-round/blob/8ac82a4abeccef52586580598f14ba93ff607bdc/auto_round/envs.py#L22-L25

auto_round/utils/model.py
```diff
@@ -180,13 +213,22 @@ def download_hf_model(repo_id, cache_dir=None, repo_type=None, revision=None):
...
+    model_dtype: str = None,
+    device: str = "cpu",
     **kwargs,
 ):
-    from transformers import AutoModel, AutoModelForCausalLM, AutoTokenizer
+    assert platform.lower() in [
+        "hf",
+        "model_scope",
+    ], "current only support hf or model_scope platform to load pretrained model."
+    os.environ["AUTOROUND_USE_MODELSCOPE"] = "model_scope" if platform.lower() == "model_scope" else "hf"
```
yiliu30: It would be better to move it into `envs.py`.

yiliu30 (approved):

PR #969, opened 2025-10-31
fix guff scheme and device_map bug

n1ck-guo:


auto_round/compressors/base.py
```diff
@@ -472,7 +473,15 @@ def _parse_and_set_scheme(self, scheme: Union[str, dict, QuantizationScheme], kw
         """Parse and set the quantization scheme."""
 
         def _parse_and_set(scheme, kwargs):
-            res = ""
+            if kwargs.get("data_type", None) and kwargs["data_type"].endswith("_dq") and not scheme.startswith("gguf"):
+                if "bits" not in kwargs:
+                    data_type = kwargs["data_type"]
+                    raise KeyError(
```
wenhuach21, line 479: why need this, the default bits is in scheme="W4A16"

wenhuach21: all the default values are in scheme, if users want to override some values in  the scheme, then set the specific keys

auto_round/utils/device.py
```diff
@@ -355,6 +355,19 @@ def get_packing_device(device: str | torch.device | None = "auto") -> torch.devi
     raise TypeError(f"Unsupported device type: {type(device)} ({device})")
 
 
+def use_device_map(device_map):
```
wenhuach21: I  have no answer but I think we need a better func_name

auto_round/utils/device.py
```diff
@@ -840,7 +849,7 @@ def set_auto_device_map_for_block_with_tuning(
         num_devices = torch.xpu.device_count()
         device_name = "xpu"
     else:
-        raise RuntimeError("No CUDA or XPU devices found.")
+        return
```
wenhuach21, line 852: Is it possible to support all devices that are already supported by PyTorch,  at least for single-device setups, such as NPUs or other accelerators?

auto_round/compressors/base.py
n1ck-guo: if use auto_round api as this:  auto_round(data_type="rtn_int_sym_dq", bits=3), it will have some bugs since default scheme is W4A16 with super_bits = None and super_group_size = None

auto_round/utils/device.py
n1ck-guo: this change is to solve device_map={"*": "cpu"}

wenhuach21 (approved):

PR #980, opened 2025-11-03
add support for Magistral-Small

n1ck-guo:


auto_round/compressors/mllm/compressor.py
```diff
@@ -189,7 +189,9 @@ def __init__(
 
         if model.config.model_type == "llava" and isinstance(model, PreTrainedModel):
             template = "default"
-        if hasattr(model, "name_or_path") and "Mistral-Small-3.2" in model.name_or_path:
+        if hasattr(model, "name_or_path") and any(
+            [name in model.name_or_path for name in ["Mistral-Small-3.2", "Magistral-Small", "Devstral-Small"]]
```
wenhuach21: is it possible to use architectures? using name is not so robust
https://huggingface.co/mistralai/Devstral-Small-2507/blob/main/config.json#L2>

If using names, better name this list and move to constants ?

n1ck-guo: cannot use architectures or model_config, because they are same with 3.1 and lower version, but need different processor.

PR #981, opened 2025-11-03
Reduce peak gpu memory usage and support moe estimation

xin3he:
- Reduce peak memory usage by calling clear_memory cosidering performance effort. 
- Move best_params to CPU and make sure clear memory before moving back.
- [move loss device to the second card if card_0_in_high_risk](https://github.com/intel/auto-round/pull/981/commits/a27e1dd042d6e19a930a09da6f07ddd0b4c17ce1)
- support Deepseek R1 W4A16 tuning with 3 CUDA cards (80GB) （--enable_torch_compile)
- support llama3.3 70B W4A16 tuning with 2 Intel GPU cards (24GB)（--enable_torch_compile)

### DIFF
--- a/auto_round/utils/device.py
+++ b/auto_round/utils/device.py
@@ -830,7 +830,7 @@ def get_moe_memory_ratio(block: torch.nn.Module) -> float:
 
 
 def estimate_tuning_block_mem(
-    block: torch.nn.Module, input_ids: list[torch.Tensor], pick_samples: int
+    block: torch.nn.Module, input_ids: list[torch.Tensor], batch_size: int
 ) -> tuple[dict, float]:
     """
     Calculates the memory consumption of a specific block in the model.
@@ -838,7 +838,7 @@ def estimate_tuning_block_mem(
     Args:
         block (torch.nn.Module): The block of the model to analyze.
         input_ids (list[torch.Tensor]): A list of input tensors for the block.
-        pick_samples (int): Number of samples to consider for memory estimation.
+        batch_size (int): Number of samples to consider for memory estimation.
 
     Returns:
         tuple: A tuple containing the following:
@@ -871,12 +871,12 @@ def estimate_tuning_block_mem(
             in_features, out_features = get_layer_features(module)
             if in_features is not None and out_features is not None:
                 # Output tensor size: batch_size * seq_len * out_features * element_size
-                output_size = pick_samples * seq_len * out_features * element_size
+                output_size = batch_size * seq_len * out_features * element_size
                 output_memory_gb = output_size / 1024**3
 
                 # If enable_act_quant, add input tensor memory to param_memory
                 if enable_act_quant:
-                    input_size = pick_samples * seq_len * in_features * element_size
+                    input_size = batch_size * seq_len * in_features * element_size
                     input_memory_gb = input_size / 1024**3
                     param_memory_gb += input_memory_gb
             else:
@@ -937,7 +937,7 @@ def set_auto_device_map_for_block_with_tuning(
     device_map,
     input_ids: list[torch.Tensor],
     low_gpu_mem_usage: bool = False,
-    pick_samples: int = 8,
+    batch_size: int = 8,
     output_device: str | torch.device = None,
     card_0_threshold: float = 0.9,
 ):
@@ -949,7 +949,7 @@ def set_auto_device_map_for_block_with_tuning(
         device_map (str | int | dict): Specifies the device mapping.
         input_ids (list[torch.Tensor]): List of input tensors used for estimating memory requirements.
         low_gpu_mem_usage (bool, optional): If True, ignoring input/output memory. Defaults to False.
-        pick_samples (int, optional): Number of samples to consider for memory estimation. Defaults to 8.
+        batch_size (int, optional): Number of samples to consider for memory estimation. Defaults to 8.
         output_device (str | torch.device, optional): Device to move unassigned modules to. Defaults to None.
         card_0_threshold (float, optional): Threshold ratio to determine if the first device is at high risk of
             running out of memory. Defaults to 0.9 (90%).
@@ -996,7 +996,7 @@ def set_auto_device_map_for_block_with_tuning(
     device_0_memory = get_device_memory(device_list[0] if device_list else 0)
     device_1_memory = get_device_memory(device_list[1] if device_list else 1)
     layer_memory_dict, layer_activation_memory, block_input_output_memory, additional_memory = (
-        estimate_tuning_block_mem(block, input_ids, pick_samples)
+        estimate_tuning_block_mem(block, input_ids, batch_size)
     )
     loss_memory = block_input_output_memory / 2  # GB, rough estimate for loss tensor memory
     if low_gpu_mem_usage:


auto_round/compressors/utils.py
```diff
@@ -199,13 +199,16 @@ def check_awq_gemm_compatibility(model, bits, group_size, sym, layer_configs=Non
...
-def collect_best_params(block):
+def collect_best_params(block, low_gpu_mem_usage: bool = False):
     params = {}
     for n, m in block.named_modules():
         if hasattr(m, "orig_layer"):
             params[n] = {}
             for key in m.params.keys():
-                params[n][key] = copy.deepcopy(m.params[key].data)
+                if low_gpu_mem_usage:
+                    params[n][key] = m.params[key].data.cpu()
```
wenhuach21: if the data is already on cpu like cpu tuning, this will causes some issue

auto_round/compressors/base.py
```diff
@@ -2760,6 +2770,7 @@ def _quantize_blocks(
                 modules = [get_module(model, n) for n in names]
                 m = WrapperMultiblock(modules)
 
+            m.config = model.config if hasattr(model, "config") else None
```
wenhuach21, line 2801: del this after tuning

auto_round/compressors/utils.py
xin3he: cpu() will copy and it's expected behavior

auto_round/utils/device.py
```diff
@@ -732,6 +731,75 @@ def find_best_device(layer_name, estimated_memory, layer_idx):
...
+            # Get total number of experts
+            num_experts = getattr(config, "num_local_experts", None)  # Mixtral, PhiMoE, Grok, Llama4
+            if num_experts is None:
+                num_experts = getattr(
+                    config, "num_experts", None
+                )  # Qwen2MoE, Olmo, BailingMoE, GroveMoE, HunYuan, LLaDAMoE
+            if num_experts is None:
+                num_experts = getattr(config, "moe_num_primary_experts", None)  # SmallThinker
+            if num_experts is None:
+                num_experts = getattr(config, "n_routed_experts", None)  # DeepSeek
```
wenhuach21: if num_experts is none, better use all the linear number in the block

auto_round/utils/device.py
```diff
@@ -732,6 +731,75 @@ def find_best_device(layer_name, estimated_memory, layer_idx):
...
+
+    Examples:
+        - Non-MoE model: returns 1.0
+        - Mixtral (2/8 experts): returns 0.25
+        - Qwen2MoE (4/60 experts): returns ~0.067
+    """
+    from auto_round.utils.model import is_moe
+
+    for name, module in block.named_modules():
+        if is_moe(module):
```
wenhuach21: If not is_moe, continue. This way, all the remaining code will have fewer indentation levels, making it easier to read.

auto_round/utils/device.py
```diff
@@ -732,6 +731,75 @@ def find_best_device(layer_name, estimated_memory, layer_idx):
...
+            config = getattr(block, "config", None)
+            if config is None:
+                break
+
+            # Try to get num_experts_per_tok (active experts count)
+            num_experts_per_tok = getattr(
+                config, "num_experts_per_tok", None
```
wenhuach21: this code could be refined to be a function,  get_attr(config, list[str]) and use for

auto_round/compressors/utils.py
wenhuach21: If the parameter is already on the CPU, it won’t be copied but will instead share the same memory.

auto_round/utils/device.py
xin3he: Yes, if any required value is None, we set moe_ratio=1.0 meaning all linear is considered.

auto_round/compressors/base.py
```diff
@@ -2644,7 +2654,6 @@ def _quantize_block(
             set_amax_for_all_moe_layers(block, attr_name="orig_layer.act_max")
 
         if self.enable_quanted_input:
-            clear_memory()
```
wenhuach21: Why remove this clear_memory? All such calls are meant to be added, not omitted.

xin3he: I moved it to a upper place, right after tuning end. Because unwrapper model also requires cuda memory.

xin3he: I add self.low_gpu_memory_usage after moved to upper place, what do you think?

auto_round/compressors/base.py
```diff
@@ -2644,7 +2651,6 @@ def _quantize_block(
...
```
xin3he: fixed

auto_round/compressors/base.py
```diff
@@ -2614,21 +2612,29 @@ def _quantize_block(
...
+                    clear_memory_if_reached_threshold(threshold=0.8)
 
             if i == 0:
                 init_loss = total_loss
 
             if total_loss < best_loss:
                 best_loss = total_loss
                 if not self.not_use_best_mse:
-                    best_params = collect_best_params(block)
+                    best_params = collect_best_params(block, self.low_gpu_mem_usage)
```
wenhuach21: better use cache_devcie, which is more clearer.

auto_round/compressors/base.py
```diff
@@ -2413,6 +2413,7 @@ def _get_current_q_output(
         input_others: dict,
         indices: list[int],
         device: str,
+        output_device: str = "cpu",
```
wenhuach21: better use cache device

auto_round/compressors/base.py
```diff
@@ -2781,13 +2791,15 @@ def _quantize_blocks(
...
 
+            m.config = model.config if hasattr(model, "config") else None
             q_input, input_ids = quantize_block(
                 m,
                 input_ids,
                 input_others,
                 q_input=q_input,
                 device=device,
             )
+            del m.config
```
wenhuach21: if model has no config, this will raise exception

auto_round/utils/device.py
```diff
@@ -412,8 +412,10 @@ def _clear_memory_for_cpu_and_cuda(tensor=None):
         del tensor
     gc.collect()
     if torch.cuda.is_available():
+        torch.cuda.synchronize()
```
wenhuach21: rm this if useless

auto_round/utils/device.py
```diff
@@ -821,6 +937,7 @@ def set_auto_device_map_for_block_with_tuning(
     low_gpu_mem_usage=False,
     pick_samples=8,
     output_device=None,
```
wenhuach21: type annotation

auto_round/wrapper.py
```diff
@@ -601,7 +601,7 @@ def __init__(self, orig_layer, bit=4, group_size=-1, device="cpu"):
     def unwrapper(self, best_params):
         if best_params is None:
             return self.orig_layer
-        v = best_params["v"]
+        v = best_params["v"].to(self.device)
```
wenhuach21, line 604: have not do the same for the  scale parameters, is it correct?

auto_round/compressors/base.py
```diff
@@ -2458,13 +2459,11 @@ def _quantize_block(
...
-
-        if is_complex_device_mapping(self.device_map):
-            set_auto_device_map_for_block_with_tuning(
-                block, self.device_map, input_ids, self.low_gpu_mem_usage, self.batch_size, device
-            )
-        else:
-            block = block.to(device)
+        # card_0_in_high_risk indicates that card_0 memory is already in high usage (90%) w/o any weights
+        # loss_device is used to calculate loss on the second device if available and card_0_in_high_risk
+        card_0_in_high_risk, loss_device = set_auto_device_map_for_block_with_tuning(
```
wenhuach21: The function name doesn’t accurately reflect its behavior. I recommend moving the code that handles the non-auto-device mapping outside the function.

auto_round/wrapper.py
xin3he: In my check, only v is used in the best params

PR #985, opened 2025-11-04
support model_dtype and fix bug of scheme contains quotes, mllm eval

n1ck-guo:


auto_round/compressors/diffusion/compressor.py
```diff
@@ -98,6 +99,7 @@ def __init__(
         device_map: Union[str, torch.device, int, dict] = 0,
         enable_torch_compile: bool = False,
         seed: int = 42,
+        model_dtype: str = None,
```
wenhuach21: hidden in kwargs

PR #992, opened 2025-11-05
fix bug of cannot create adam compressor

n1ck-guo:


auto_round/compressors/base.py
```diff
@@ -3107,3 +3107,7 @@ def _sampling_inputs(
                 current_input_others[key] = input_others[key]
 
         return current_input_ids, current_input_others
+
+
+class LLMCompressor(BaseCompressor):
+    pass
```
wenhuach21, line 3113: add an ut

Issue #1004, opened 2025-11-07
multiple device issues


wenhuach21:
~~~bash
python3 -m auto_round --model /models/Qwen3-8B  --scheme "W4A16"  --format "auto_round" --device_map 2,6
~~~


(autoround) wenhuach@mlp-dgx-01:~/auto-round$ python3 -m auto_round --model /models/Qwen3-8B  --scheme "W4A16"  --format "auto_round" --device_map 2,6
2025-11-06 22:31:03 INFO __main__.py L485: start to quantize /models/Qwen3-8B
2025-11-06 22:31:04 WARNING modeling_utils.py L4670: `torch_dtype` is deprecated! Use `dtype` instead!
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 80.30it/s]
2025-11-06 22:31:07 INFO base.py L596: 'enable_torch_compile' is set to `False` by default. Enabling it can reduce tuning cost by 20%%, but it might throw an exception.
2025-11-06 22:31:07 INFO base.py L354: using torch.bfloat16 for quantization tuning
2025-11-06 22:31:07 INFO base.py L1583: start to cache block inputs
2025-11-06 22:31:17 INFO base.py L1598: caching done
Quantizing model.layers.0:   0%|                                                                                                                                                                                                                                       lib/python3.13/site-packages/torch/autograd/graph.py:841: UserWarning: Flash Attention defaults to a non-deterministic algorithm. To explicitly enable determinism call torch.use_deterministic_algorithms(True, warn_only=False). (Triggered internally at /pytorch/aten/src/ATen/native/transformers/cuda/attention_backward.cu:114.)
  return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
2025-11-06 22:31:56 INFO base.py L2651: quantized 7/7 layers in the block, loss iter 0: 0.000127 -> iter 132: 0.000056
Quantizing model.layers.1:   3%|██████▏                                                                                                                                                                                                                        | 1/36 [00:42<24:36, 42.18s/it]2025-11-06 22:32:27 INFO base.py L2651: quantized 7/7 layers in the block, loss iter 0: 0.000299 -> iter 184: 0.000103
[...]

PR #1007, opened 2025-11-07
fix multiple devices issue in Compressor and AutoScheme

wenhuach21:


wenhuach21:
fix https://github.com/intel/auto-round/issues/1004

Issue #972, opened 2025-10-31
fix v0.9.0 release CUDA CI


xin3he:

n1ck-guo:
pr https://github.com/intel/auto-round/pull/1014

PR #1010, opened 2025-11-10
fix cuda low_cpu_mem_usage ut

Kaihui-intel:
ut bug https://github.com/intel/auto-round/issues/972 test_main_func

### DIFF
--- a/test/test_cuda/test_main_func.py
+++ b/test/test_cuda/test_main_func.py
@@ -212,7 +212,6 @@ class TestMainFunc(unittest.TestCase):
             group_size=group_size,
             iters=2,
             seqlen=10,
-            dataset=self.llm_dataloader,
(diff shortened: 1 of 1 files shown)


test/test_cuda/test_main_func.py
```diff
@@ -212,7 +212,6 @@ def test_low_cpu_mem_usage(self):
             group_size=group_size,
             iters=2,
             seqlen=10,
-            dataset=self.llm_dataloader,
```
wenhuach21, line 215: May I know what's the real reason that causes the ut fail and why change this could fix it.

Kaihui-intel: This test was copied from the test_cpu and mistakenly included llm_dataloader.

PR #1025, opened 2025-11-13
fix bug of lm_head and dispatch model,gguf eval

n1ck-guo:


wenhuach21:
add more uts if the existing lm-head uts could not expose the issue

auto_round/compressors/base.py
```diff
@@ -2005,8 +2007,25 @@ def try_cache_inter_data_gpucpu(self, block_names, nsamples, layer_names=None, l
                             device_map = infer_auto_device_map(
                                 self.model, max_memory=new_max_memory, no_split_module_classes=no_split_modules
                             )
-
-                            self.model = dispatch_model(self.model, device_map=device_map)
+                            if len(devices) > 1 and "cpu" in device_map.values():
+                                logger.warning(
+                                    "Not enough memory cause the CPU to be used, which may severely impact speed."
```
wenhuach21: memory->vram, cpu->ram

auto_round/compressors/base.py
```diff
@@ -2005,8 +2007,25 @@ def try_cache_inter_data_gpucpu(self, block_names, nsamples, layer_names=None, l
...
+                                    " Please consider using more cards."
+                                )
+
+                            try:
+                                self.model = dispatch_model(self.model, device_map=device_map)
```
wenhuach21, line 2017: besides this one,  there are several other dispatch_model calls in the project, so it would be better to wrap this as a function

n1ck-guo: I checked, other dispatch_model have different use cases than this one and will not cause the offload_dir bug.

PR #1043, opened 2025-11-18
[High Risk]reduce vram usage for optimized RTN mode

wenhuach21:


auto_round/compressors/base.py
```diff
@@ -1354,10 +1355,14 @@ def _quantize_rtn(self) -> tuple[torch.nn.Module, dict[str, Any]]:
...
             any("gguf" in fmt and "k" in fmt for fmt in getattr(self, "formats", [])) or self.super_bits is not None
         )
+        if has_gguf_k:
+            self.model.to(torch.float32)
 
         self._quantize_embedding_layer()
 
         self.model.to("cpu")
+        # Release memory
+        clear_memory(device_list=self.device_list)
```
xin3he, line 1371: no need to clear memory, when quantizing block, the memory usage is not such high in my experiment.

auto_round/compressors/base.py
```diff
@@ -1050,7 +1050,7 @@ def _get_save_folder_name(self, format_str: str) -> str:
 
         return self.orig_output_dir
 
-    @torch.inference_mode()
+    # @torch.inference_mode()
```
xin3he: why removing it？

auto_round/export/export_to_gguf/convert.py
```diff
@@ -406,9 +406,13 @@ def prepare_tensors(cls):
 
         modify_name = _special_name_handle(cls, name)
         orig_device = data_torch.device
-        data_torch = data_torch.to("cpu")
+        import os
+
+        import psutil
+
```
wenhuach21: delete later

auto_round/data_type/gguf.py
```diff
@@ -437,14 +468,17 @@ def quant_tensor_gguf_asym_dq(
...
     orig_dtype = tensor.dtype
     maxq = 2**bits - 1
     group_size = 16 if bits == 2 else 32
-    split_num = 1
-    for dim in tensor.shape:
-        if dim > 100_000:
-            split_num = 16
-            break
+    if split_num is None:
+        split_num = 1
```
n1ck-guo: split_num = 16 if max(tensr.shape) > 100_000 else 1

auto_round/data_type/int.py
```diff
@@ -79,9 +109,8 @@ def quant_tensor_rtn_sym(tensor, bits=4, group_size=-1, v=0, q_scale_thresh=1e-5
 
     scale = search_scales(tensor, bits, qw=imatrix)
     scale = torch.where(scale < 0, torch.clamp(scale, max=-q_scale_thresh), torch.clamp(scale, min=q_scale_thresh))
-    int_w = round_ste(tensor / scale + v)
-    q = torch.clamp(int_w, -maxq, maxq - 1)
-    qdq_result = (scale * q).to(tensor.dtype)
+    int_w = tensor.div(scale).round_().clamp_(-maxq, maxq - 1)
```
n1ck-guo, line 112: div_?

wenhuach21: to avoid changing the origin tensor

auto_round/export/export_to_gguf/packing.py
```diff
@@ -52,13 +52,28 @@ def ggml_quant(
...
     n_blocks = data.nelement() // block_size
+    split_num = 1
+    for dim in data.shape:
+        if dim > 100_000:
+            split_num = 16
+            break
+
     blocks = data.reshape((n_blocks, block_size))
     quant_func = GGML_QUANT_TYPE[ggml_type]
     try:
```
n1ck-guo, line 33: Why not process the data directly here if the split_num greater than 1, so that the q*k_quant funcs do not need to be modified?

wenhuach21: split>1 will move some packing operation to cpu

n1ck-guo: I mean, split the data. Move the part data to cuda and pass to quant func. This method should be faster and require less code modification. It would also be easier to support multiple nodes.

wenhuach21: this function conducts the quantitation and packing, so the tensor could not split which requires complex concat of packed tensor. The only place that could change is the quantization function. In that case, we still need to change each packing function and each quantization function in  datatype that call the search funciton.

n1ck-guo: I don't quite understand. Just spilt the block with dim=0, and concatenate the returned results with dim=0. llamacpp uses a similar method.

PR #1045, opened 2025-11-19
Add static FP8 attention support

yiliu30:
resolve https://github.com/intel/auto-round/issues/938
LLMC format export is not supported for now, as vLLM loading is still a work in progress.
But it can be used on HPU.

### DIFF
--- a/auto_round/__main__.py
+++ b/auto_round/__main__.py
@@ -549,8 +549,6 @@ def tune(args):
         nblocks=args.nblocks,
         to_quant_block_names=args.to_quant_block_names,
         scale_dtype=args.scale_dtype,
-        static_kv_dtype=args.static_kv_dtype,
-        static_attention_dtype=args.static_attention_dtype,
     )
     scheme_config = SchemeExtraConfig(
         bits=args.bits,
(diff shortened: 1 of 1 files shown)


auto_round/experimental/utils.py
```diff
@@ -0,0 +1,75 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+from auto_round.utils import logger
+
+
+def fp8_per_tensor_qdq(
```
n1ck-guo: The function name is a bit strange.I suggest changing it to a more easily understood name.

auto_round/compressors/base.py
```diff
@@ -237,6 +237,7 @@ def __init__(
         enable_deterministic_algorithms = kwargs.pop("enable_deterministic_algorithms", False)
         self.momentum = kwargs.pop("momentum", 0.0)
         static_kv_dtype = kwargs.pop("static_kv_dtype", None)
+        static_attention_dtype = kwargs.pop("static_attention_dtype", None)
```
n1ck-guo, line 240: Does \_\_main\_\_.py also need add this parameter?

auto_round/experimental/utils.py
yiliu30: Good catch, updated.

Issue #1040, opened 2025-11-17
lm_head is not converted to QuantLinear with MXFP4/8


xin3he:
lm_head quantization still have some issues.
- need deepcopy if tied_word_embedding = True
- export is not applied for lm_head

Shall we warn user that lm_head is not supported? @WeiweiZhang1  @wenhuach21

xin3he:
If user prefer to quantize lm_head, what's the solution?

<img width="1072" height="88" alt="Image" src="https://github.com/user-attachments/assets/2b73b39c-a14f-4d64-9599-a0501cff6d45" />

xin3he:
I'm using the model for export after quantize_and_save() and just aware that AutoRound requires reloading before inference.

wenhuach21:
do you enable quant_lm_head or set bits for lm-head

xin3he:
<img width="1059" height="44" alt="Image" src="https://github.com/user-attachments/assets/4f080626-dcf0-4de6-90f9-1050ebbde481" />

xin3he:
<img width="601" height="497" alt="Image" src="https://github.com/user-attachments/assets/411f4be8-b9fd-4a50-a2e2-179df226dd3c" />

After reloading I saw the lm_head is quantized, not sure what is happening. @WeiweiZhang1  Do you have any comments? Do you think it's a bug or it is designed.

PR #1051, opened 2025-11-20
Support mxfp nvfp lmhead quant

WeiweiZhang1:
related issue: https://github.com/intel/auto-round/issues/1040

### DIFF
--- a/test/test_cpu/test_export.py
+++ b/test/test_cpu/test_export.py
@@ -334,7 +334,6 @@ class TestAutoRound(unittest.TestCase):
 
         shutil.rmtree(quantized_model_path, ignore_errors=True)
 
-
     def test_awq_lmhead_export(self):
         bits, sym, group_size = 4, False, 128
         model_name = "/tf_dataset/auto_round/models/microsoft/phi-2"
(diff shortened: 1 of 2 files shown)


auto_round/compressors/base.py
```diff
@@ -1731,6 +1731,11 @@ def _quantize_layers(self, layer_names: list, layer_inputs: dict) -> None:
 
         for layer_name in copy.deepcopy(layer_names):
             if layer_name not in layer_inputs:
+                if self.act_bits < 16 and not self.act_dynamic:
+                    logger.warning(
+                        f"Due to insufficient resources: act_max hook for layer '{layer_name}' is unavailable. "
```
wenhuach21: support it via block-wise forward, please refer to auto-scheme code. @xin3he  could you take this task?

wenhuach21: if there are quantized layers outside the blocks, we could switch to this mode

WeiweiZhang1: > if there are quantized layers outside the blocks, we could switch to this mode

this happens within _quantize_layers function, which only for outside of block layers. don't understand what switching to this mode means.

wenhuach21: AutoScheme provide an advanced way that could calibrate all the layers in the model with block-wise forward

auto_round/compressors/base.py
```diff
@@ -1731,6 +1731,11 @@ def _quantize_layers(self, layer_names: list, layer_inputs: dict) -> None:
...
+                        f"Static activation quantization is currently not supported or ineffective for this layer."
```
wenhuach21: why not converting the layers to 16 bits except for FP8 E5M2, in which case we should log a  warning to remind that unit scale is used

auto_round/export/export_to_llmcompressor/export_to_fp.py
```diff
@@ -51,8 +51,6 @@
 
 
 def pack_layer(name, model, backend, device=None):
-    if name == "lm_head":  # TODO: Check vLLM inference status to determine whether to enable this feature
```
wenhuach21, line 54: export to gptq/awq has such code as well, which is incorrect

xin3he (approved):
LGTM

auto_round/export/export_to_autogptq/export.py
```diff
@@ -130,8 +130,6 @@ def convert_from_autogptq_dynamic(dynamic_config: dict) -> dict:
 
 
 def pack_layer(name, model, backend, device=None):
-    if name == "lm_head":  ##dese not support lm-head
```
wenhuach21, line 133: could you help add ut for gptq and awq with quantized lm-head

PR #1067, opened 2025-11-25
Support transformers loading quantized moe model

mengniwang95:
- Update modeling files of llama4 and gptoss
- Add moe handler in convert_hf_model
- Add ut for llama4 and gptoss

### DIFF
--- a/.azure-pipelines/docker/Dockerfile.devel
+++ b/.azure-pipelines/docker/Dockerfile.devel
@@ -19,11 +19,8 @@ FROM ubuntu:${UBUNTU_VER}
 ENV LANG C.UTF-8
 
 RUN apt-get update && apt-get install -y --no-install-recommends --fix-missing \
-    python3 \
-    python3-pip \
-    python3.12-dev \
-    autoconf \
     build-essential \
+    ca-certificates \
     git \
     libomp-dev \
     numactl \
@@ -33,7 +30,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends --fix-missing \
     jq \
     vim
 
-RUN ln -sf $(which python3) /usr/bin/python
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 
 ARG USER_ID=1000
 ARG GROUP_ID=1000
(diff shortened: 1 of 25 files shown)


auto_round/inference/convert_model.py
```diff
@@ -582,6 +583,9 @@ def convert_hf_model(model: nn.Module, target_device: str = "cpu") -> tuple[nn.M
     elif packing_format == "auto_round:gptq":
         packing_format = "auto_round:auto_gptq"
 
+    # Preprocess model before replace layers
+    model = _handle_moe_model(model)
```
yiliu30, line 587: Overall, looks good to me! 
I have one concern about RAM usage—could you confirm if all weights are on the RAM at this stage? And does this change move all model weights into RAM?

test/test_cpu/test_moe_model.py
```diff
@@ -0,0 +1,78 @@
...
+        scheme=scheme,
+        nsamples=2,
+        iters=iters,
+        fp_layers="self_attn,router,lm_head,mlp.gate",
+    )
+    quantized_model, save_folder = autoround.quantize_and_save(format="auto_round", output_dir=output_dir)
+    return quantized_model
+
+
+def test_gptoss(setup_gpt_oss):
```
yiliu30, line 49: It would be better to add GPU UT that runs a simple prompt to verify the correctness of the replacement.

auto_round/inference/convert_model.py
mengniwang95: when you use transformers to load the quantzied model, model is meta and weights are not loaded yet at this stage

yiliu30: If I understand correctly, `_handle_moe_model` move all the weights into RAM?

mengniwang95: > If I understand correctly, `_handle_moe_model` move all the weights into RAM?

the model is meta and weights are not loaded yet, calling _handle_moe_model will skip moving weights into RAM since the model is meta

test/test_cpu/test_moe_model.py
mengniwang95: done

PR #1086, opened 2025-12-03
Fix MXFP/NVFP + FP8 Attn/KV

yiliu30:
- **add fp8 kv**
- **calib for kv/attn**
- **add ut**

### DIFF
--- a/auto_round/inference/backend.py
+++ b/auto_round/inference/backend.py
@@ -1003,10 +1003,6 @@ def process_requirement(requirements: list, target_device="cuda", logger_level="
 
         if gptq_req:
             commands.append(f"pip install -v {gptq_req} --no-build-isolation")
-            try:
-                require_version("numpy<2.0")
-            except:
-                commands.append("pip install 'numpy<2.0'")
 
         if other_reqs:
             other_str = " ".join(other_reqs)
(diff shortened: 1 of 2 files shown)


test/test_cpu/test_mxfp_nvfp.py
```diff
@@ -370,6 +370,62 @@ def test_qwen_moe_quant_infer(self):
...
+            # scheme,  static_kv_dtype, static_attention_dtype
+            ("MXFP4", None, "fp8"),
+            ("MXFP4", "fp8", None),
+            ("MXFP8", None, "fp8"),
+            ("MXFP8", "fp8", None),
+            ("NVFP4", None, "fp8"),
+            ("NVFP4", "fp8", None),
+        ]
+    )
+    def test_fp8_kv_attn(self, scheme, static_kv_dtype, static_attention_dtype):
```
n1ck-guo, line 384: Will this ut increase the processing time significantly? If so, I think it would be best to move the full test to CUDA UT. The cpu ut retains two is enough, in my opinion.

n1ck-guo: Please help to test how long it takes to run once on the CPU. If the speed is acceptable, then there are no problems.

yiliu30: I’d prefer to cover all these cases and reduce the test model to one decoding layer. Thanks!

yiliu30:
@chensuyue , please help cherry-pick to 0.9.2, thx!

PR #1098, opened 2025-12-05
fix gguf acc and oom bug when iters > 0

n1ck-guo:
* fix gguf acc bug when iters > 1
* fix gguf oom when iters > 1

The main reason is that the deleted code did not consider the case where iters > 0, the logic differs from RTN. `layer_names = self._get_quantized_layer_names_outside_blocks()` only returns `lm_head`, and the new code will cause a loss of embedding data. Testing shows that removing it does not affect memory consumption.
Llama-3.1-8B-Instruct --iters 0:
 'peak_ram': 12.49GB, 'peak_vram': {'0': 9.34GB}  same as before, no change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ See our [paper](https://arxiv.org/pdf/2309.05516) for more details. For usage in
 
 ## 🆕 What's New
 
+* [2025/12] The [AutoRoundV2 paper](http://arxiv.org/abs/2512.04746) is now publicly available on arXiv. Turn on  `enable_alg_ext` and use the **AutoScheme** API for mixed-precision quantization to reproduce the results.
+
 * [2025/11] AutoRound has landed in **LLM-Compressor**: [*Usage*](https://github.com/vllm-project/llm-compressor/tree/main/examples/autoround/README.md).
 
 * [2025/11] An **enhanced GGUF** quantization algorithm is available via `--enable_alg_ext`: [*Accuracy*](./docs/gguf_alg_ext_acc.md).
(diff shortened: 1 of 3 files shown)


auto_round/compressors/base.py
```diff
@@ -2440,16 +2440,18 @@ def _quantize_layer(
...
             with torch.no_grad():
-                layer(torch.cat(inputs, dim=0))
-            for handle in hook_handles:
-                handle.remove()
-        else:
-            hook_handles = self._register_act_max_hook(layer)
-            if hook_handles:
-                layer(torch.cat(q_inputs, dim=0))
+                for input in tmp_inputs:
+                    layer(input)
```
wenhuach21, line 2454: why remove torch.cat, as we may use different batch_size.
how about removing the  cat only for batch_size=1

auto_round/export/export_to_gguf/export.py
```diff
@@ -205,21 +203,6 @@ def pack_gguf_layer(
         if len(model.last_layer_name_to_block_name) == 0:
             for gguf_model in gguf_model_instance_global:
                 gguf_model.current_packing_block = None
-    if name in model.names_outside_blocks:
-        # Packing block
-        for gguf_model in gguf_model_instance_global:
-            gguf_model.current_packing_block = name
```
wenhuach21, line 211: What device is used to pack the LM head currently, CPU or CUDA? If it’s on the CPU, that’s suboptimal. If it’s on CUDA, there is a potential risk: when many layers outside the main block need to be quantized, all  layers might be moved to CUDA and then transferred back to the CPU one by one after packing, which could cause out of vram.

wenhuach21: Please check both iters=0 and iters=1.

n1ck-guo: lm head work on cuda. In fact, deleting this code will not affect the logic. For all layers, the packing process is the same, packing on cuda, add packed data to gguf_writer and clean the weight.

n1ck-guo: Each layer will clean parameters after packing. (set the parameters to meta which is more safe than set to None)

wenhuach21: please do not merge the pr if the issue has not been resolved. 

If the lm-head is on cuda, then that means all layers outside blocks are on cuda. Because you delete the code, then the layer will not be packed immediately

Issue #1099, opened 2025-12-05
[CI] use common requirements.txt as much as possible for UT


chensuyue:
https://github.com/intel/auto-round/blob/main/requirements.txt   
https://github.com/intel/auto-round/blob/main/test/test_cuda/requirements.txt   
Make sure the released package dependency aligned with the tested one.

PR #1103, opened 2025-12-08
remove non-essential requirements

n1ck-guo:
remove non-essential requirements,  check when use.

sentencepiece: used by gguf
pillow: used to quant vlm non-text module
lm-eval: used for evaluation

### DIFF
--- a/auto_round/compressors/mllm/compressor.py
+++ b/auto_round/compressors/mllm/compressor.py
@@ -187,7 +187,8 @@ class MLLMCompressor(BaseCompressor):
             from transformers.utils.versions import require_version
 
             require_version(
-                "PIL", "Pillow is required for quantizing non-text modules, please install it with `pip install Pillow`"
+                "pillow",
+                "pillow is required for quantizing non-text modules, please install it with `pip install pillow`",
             )
         all_blocks = get_block_names(model, quant_nontext_module)
         self.quant_block_list = find_matching_blocks(model, all_blocks, to_quant_block_names)


requirements.txt
```diff
@@ -1,13 +1,10 @@
...
 datasets
-py-cpuinfo
-sentencepiece
 numpy
-tqdm
-packaging
-pillow
-torch
-transformers>=4.38
+packaging # for python version <= 3.9
```
wenhuach21: we don't support python 3.9 anymore, so is it safe to remove this package?

requirements.txt
```diff
@@ -1,13 +1,10 @@
 # 1.5.1<accelerate<1.10.0 may cause potentially high RAM usage; versions 1.10.0 or above are recommended.
```
wenhuach21, line 1: this checker should be added in code and a warning should be dumped

auto_round/compressors/mllm/compressor.py
```diff
@@ -182,6 +183,9 @@ def __init__(
 
         self.model = model
         quant_nontext_module = self._check_quant_nontext(layer_config, quant_nontext_module)
+        if quant_nontext_module and importlib.util.find_spec("PIL") is None:
```
wenhuach21: how about using the existing  `require_version` function

requirements.txt
n1ck-guo: \> 3.9 Python comes with built-in packing.

test/test_cpu/requirements.txt
```diff
@@ -1,8 +1,11 @@
 addict
 modelscope
 gguf
+sentencepiece
```
yiliu30, line 4: @XuehaoSun @chensuyue, for some of test, we'd better use the same requirement.txt as AutoRound’s.

test/test_cpu/requirements.txt
```diff
@@ -1,8 +1,11 @@
...
 torchvision
 parameterized
+pillow
 numba
 #TODO: (yiliu30) replace it with the release version
-llmcompressor @ git+https://github.com/vllm-project/llm-compressor.git@7b28d78
\ No newline at end of file
+llmcompressor @ git+https://github.com/vllm-project/llm-compressor.git@7b28d78
```
chensuyue, line 10: let's update to main branch? @yiliu30

test/test_cpu/requirements.txt
chensuyue: https://github.com/intel/auto-round/issues/1099 tracking by this issue.

test/test_cpu/requirements.txt
yiliu30: I’ll update it in a separate PR and add the CUDA UT for LLMC.

PR #1104, opened 2025-12-08
Move packed res to cpu

yiliu30:
Signed-off-by: yiliu30 <yi4.liu@intel.com>

### DIFF
--- a/test/test_cpu/test_export.py
+++ b/test/test_cpu/test_export.py
@@ -339,6 +339,7 @@ class TestAutoRound(unittest.TestCase):
         model_name = "/tf_dataset/auto_round/models/microsoft/phi-2"
         layer_config = {
             "lm_head": {"bits": 4},  # set lm_head quant
+            "layer": {"bits": 16},
         }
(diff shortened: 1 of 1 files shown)


auto_round/compressors/base.py
```diff
@@ -1332,6 +1332,8 @@ def _quantize_layer_via_rtn(self, name: str, dtype: torch.dtype = None, to_cpu=T
             self._immediate_pack(name)
             if to_cpu:
                 m = m.to("cpu")
+                packed_m = get_module(self.model, name)
```
wenhuach21, line 1335: Are there different modules? I always thought they were the same. Is it possible to make m is self.model.name identical?

yiliu30: The `m` is the `WrapperLinear` while the `packed_m` here is `QuantLinear`.
During packing, `self.model` gets updated from `WrapperLinear` to `QuantLinear`. So we can't move packing result to cpu by operating on m.

PR #1124, opened 2025-12-11
Use modelscope cache in CPU UT

XuehaoSun:
This pull request introduces improvements to the test setup and cleanup process for model caching, as well as a minor dependency change in the test environment script. The most important changes are grouped below:

**Test infrastructure improvements:**

* Added logic in `test/test_cpu/test_model_scope.py` to copy model files from `/tf_dataset/auto_round/modelscope` to the cache directory `/home/hostuser/.cache/modelscope` before running tests, ensuring required models are available for testing.
* Updated the test cleanup process to remove the cache directory after tests, preventing leftover files and ensuring a clean environment for subsequent test runs.
* Imported the `os` module in `test/test_cpu/test_model_scope.py` to support new filesystem operations.

**Environment setup change:**

* Removed installation of `lm_eval==0.4.9.1` from the `.azure-pipelines/scripts/ut/run_ut.sh` script, streamlining dependency management for the test environment.

### DIFF
--- a/test/test_cpu/test_model_scope.py
+++ b/test/test_cpu/test_model_scope.py
@@ -1,4 +1,5 @@
 import copy
+import os
 import shutil
 import sys
 import unittest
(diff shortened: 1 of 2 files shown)


test/test_cpu/test_model_scope.py
```diff
@@ -25,10 +26,18 @@ def setUpClass(self):
         self.saved_path = "./saved"
         self.dataset = LLMDataLoader()
 
+        self.source_path, self.cache_path = "/tf_dataset/auto_round/modelscope", "/home/hostuser/.cache/modelscope"
+        if os.path.exists(self.source_path):
+            if not os.path.exists("/home/hostuser/.cache"):
+                os.makedirs("/home/hostuser/.cache")
+            shutil.copytree(self.source_path, self.cache_path, dirs_exist_ok=True)
```
xin3he, line 33: Instead of copying model, we can use `os.environ["MODELSCOPE_CACHE"]="/tf_dataset/auto_round/modelscope"`

PR #1128, opened 2025-12-11
fix bug of outside block layers do not use best_params

n1ck-guo:


WeiweiZhang1 (approved):

auto_round/compressors/utils.py
```diff
@@ -208,12 +208,18 @@ def check_awq_gemm_compatibility(model, bits, group_size, sym, layer_configs=Non
...
     """Collect the best parameters from the block to the specified device."""
+    from auto_round.wrapper import WrapperLinear
+
     params = {}
-    for n, m in block.named_modules():
-        if hasattr(m, "orig_layer"):
-            params[n] = {}
-            for key in m.params.keys():
-                params[n][key] = m.params[key].data.to(cache_device, copy=True)
+    if isinstance(block, WrapperLinear):
```
wenhuach21: note quite safe. We may have other wrapper layers like wrapper_embedding or something like that.

how about using if hasattr(m, "orig_layer"):

PR #1135, opened 2025-12-12
Ensure that None is returned if the corresponding value does not exist in the dictionary.

xin3he:
We need to enhance the CI to find this issue as early as possible


xin3he:
https://github.com/intel/auto-round/issues/1136

PR #1138, opened 2025-12-12
fix bug and support shared_layers in cli

xin3he:


auto_round/__main__.py
```diff
@@ -162,6 +162,14 @@ def __init__(self, *args, **kwargs):
             action="store_true",
             help="for auto scheme whether ignore scale zp bits calculation ",
         )
+        tuning.add_argument(
+            "--shared_layer",
```
wenhuach21: please change to `shared_layers` that used in our API, or " --shared_layers and --shared_layer" to support them both

wenhuach21 (changes requested):
check the comment

auto_round/__main__.py
wenhuach21: and better move it from tuning args to the scheme argument

auto_round/__main__.py
```diff
@@ -285,6 +285,14 @@ def __init__(self, *args, **kwargs):
         scheme.add_argument(
             "--disable_act_dynamic", action="store_true", help="Use static instead of dynamic activation quantization. "
         )
+        scheme.add_argument(
+            "--shared_layers",
+            type=str,
+            nargs="+",
+            action="append",
+            default=[],
```
n1ck-guo: default value better not use []. Because all calls to parse_args() share the same list object, it may cause some bugs. Better set to None.

xin3he:
```suggestion
            default=None,
```

wenhuach21 (approved):

PR #1143, opened 2025-12-15
Fix device mismatch of nvfp fuse scale

WeiweiZhang1:


auto_round/data_type/utils.py
```diff
@@ -235,78 +235,76 @@ def get_gaudi_fp8_ste_func():
...
             return
-        if hasattr(submodule.q_proj, global_scale_name):
-            setattr(submodule.q_proj, global_scale_name, global_scale.clone())
-        if hasattr(submodule.k_proj, global_scale_name):
-            setattr(submodule.k_proj, global_scale_name, global_scale.clone())
-        if hasattr(submodule.v_proj, global_scale_name):
-            setattr(submodule.v_proj, global_scale_name, global_scale.clone())
 
-        del global_scale
+        device = scales[0].device
```
n1ck-guo: how about handling the device and dtype in _collect_scales

Issue #899, opened 2025-10-15
Refactor modelling replacement code


yiliu30:
More details please refer to https://github.com/intel/auto-round/pull/887#discussion_r2427738679

PR #1153, opened 2025-12-17
Refactor module replacement

yiliu30:
resolve https://github.com/intel/auto-round/issues/899
- [x] Replace GPT-OSS/LLAMA
- [x] UT

### DIFF
--- a/auto_round/inference/convert_model.py
+++ b/auto_round/inference/convert_model.py
@@ -20,7 +20,7 @@ import torch.nn as nn
 from tqdm import tqdm
 from transformers.pytorch_utils import Conv1D
 
-from auto_round.export.export_to_autoround import AutoRoundFormat
+from auto_round.formats import AutoRoundExportFormat
 from auto_round.inference.backend import (
     BackendInfos,
     dynamic_import_inference_linear,
@@ -424,10 +424,10 @@ def _create_quant_layer(layer, layer_backend, config, in_features, out_features)
             bias=bias,
         )
     elif (
-        AutoRoundFormat.FP8_STATIC.value in layer_backend
-        or AutoRoundFormat.MXFP8.value in layer_backend
-        or AutoRoundFormat.MXFP4.value in layer_backend
-        or AutoRoundFormat.NVFP4.value in layer_backend
+        AutoRoundExportFormat.FP8_STATIC.value in layer_backend
+        or AutoRoundExportFormat.MXFP8.value in layer_backend
+        or AutoRoundExportFormat.MXFP4.value in layer_backend
+        or AutoRoundExportFormat.NVFP4.value in layer_backend
     ):
         return QuantLinear.from_original(config, layer)
 
--- a/auto_round/__main__.py
+++ b/auto_round/__main__.py
@@ -140,7 +140,9 @@ class BasicArgumentParser(argparse.ArgumentParser):
             "--format",
             default="auto_round",
             type=str,
-            help="Output format for the quantized model." "'auto_round' is the recommended format",
+            help="Output format for the quantized model."
+            "'auto_round' is the recommended format"
+            "use command `auto_round list format` to show all supported formats with support scheme.",
         )
         basic.add_argument(
             "--output_dir",
@@ -462,14 +464,25 @@ class BasicArgumentParser(argparse.ArgumentParser):
         )
 
 
-def setup_parser(recipe="default"):
+def list_item():
+    args = argparse.ArgumentParser()
+    args.add_argument("item", type=str, help="item to list, e.g., format")
+    args = args.parse_args()
+    if args.item == "format":
+        from auto_round.formats import OutputFormat
+
+        print("AutoRound supported output formats and quantization scheme:")
+        print(OutputFormat.get_support_matrix())
+
+
+def start(recipe="default"):
     recipe = RECIPES[recipe]
     parser = BasicArgumentParser()
     args = parser.parse_args()
     for k, v in recipe.items():
         if getattr(args, k) is None:
             setattr(args, k, v)
-    return args
+    tune(args)
 
 
 def tune(args):
@@ -881,27 +894,33 @@ def run_eval():
 
 
 def run():
-    if "--eval" in sys.argv:
-        sys.argv.remove("--eval")
+    if "list" in sys.argv or "--list" in sys.argv:
+        if "list" in sys.argv:
+            sys.argv.remove("list")
+        if "--list" in sys.argv:
+            sys.argv.remove("--list")
+        list_item()
+        exit()
+    if "--eval" in sys.argv or "eval" in sys.argv:
+        if "--eval" in sys.argv:
+            sys.argv.remove("--eval")
+        if "eval" in sys.argv:
+            sys.argv.remove("eval")
         run_eval()
     else:
-        args = setup_parser()
-        tune(args)
+        start()
 
 
 def run_best():
-    args = setup_parser("best")
-    tune(args)
+    start("best")
 
 
 def run_light():
-    args = setup_parser("light")
-    tune(args)
+    start("light")
 
 
 def run_fast():
-    args = setup_parser("fast")
-    tune(args)
+    start("fast")
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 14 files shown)


auto_round/inference/convert_model.py
```diff
@@ -583,7 +584,7 @@ def convert_hf_model(model: nn.Module, target_device: str = "cpu") -> tuple[nn.M
         packing_format = "auto_round:auto_gptq"
 
     # Preprocess model before replace layers
-    model = _handle_moe_model(model)
+    model = update_moe_module_if_needed(model)
```
wenhuach21: Is it possible to make this function more general? Besides MoE, there may be other modules that need to be replaced.

auto_round/modelling/__init__.py
```diff
@@ -11,3 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from .gpt_oss import *
```
wenhuach21: If Transformers changes a module path (e.g., transformers.models.llama4.modeling_llama4), it will cause an import error. Could we instead import the corresponding file from the modeling folder on demand?

wenhuach21: for example, some older transformers may have no llama4

auto_round/inference/convert_model.py
yiliu30: Yes, the replacement logic works for other modules as well. I’ll rename the function.

yiliu30: updated

auto_round/modelling/__init__.py
yiliu30: Good point! I’ve suppressed the import error and replaced it with a warning instead. Please take a look

auto_round/inference/convert_model.py
```diff
@@ -31,7 +31,7 @@
 from auto_round.inference.utils import _expand_regex_config
 from auto_round.logger import logger
 from auto_round.schemes import QuantizationScheme
-from auto_round.special_model_handler import _handle_moe_model
+from auto_round.special_model_handler import update_module_if_needed
```
wenhuach21: I think replace module is more clear, but it's up to you

auto_round/modelling/__init__.py
```diff
@@ -11,3 +11,22 @@
...
+    if module_name.split(".")[-1].startswith("_"):
+        continue
+    try:
+        mod = importlib.import_module(module_name)
+        # Re-export symbols
+        if hasattr(mod, "__all__"):
+            for name in mod.__all__:
+                globals()[name] = getattr(mod, name)
+    except (ImportError, ModuleNotFoundError) as e:
+        logger.warning("Optional module %s not available: %s", module_name, e)
```
wenhuach21: Warning is not a good choice as users may not quantize the llama4/gpt-oss, 
Better change it to on demand import .

 we have lazy import or place the import in the code and use a dict to mapping

yiliu30: Updated

auto_round/modelling/replace_modules.py
```diff
@@ -0,0 +1,180 @@
...
+import torch
+from tqdm import tqdm
+from transformers import PreTrainedModel
+
+from auto_round.utils import LazyImport, logger
+
+BUILTIN_MODULES = {
+    "Llama4TextMoe": LazyImport("auto_round.modelling.llama4"),
+    "GptOssMLP": LazyImport("auto_round.modelling.gpt_oss"),
+}
```
yiliu30, line 28: @wenhuach21 please help check if that makes sense to you, thanks.
No additional warnings for other models.

wenhuach21: Thanks for the refinements. I don’t have any further issues.

auto_round/modelling/replace_modules.py
```diff
@@ -0,0 +1,180 @@
...
+
+import torch
+from tqdm import tqdm
+from transformers import PreTrainedModel
+
+from auto_round.utils import LazyImport, logger
+
+BUILTIN_MODULES = {
+    "Llama4TextMoe": LazyImport("auto_round.modelling.llama4"),
+    "GptOssMLP": LazyImport("auto_round.modelling.gpt_oss"),
```
wenhuach21, line 26: it's possible to  use a register or some other way to append this automatically

yiliu30: For built‑in replacement modules, we need to add them manually to avoid importing them for all models. But for out‑of‑tree modules, there is no need to update them. Please check the newly added UT for details.

auto_round/modelling/replace_modules.py
```diff
@@ -0,0 +1,180 @@
...
+from abc import ABC, abstractmethod
+from typing import Dict, Type
+
+import torch
+from tqdm import tqdm
+from transformers import PreTrainedModel
+
+from auto_round.utils import LazyImport, logger
+
+BUILTIN_MODULES = {
```
yiliu30, line 24: Hi @WeiweiZhang1 , Please be aware of this change: for any newly added built‑in replacement module, we need to 1) implement the new module by inheriting from `ReplacementModuleBase`, and 2) add it to `BUILTIN_MODULES` manually.

yiliu30:
> If it's not targetd to 0.9.3, then please hold on the merge

It’s not targeting 0.9.3, but we’d like to merge it this week. 
Since we already have an RC branch for 0.9.3, would it be okay to merge this into main? cc @chensuyue  @thuang6

test/test_cuda/test_moe_model.py
```diff
@@ -103,3 +107,86 @@ def test_llama4(setup_llama4):
...
+        self.config = config
+        self.hidden_size = config.hidden_size
+        self.intermediate_size = config.intermediate_size
+        self.new_gate_proj = original.gate_proj
+        self.new_up_proj = original.up_proj
+        self.new_down_proj = original.down_proj
+        self.act_fn = nn.GELU()
+
+    def forward(self, x):
+        gate = self.new_gate_proj(x)
```
wenhuach21, line 141: I remember that there is a warning to remind user that transformers could run these models anymore. 
In this pr, this has been supported. Have we deleted this warning? please have a double check

yiliu30: I’m not aware of the background, could you point me to the code? thanks

wenhuach21: I couldn't find the warnings. Better run one model and double check the log if there are some warning like this.

yiliu30: did not see such suspicious warnings.
```bash
2025-12-23 05:58:48 WARNING modeling_utils.py L4793: `torch_dtype` is deprecated! Use `dtype` instead!
Converting model: 0it [00:00, ?it/s]2025-12-23 05:58:49 WARNING configuration_utils.py L430: `torch_dtype` is deprecated! Use `dtype` instead!
2025-12-23 05:58:50 WARNING mx.py L159: MXFP quantization is still in experimental stage, the inference speed might be slow.
2025-12-23 05:59:08 WARNING utils.py L2077: The attention mask is not set and cannot be inferred from input because pad token is same as eos token. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
(ao) yliu7@mlp-dgx-01:examples$ 

```

mengniwang95: the warning has been removed in this PR: https://github.com/intel/auto-round/pull/1067

auto_round/modelling/replace_modules.py
```diff
@@ -0,0 +1,180 @@
...
+    _import_required_replacements(model)
+    replaced = {}
+
+    # Step 1: Collect all modules that need replacement
+    logger.debug("Scanning for modules to replace")
+    modules_to_replace = []
+    for name, module in model.named_modules():
+        class_name = (
+            module.original_module_class()
+            if hasattr(module, "original_module_class") and callable(module.original_module_class)
```
mengniwang95: if the module has attrbute original_module_class, the module is already replaced. Why do we need to replace it again?

yiliu30: Good catch! corrected.

PR #1160, opened 2025-12-18
UT refactor for preview

xin3he:
- [x] Remove the duplicate LLMDataloader code and use the `dataloader` fixture to initialize it, ensuring it is initialized only once.
- [x] Remove the code related to unit testing and follow the Pytest coding style.
- [x] Add fixtures: `tiny_opt_model`(three layers), `opt_model`(for accurcay), `opt_tokenizer` for easy-of-use
- [x] Remove the duplicate model_infer function and move to `helper.py` for common usage.
- [x] Add README to introduce how to use `fixtures` and `helper` module.
- [x] Replace all UTs with fixture and helper function.

Co-authors: @n1ck-guo @XuehaoSun

### DIFF
--- a/test/test_cuda/test_gguf.py
+++ b/test/test_cuda/test_gguf.py
@@ -159,10 +159,10 @@ class TestAutoRound:
                 model_path = get_model_path(model_name)
                 tiny_model_path = "tmp_tiny_model"
                 tiny_model_path = save_tiny_model(model_path, tiny_model_path, num_layers=2)
-                ar = AutoRound(tiny_model_path, scheme=gguf_format, iters=0, nsampels=1, seqlen=16)
+                ar = AutoRound(tiny_model_path, scheme=gguf_format, iters=0, nsamples=1, seqlen=16)
                 ar.quantize_and_save(output_dir=self.save_dir, format=gguf_format)
 
-                ar = AutoRound(tiny_model_path, scheme=gguf_format, iters=1, nsampels=1, seqlen=16)
+                ar = AutoRound(tiny_model_path, scheme=gguf_format, iters=1, nsamples=1, seqlen=16)
                 ar.quantize_and_save(output_dir=self.save_dir, format=gguf_format)
 
                 shutil.rmtree(tiny_model_path, ignore_errors=True)


test/test_ark/test_model.py
```diff
@@ -1,70 +1,32 @@
 import shutil
-import sys
-import unittest
 
 import pytest
-
-sys.path.insert(0, "../..")
```
wenhuach21, line 6: Could we keep this? It would make debugging easier in the IDE.

test/test_cpu/test_alg_ext.py
```diff
@@ -1,37 +1,30 @@
...
-import sys
-import unittest
-
-from parameterized import parameterized
-
-sys.path.insert(0, "../..")
-
 from auto_round import AutoRound
 
+from ..helpers import opt_name_or_path, qwen_name_or_path
```
wenhuach21: how about providing a mapping dict or funcitons,  for qwen,  we need qwen-0.6, qwen-8B both

test/test_ark/test_model.py
xin3he: It's kept in confeest.py and works for all test files, only requiring trigger `pytest` under `test/` path

test/test_cpu/test_alg_ext.py
xin3he: @n1ck-guo Do you have any idea?

n1ck-guo: using get_model_path("qwen/Qwen3-0.6B")

.azure-pipelines/scripts/ut/run_ut_cuda.sh
```diff
@@ -27,16 +27,14 @@ function create_conda_env() {
 
     # install AutoRound
     cd ${REPO_PATH}
-    pip uninstall auto-round -y
+    uv pip install torch==2.8.0 torchvision
```
wenhuach21, line 30: It would be better to run most UTs on the latest version, while maintaining a specific environment for those that have known issues.

n1ck-guo (approved):

PR #1161, opened 2025-12-18
Fix code for llmc llama4 quantization

mengniwang95:
- LLMC uses quantize_block API to quantize block, where iters maybe 0, add process for dump_info in _quantize_block
- LLMC passes decoding layer into quantize_block, which doesn't have config attribute

### DIFF
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -656,8 +656,8 @@ class BaseCompressor(object):
                 "'enable_torch_compile' is set to `False` by default. "
                 "Enabling it can reduce tuning cost by 20%, but it might throw an exception.",
             )
-
-        if (self.data_type.startswith("fp") or self.act_data_type.startswith("fp")) and self.enable_torch_compile:
+        # On HPU, we rely on torch.compile to speed up the model execution.
+        if self.enable_torch_compile and is_wfp8afp8(self) and not is_hpex_available():
             self.enable_torch_compile = False
             logger.warning("reset enable_torch_compile to `False` as fp8 is enabled")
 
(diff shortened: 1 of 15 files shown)


yiliu30:
@XuehaoSun  @chensuyue Is it possible to cherry‑pick this into 0.9.3, so we can use it when we upgrade AutoRound on the LLMC side?
cc @thuang6

auto_round/compressors/base.py
```diff
@@ -2980,10 +2980,16 @@ def _quantize_block(
...
-        dump_info = (
-            f"quantized {len(quantized_layer_names)}/{(len(quantized_layer_names) + len(unquantized_layer_names))} "
-            f"layers in the block, loss iter 0: {init_loss:.6f} -> iter {best_iter}: {last_loss:.6f}"
-        )
+        if self.iters > 0:
+            dump_info = (
+                f"quantized {len(quantized_layer_names)}/{(len(quantized_layer_names) + len(unquantized_layer_names))} "
+                f"layers in the block, loss iter 0: {init_loss:.6f} -> iter {best_iter}: {last_loss:.6f}"
+            )
+        else:
```
xin3he, line 3011: Is it possible to use `_quantize_via_rtn_blockwise` for iters=0?

auto_round/compressors/mllm/compressor.py
```diff
@@ -200,7 +200,7 @@ def __init__(
         self.image_processor = image_processor
         from transformers import PreTrainedModel
 
-        if model.config.model_type == "llava" and isinstance(model, PreTrainedModel):
+        if isinstance(model, PreTrainedModel) and model.config.model_type == "llava":
```
wenhuach21, line 204: if order  matters, please leave a comment to let developer know why

auto_round/compressors/base.py
mengniwang95: _quantize_block is better.
LLMC quantizes model per decoding layer where there are more than 1 Linears. If we need to quantize activation like static FP8, _quantize_via_rtn_blockwise will cache data in itself, which will fail since the model in _quantize_via_rtn_blockwise is actually a block.

wenhuach21: decoding layer is equal to block, right? don't quite understand your point why it will fail

mengniwang95: _quantize_via_rtn_blockwise will call calib function, calib function use self.model for inference. If the self.model is actually a middle block, can we make sure the input is correct?

n1ck-guo (approved):

auto_round/compressors/base.py
mengniwang95: Have tried _quantize_via_rtn_blockwise locally. During calibration, AR cached ['positional_inputs', 'input_ids', 'attention_mask'] inputs which is the model input rather than block/layer's layer and it failed.

PR #1180, opened 2025-12-23
Add FP8 KV Support for DS

yiliu30:


auto_round_extension/vllm_ext/vllm_oot_patches.py
```diff
@@ -0,0 +1,119 @@
...
+            logger.warning_once(
+                "Found kv_scale in the checkpoint (e.g. %s), but not found the expected name in the model (e.g. %s). kv_scale is not loaded.",  #  noqa: E501
+                name,
+                remapped_name,
+            )
+            return None
+        return remapped_name
+
+    if any("mla_attn" in key for key in params_dict):
+        attn_str = "mla_attn.mla_attn"
```
mengniwang95, line 59: Are you sure the attn_str should be "mla_attn.mla_attn"? I noticed in vllm, even though the attention is MLA, the attn module name in decoder layer is still self_attn: https://github.com/vllm-project/vllm/blob/ba25a6599212f01e6b877e22549612333e95b2c1/vllm/model_executor/models/deepseek_v2.py#L1151

yiliu30: Yes, please check the code here, https://github.com/vllm-project/vllm/blob/ba25a6599212f01e6b877e22549612333e95b2c1/vllm/model_executor/models/deepseek_v2.py#L1085.
Attached modeling FYI
```bash
self.model.model.layers[0].self_attn.mla_attn.mla_attn
MLAAttention(
  (impl): HPUMLAImpl(
    (kv_b_proj): ColumnParallelLinear(in_features=512, output_features=1024, bias=False, tp_size=1, gather_output=False)
    (matmul_qk): FP8Matmul()
    (softmax): Softmax()
    (matmul_av): FP8Matmul()
    (batch2block_matmul): FP8Matmul()
    (block2batch_matmul): FP8Matmul()
    (latent_cache_k): VLLMFP8KVCache()
    (fused_scaled_dot_product_attention): ModuleFusedSDPA()
  )
)

```

PR #1182, opened 2025-12-23
Enable qwen3 vl moe quant and load

WeiweiZhang1:


auto_round/modelling/qwen3_vl_moe.py
```diff
@@ -0,0 +1,149 @@
...
+    experts.
+    """
+
+    is_permanent = True
+
+    def __init__(
+        self,
+        original: "Qwen3VLMoeTextSparseMoeBlock",
+        config: "Qwen3VLMoeConfig",
+        calibrate_all_experts: bool = True,
```
yiliu30: It would be better to keep the original behavior until we enable that feature.
```suggestion
        calibrate_all_experts: bool = False,
```

auto_round/modelling/qwen3_vl_moe.py
```diff
@@ -0,0 +1,149 @@
...
+                weighted_output = expert_out * routing_weights[token_idx, idx, None]
+                next_states.index_add_(0, token_idx, weighted_output.to(hidden_states.dtype))
+        next_states = next_states.reshape(batch_size, sequence_length, hidden_dim)
+
+        if transformers_version <= version.parse("4.57.3"):
+            return next_states, router_logits
+        else:
+            return next_states
+
+    def restore(self, original: torch.nn.Module) -> torch.nn.Module:
```
yiliu30:
```suggestion

```

auto_round/experimental/qmodules/fp4_utils.py
```diff
@@ -31,7 +31,18 @@
 
 import torch
 
-kE2M1ToFloat = torch.tensor([0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0], dtype=torch.float32)
+_DEVICE_E2M1_TENSORS = {}
```
n1ck-guo, line 34: Are default values ​​not needed?

WeiweiZhang1: Tensor-type global variables may lead to device-related issues in future Transformers releases (e.g., v5.0.0); therefore, they are initialized lazily upon first use.

PR #1217, opened 2026-01-04
fix torch dtype

wenhuach21:


xin3he (approved):
This `monkey_patch_transformers` is really cool!

auto_round/utils/common.py
```diff
@@ -73,6 +74,51 @@ def __call__(self, *args, **kwargs):
...
+                    kwargs[new_name] = kwargs.pop(old_name)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+# TODO this is not very robust as only AutoModelForCausaLM is patched
+def monkey_patch_transformers():
```
yiliu30, line 94: nit: It might be better to add ` @lru_cache(None)` so we don’t patch it multiple times.

test/test_ark/test_model.py
```diff
@@ -76,6 +76,10 @@ def test_awq_fp16(self, format, bits, group_size, sym, dtype, device):
     # def test_other_bits(self, format, bits, group_size, sym, dtype, device):
     #     self.main_op(format, bits, group_size, sym, dtype, device, False, 0.2)
 
+    # TODO all the above tests are skipped, add a dummy test to make sure the file is collected
```
chensuyue: I have disabled this UT workflow since cpu path is closed. So you can choose to remove those lines.

PR #1226, opened 2026-01-06
gguf add support for Mistral-Magistral-Devstral and Granite

n1ck-guo:


auto_round/export/export_to_gguf/convert.py
```diff
@@ -229,33 +249,20 @@ def _quant_data(cls, data_torch, data_qtype, name, modify_name, bid, device=None
...
-                    zp,
-                    wmin=wmin,
-                    d_scale=d_scale,
-                    d_wmin=d_wmin,
-                    imatrix=imatrix,
-                    device=device,
-                )
-            else:
-                data = ggml_quant(data_torch, data_qtype.name.lower(), scale, zp, device=device)
+                kwargs["imatrix"] = module.imatrix.to(torch.float32) if hasattr(module, "imatrix") else None
```
wenhuach21: it seems L257 has already done this, right? if not, other keys are also need to code like this

wenhuach21 (approved):

auto_round/export/export_to_gguf/convert.py
```diff
@@ -248,19 +248,15 @@ def _quant_data(cls, data_torch, data_qtype, name, modify_name, new_name, bid, d
...
-                        ori_shape = attr_tensor.shape
                         attr_tensors_dict = dict(cls.modify_tensors(attr_tensor.reshape(bs, -1), modify_name, bid))
                         attr_tensor = attr_tensors_dict[new_name]
-                        # attr_tensor = attr_tensor.reshape(-1, ori_shape[-1])
-                        # attr_tensor = attr_tensor.reshape(ori_shape)
-                        # setattr(module, attr, attr_tensor)
                         if attr in kwargs:
-                            kwargs[attr] = attr_tensor.to(torch.float32)
+                            if attr != "imatrix":
+                                attr_tensor = attr_tensor.to(torch.float32)
```
wenhuach21: may I know the reason why only imatrix cannot be converted to float32

auto_round/export/export_to_gguf/convert.py
```diff
@@ -248,19 +248,15 @@ def _quant_data(cls, data_torch, data_qtype, name, modify_name, new_name, bid, d
...
+                            kwargs[attr] = attr_tensor
                         else:
                             kwargs[attr.replace("w_", "")] = attr_tensor
             data_torch = data_torch.to(torch.float32)
-            if data_qtype.name.lower().endswith("_k"):
-                kwargs["imatrix"] = module.imatrix.to(torch.float32) if hasattr(module, "imatrix") else None
 
             data = ggml_quant(data_torch, data_qtype.name.lower(), device=device, **kwargs)
```
wenhuach21, line 259: if imatrix must be half or bf16, maybe we'd better also have a checker and convert the dtype in ggml_quant

auto_round/export/export_to_gguf/convert.py
n1ck-guo: my bad, I misread the original code, all tensor should be torch.float32

PR #1228, opened 2026-01-07
Add FP8 dtype info to quant config

yiliu30:
Resolve https://github.com/vllm-project/vllm-gaudi/pull/763#issuecomment-3715570920

cc @a32543254


yiliu30:
Hi @thuang6, I hope this change helps address your concern as well.

PR #1229, opened 2026-01-07
Remove failed UT since autoround:awq is entering the depracated phase

xin3he:


wenhuach21:
we'd better keep it. For xpu, ark could support the inference while you could delete the cuda ut related to awq

PR #1231, opened 2026-01-07
set disable_opt_rtn to optional bool and change default value  to None

WeiweiZhang1:


requirements.txt
```diff
@@ -1,10 +1,4 @@
 # 1.5.1<accelerate<1.10.0 may cause potentially high RAM usage; versions 1.10.0 or above are recommended.
 accelerate
 datasets
-numpy
-# packaging # for python version <= 3.9
-py-cpuinfo
-threadpoolctl
-torch
-tqdm
```
wenhuach21, line 9: ?

auto_round/autoround.py
```diff
@@ -87,7 +87,7 @@ def __new__(
         # for MLLM and Diffusion
         extra_config: ExtraConfig = None,
         enable_alg_ext: bool = None,
-        disable_opt_rtn: bool = None,
+        disable_opt_rtn: Optional[bool] = None,
```
wenhuach21, line 88: no need to add Optional

auto_round/compressors/base.py
```diff
@@ -388,6 +388,9 @@ def __init__(
             logger.warning(
                 "for bits <= 2, it is recommended to enable `auto-round-best` " "and turn on `--enable_alg_ext` "
             )
```
wenhuach21, line 390: add this before L391
if disable_opt_rtn is none:
   disable_otp_rtn=False

auto_round/compressors/base.py
```diff
@@ -388,6 +388,9 @@ def __init__(
...
+        if self.bits == 8 and self.iters == 0 and self.data_type == "int" and disable_opt_rtn is None:
```
wenhuach21: and act_bits>=16

wenhuach21: self.bits >= 8

auto_round/autoround.py
WeiweiZhang1: Without Optional, the code executes correctly, but the semantics are imprecise.

auto_round/compressors/config.py
```diff
@@ -32,7 +32,7 @@ def __init__(
         self,
         # tuning
         amp: bool = True,
-        disable_opt_rtn: bool = True,
+        disable_opt_rtn: Optional[bool] = True,
```
wenhuach21, line 35: why is it True here?

requirements.txt
```diff
@@ -2,7 +2,6 @@
 accelerate
 datasets
 numpy
-# packaging # for python version <= 3.9
```
wenhuach21, line 5: revert

README.md
```diff
@@ -195,7 +195,7 @@ ar.quantize_and_save(output_dir="./qmodel", format="auto_round")
 
 ##### Algorithm Settings
 - **`enable_alg_ext` (bool)**: [Experimental Feature] Only for `iters>0`. Enable algorithm variants for specific schemes (e.g., MXFP4/W2A16) that could bring notable improvements. Default is `False`.
-- **`disable_opt_rtn` (bool)**: Use pure RTN mode for specific schemes (e.g., GGUF and WOQ). Default is `False` (improved RTN enabled).
+- **`disable_opt_rtn` (bool)**: Use pure RTN mode for specific schemes (e.g., GGUF and WOQ). Default is `None` (improved RTN enabled).
```
wenhuach21, line 198: If None, it defaults to True in most cases, but may be disabled due to known issues

auto_round/compressors/base.py
```diff
@@ -389,6 +389,19 @@ def __init__(
...
+            self.bits >= 8
+            and self.act_bits >= 16
+            and self.iters == 0
+            and self.data_type == "int"
+            and disable_opt_rtn is None
+        ):
+            logger.warning("for INT8 RTN quantization, set `--disable_opt_rtn` as default.")
+            disable_opt_rtn = True
+        if disable_opt_rtn is None:
+            disable_otp_rtn = False
```
wenhuach21, line 403: logger some info here as well

requirements.txt
WeiweiZhang1: AR does not support python version <= 3.9. This could mislead users. After confirming with Xuehao, it can be removed.

wenhuach21: ok, thanks

auto_round/compressors/config.py
WeiweiZhang1: in config.py and gen_auto_scheme.py some func have a default disable_opt_rtn value of True. To maintain consistent behavior, I have not modified this value.

auto_round/compressors/base.py
```diff
@@ -389,6 +389,19 @@ def __init__(
...
+        if (
+            self.bits >= 8
+            and self.act_bits >= 16
+            and self.iters == 0
+            and self.data_type == "int"
+            and disable_opt_rtn is None
+        ):
+            logger.warning("for INT8 RTN quantization, set `--disable_opt_rtn` as default.")
+            disable_opt_rtn = True
+        if disable_opt_rtn is None:
```
yiliu30, line 402: Please add comments here to explain why we reset `disable_opt_rtn ` to `True`.

auto_round/__main__.py
```diff
@@ -255,7 +255,9 @@ def __init__(self, *args, **kwargs):
         )
         tuning.add_argument(
             "--disable_opt_rtn",
-            action="store_true",
+            "--disable-opt-rtn",
```
yiliu30, line 258: duplicate `--disable-opt-rtn`
```suggestion

```

README.md
WeiweiZhang1: you mean when it's None,  will automatically set ‘disable_opt_rtn’ to False in most cases, right?

better refine like this to avoid ambiguity:
The default is None, which usually enables improved RTN, but may disable it in the presence of known issues.

PR #1245, opened 2026-01-08
WNA16  does not apply optimized RTN for moe layers by default

wenhuach21:


auto_round/utils/model.py
```diff
@@ -1191,6 +1191,17 @@ def mv_module_from_gpu(module):
         return module.to("cpu")
 
 
+def is_moe_model(model: torch.nn.Module) -> bool:
```
yiliu30, line 1194: We already have such function:
https://github.com/intel/auto-round/blob/cc521cab9776b2519efcdb3f0245047583407486/auto_round/utils/model.py#L601

yiliu30 (approved):
Others LGTM.

PR #1254, opened 2026-01-10
(feat): add support for g2 fp8 on cpu

### LINKED ISSUE #1196: Deuqantize block-wise FP8 weight on CPU (G2-Specific)

SwekeR-463:
Fixes #1196 

- ~~Implemented manual dequant fallback for g2-fp8 format using a lookup table (LUT)~~.
- Added test to check for range limits and bit-exact dequant values.

### DIFF
--- a/auto_round/utils/model.py
+++ b/auto_round/utils/model.py
@@ -911,7 +911,7 @@ def pad_block_fp8_weight_naive(
 
 class with_thread_limits(ContextDecorator):
     """
-    Context manager and decorator to temporarily set OMP_NUM_THREADS and PyTorch threads.
+    Context manager and decorator to temporarily set AR_OMP_NUM_THREADS and PyTorch threads.
     Inspired by vLLM's thread limit decorator.
     https://github.com/HabanaAI/vllm-fork/blob/f943a89a20e0e57bca64e1cca05469bfcaaec6f8/vllm/worker/hpu_model_runner.py#L1063-L1115
     """
@@ -923,7 +923,7 @@ class with_thread_limits(ContextDecorator):
 
     def __enter__(self):
         # Save original settings
-        self.old_omp = envs.OMP_NUM_THREADS
+        self.old_omp = envs.AR_OMP_NUM_THREADS
         self.old_torch = torch.get_num_threads()
 
         try:
@@ -936,7 +936,7 @@ class with_thread_limits(ContextDecorator):
 
             # Set new limits
             new_threads = max(1, num_cores // self.div)
-            os.environ["OMP_NUM_THREADS"] = str(new_threads)
+            os.environ["AR_OMP_NUM_THREADS"] = str(new_threads)
             torch.set_num_threads(new_threads)
         except Exception:
             pass
@@ -945,9 +945,9 @@ class with_thread_limits(ContextDecorator):
     def __exit__(self, exc_type, exc_val, exc_tb):
         # Restore original settings
         if self.old_omp is None:
-            os.environ.pop("OMP_NUM_THREADS", None)
+            os.environ.pop("AR_OMP_NUM_THREADS", None)
         else:
-            os.environ["OMP_NUM_THREADS"] = self.old_omp
+            os.environ["AR_OMP_NUM_THREADS"] = self.old_omp
         torch.set_num_threads(self.old_torch)
 
 


auto_round/utils/model.py
```diff
@@ -906,12 +906,84 @@ def pad_block_fp8_weight_naive(
...
+    orig_shape = weight_uint8.shape
+
+    weight_flat = weight_uint8.flatten().long()
+    dequant_flat = lut[weight_flat]
+    dequant = dequant_flat.reshape(orig_shape)
+
+    return dequant
+
+
+def dequant_block_fp8_weight(
```
yiliu30, line 955: Hey @SwekeR-463 sorry for the slow response, I’ve had a lot on my plate.
I think there might be some misunderstanding about that feature. Was this part enabled for `e4m3fnuz` as input?
What we actually want is to handle the model with `e4m3fn`, then dequantize it to bf16, and finally requantize it to the target dtype.

SwekeR-463: Yes, misunderstanding happened from my end.
Would love if more input is given else can make changes with respect to dequantize `e4m3fn` to `bf16` then requantize to target dtype

yiliu30: Yeah, please dequantize `e4m3fn` to `bf16`. Thanks!

SwekeR-463: done

auto_round/utils/model.py
```diff
@@ -906,12 +909,162 @@ def pad_block_fp8_weight_naive(
...
+    weight: torch.Tensor, weight_scale: torch.Tensor, block_size: list, data_type: str = None
+) -> torch.Tensor:
     dtype = torch.bfloat16
     if weight_scale is None:
         return weight
     assert len(block_size) == 2
 
+    variant = "e4m3fn"
+    if data_type is not None and "e4m3fnuz" in data_type.lower():
```
yiliu30: Hi @SwekeR-463, we can assume the input is always e4m3fn for now.

auto_round/utils/model.py
```diff
@@ -906,12 +909,162 @@ def pad_block_fp8_weight_naive(
...
+        variant = "e4m3fnuz"
+    elif hasattr(weight, "dtype") and hasattr(torch, "float8_e4m3fnuz") and weight.dtype == torch.float8_e4m3fnuz:
+        variant = "e4m3fnuz"
+
+    # Automatic fallback for Gaudi2: dequantize on CPU to avoid HPU NaN issues for values > 240
+    from auto_round.utils.device import is_gaudi2
+
+    force_cpu = is_gaudi2() and weight.device.type == "hpu"
```
yiliu30: And the weight device should be `cpu` if we handle such a case in HPU.

yiliu30: I think we can resolve it by refining the dequant_block_fp8_weight function a bit.

```python


def _dequant_block_fp8_weight(weight: torch.Tensor, weight_scale: torch.Tensor, block_size: list) -> torch.Tensor:
     ...


@with_thread_limits()
def _dequant_block_fp8_weight_cpu(...):
    return _dequant_block_fp8_weight(...)

def dequant_block_fp8_weight(...):
    if is_gaudi2():
        _dequant_block_fp8_weight_cpu(...)
    else:
        _dequant_block_fp8_weight(...)
```

SwekeR-463: here then I have to make the `with_thread_limits` func a decorator?

yiliu30: Yeah, I think both a decorator and a context manager would work well here. Using a decorator just feels a bit more straightforward.

SwekeR-463: have done the changes

yiliu30: Thanks @SwekeR-463, We can follow the same dequantization behavior as CUDA, just make sure the device is CPU. My understanding is that a lookup table is unnecessary, please let me know if you have a different view.
The initial `e4m3fnuz` may have been a bit confusing you.

SwekeR-463: Yes actually the `e4m3fnuz` lead to a different direction of approach all together.





> We can follow the same dequantization behavior as CUDA, just make sure the device is CPU

And for this basically have to use native torch casting instead of LUT or something else?

yiliu30: Hi @SwekeR-463 , yes, how about reusing the `dequant_block_fp8_weight`, but forcing the device of  `convert_fp8_layer_to_linear` to CPU for gaudi 2.
https://github.com/intel/auto-round/blob/57d1a8acfbe98ee55330caedbfb47bdc2a83276f/auto_round/utils/model.py#L909

https://github.com/intel/auto-round/blob/57d1a8acfbe98ee55330caedbfb47bdc2a83276f/auto_round/utils/model.py#L1001-L1003

SwekeR-463: done with the changes

auto_round/envs.py
```diff
@@ -29,6 +29,7 @@
     "AR_WORK_SPACE": lambda: os.getenv("AR_WORK_SPACE", "ar_work_space").lower(),
     "AR_ENABLE_UNIFY_MOE_INPUT_SCALE": lambda: os.getenv("AR_ENABLE_UNIFY_MOE_INPUT_SCALE", "False").lower()
     in ["1", "true"],
+    "OMP_NUM_THREADS": lambda: os.getenv("OMP_NUM_THREADS", None),
```
yiliu30: nit: `AR_OMP_NUM_THREADS`

SwekeR-463: done 👍🏻

auto_round/utils/model.py
```diff
@@ -906,16 +909,65 @@ def pad_block_fp8_weight_naive(
...
+    """
+
+    def __init__(self, div=1):
+        self.div = div
+        self.old_omp = None
+        self.old_torch = None
+
+    def __enter__(self):
+        # Save original settings
+        self.old_omp = envs.OMP_NUM_THREADS
```
yiliu30: Hi @SwekeR-463 , please update the name here as well, thanks!

SwekeR-463: My bad, missed these changes :).

PR #1261, opened 2026-01-12
fix XPU CI hang issue

xin3he:


test/test_xpu/test_autoround.py
```diff
@@ -75,11 +76,10 @@ def test_awq_format(self, dataloader):
         quantized_model_path = "./saved"
         autoround.quantize_and_save(output_dir=quantized_model_path, format="auto_round:auto_awq")
 
-        quantized_model_path = "./saved"
         quantization_config = AutoRoundConfig(backend="auto")
-
+        # device_map="auto" doesn't work, must use "xpu"
         model = AutoModelForCausalLM.from_pretrained(
-            quantized_model_path, device_map="auto", quantization_config=quantization_config
+            quantized_model_path, device_map=self.device, quantization_config=quantization_config
```
wenhuach21, line 82: Do we know the root cause? Or have you verified that the BF16 model is fine or not?
Maybe it's a bug of AR

xin3he: I notice that the model is on xpu:3 with `device_map='auto'` for repacking, it's abnormal and not related with AutoRound.

Issue #1212, opened 2026-01-02
replace modules causes  large ram usage


wenhuach21:
https://github.com/intel/auto-round/issues/1211

yiliu30:
Hi @wenhuach21 

I synced with @n1ck-guo and @WeiweiZhang1, and we discussed two potential approaches to mitigate this issue.

#### Method 1: Replace blocks one by one before the quantization process
This approach is straightforward and helps avoid materializing the entire model in RAM before the first quantization step.
However, our current workflow retrieves `layer_config`, `block names`, and other info at the model level before quantization starts. 
https://github.com/intel/auto-round/blob/c24490596706145fd2d32ca4ffd88b8833836a48/auto_round/compressors/base.py#L1422-L1478
Supporting block-level replacement would therefore require a significant restructuring of the code. @n1ck-guo  is currently working on the refactoring effort, and this approach will be considered as part of that work.

#### Method 2: Delay materializing the replaced model weights until the first forward pass or the first quantization step
This approach allows us to keep most of the current design and only adjust the weight replacement logic. The trade-off is increased implementation complexity of the replacement module, especially for out-of-tree replacements, where it may become harder for users to customize modules. For in-tree replacements, however, we can handle the logic internally within AutoRound.

Based on this discussion, we plan to proceed with Method 2 so that we can address urgent cases such as LLaMA 4 and DeepSeek. Once Method 1 becomes feasible, we could deprecate Method 2 or keep it, and it won't introduces additional RAM/VRAM overhead.

Additionally, @WeiweiZhang1 identified an elegant solution to using `to_empty` to release the RAM of the original module. Before that, we need to carefully clear all references and let Python GC recycle them. We will verify this on other models. This solution can complement both approaches and reduce memory usage at the block level.

cc @thuang6 @xinhe3

wenhuach21:
Thanks for the clarification.  Because some models have official FP8 version and we claimed that we supported FP8 model as input, please also try to support this, if not, an exception needs to be raised.

wenhuach21:
2 please be aware of that for vlm and quantized lm-head, the calibration phase will call forward functions of  most modules.

PR #1265, opened 2026-01-13
refine moe modellings to release orginal expert module's ram

WeiweiZhang1:
local moe model tests are done

for Qwen3-VL-Moe 30B model
before
image

after
image

### DIFF
--- a/.azure-pipelines/docker/Dockerfile.devel
+++ b/.azure-pipelines/docker/Dockerfile.devel
@@ -43,7 +43,6 @@ USER hostuser
 ENV PATH="/home/hostuser/.venv/bin:$PATH"
 ENV VIRTUAL_ENV="/home/hostuser/.venv"
 ENV UV_NO_PROGRESS=1 \
-    UV_COMPILE_BYTECODE=1 \
     UV_LINK_MODE=copy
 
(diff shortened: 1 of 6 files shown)


auto_round/modelling/qwen3_vl_moe.py
```diff
@@ -143,7 +143,9 @@ def __init__(self, config, original):
             Qwen3VLMoeTextMLP,
         )
 
-        super().__init__([Qwen3VLMoeTextMLP(config, intermediate_size) for _ in range(self.num_experts)])
+        target_device = next(original.parameters()).device
```
wenhuach21, line 146: if the device_map is auto, there are multiple devices and some hooks also needs to be copied, I guess

yiliu30: Good catch! Given that the `target_device` is usually CPU, I think it’s fine to leave it as is. 
I’ll check whether we can rebind these hooks in the `ReplacementModuleBase` class.

WeiweiZhang1: Close this PR for now. The issue (https://github.com/intel/auto-round/issues/1212
) will remain open until block-wise module replacement is enabled.

PR #1287, opened 2026-01-15
fix bug of format

n1ck-guo:


auto_round/compressors/base.py
```diff
@@ -2966,7 +2968,7 @@ def save_quantized(
             object: The compressed model object.
         """
         self.orig_output_dir = output_dir
-        if isinstance(format, str):
+        if isinstance(format, str) and getattr(self, "formats", None) is None:
```
wenhuach21, line 2971: add ut

n1ck-guo: there is ut to check this.

PR #1288, opened 2026-01-15
fix low_cpu and set it to True as default

wenhuach21:
test_lm_head quant

### DIFF
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -1470,7 +1470,7 @@ class BaseCompressor(object):
         )
 
     def _adjust_immediate_packing_and_saving(self):
-        formats = get_attr(self, "formats", [])
+        formats = getattr(self, "formats", [])
         if len(formats) == 1 and not formats[0].is_fake() and self.inplace:
             self.is_immediate_packing = True
 


auto_round/compressors/shard_writer.py
```diff
@@ -0,0 +1,230 @@
...
+        self.lm_head_name = get_lm_head_name(self.model)
+        total_params = sum(p.numel() for p in self.model.parameters())
+        # Heuristic estimate of model size in GB used to choose a default max_shard_size:
+        # - total_params * rounder.bits       -> total number of bits in all parameters
+        # - // 8                              -> convert bits to bytes
+        # - // 1e9                            -> approx convert bytes to GB (1e9 bytes ~= 1 GB)
+        # - final // 10                       -> apply a safety margin so default shards are
+        #                                         smaller than the full model; this intentionally
+        #                                         underestimates size before clamping below.
+        model_size = int(total_params * rounder.bits // 1e9 // 8) // 10
```
Kaihui-intel, line 41: Suggest aligning the following ```units={"GB": 1024 * * 3, "MB": 1024 * * 2, "KB": 1024, "B": 1}``` 1e9 -> 1024**3

auto_round/compressors/shard_writer.py
```diff
@@ -0,0 +1,230 @@
...
+        sd = m.state_dict()
+
+        for k, v in sd.items():
+            if not isinstance(v, torch.Tensor):
+                continue
+            param_name = f"{prefix}.{k}"
+            self._add_tensor(param_name, v)
+
+    def _add_tensor(self, name: str, tensor: torch.Tensor):
+        t_size = tensor.numel() * tensor.element_size()
```
Kaihui-intel:
```suggestion
        t_size = tensor.nbytes
```

auto_round/compressors/shard_writer.py
```diff
@@ -0,0 +1,230 @@
...
+        # Using a set for faster lookup of all saved parameters
+        all_saved = {p for meta in self.shard_meta for p in meta["params"]}
+
+        for param_full_name in saved_params:
+            module_path = param_full_name.rsplit(".", 1)[0]
+            try:
+                module = get_module(self.model, module_path)
+                # Check if all parameters of this module are now in 'all_saved'
+                if all(f"{module_path}.{k}" in all_saved for k in module.state_dict().keys()):
+                    module.to("meta")
```
Kaihui-intel: When is clear_cemory() called?

Kaihui-intel (approved):
Looks Good

auto_round/compressors/base.py
```diff
@@ -1272,21 +1267,30 @@ def _quantize_rtn(self) -> tuple[torch.nn.Module, dict[str, Any]]:
...
+                if hasattr(m, "global_name") and m.global_name in all_to_quantized_module_names:
+                    pbar.set_description(f"Quantizing {m.global_name}")
+                    self._quantize_layer_via_rtn(m.global_name)
+                    cnt += 1
+                    pbar.update()
+                    if cnt % clear_mem_freq == 0:
+                        clear_memory(device_list=self.device_list)
+                        memory_monitor.log_summary()
+                # A workaround to trigger garbage collection in time
+                elif not any(m.children()) and len(m.state_dict()) > 0:
```
yiliu30: Maybe we could refactor this part to be block-wise and clear the memory after each block finished.

auto_round/utils/device.py
```diff
@@ -463,16 +463,29 @@ def _clear_memory_for_cpu_and_cuda(
...
+            _clear_memory_for_cpu_and_cuda(tensor, final_device_list)
 
-    memory_monitor.update(device_list=device_list)
-    if is_hpex_available():
-        # hpu does not have empty_cache
-        return
-    else:
-        _clear_memory_for_cpu_and_cuda(tensor, device_list)
+
+clear_memory = torch._dynamo.disable()(ClearMemory(device_list=[0]))
```
yiliu30, line 488: Should the default value of `device_list` be None? If we set it to `[0]`, does that assume we only have one GPU?

auto_round/compressors/shard_writer.py
```diff
@@ -0,0 +1,230 @@
...
+            except Exception:
+                continue
```
yiliu30: Nit: please log the error instead of staying silent.

auto_round/utils/device.py
wenhuach21: to save the time cost, developers should pass the device list, otherwise, only card 0 or the last used devices list is cleared

auto_round/compressors/base.py
wenhuach21: already have similar logic

auto_round/compressors/shard_writer.py
wenhuach21: in the base compressor.py

Issue #1283, opened 2026-01-15
`ignore_layers` does not take effect on FP8 model


yiliu30:
Description:
When running quantization on an FP8 model with the `--ignore_layers` argument, the specified layers are not ignored as expected. The log shows `not_to_quantized_layers: []`, indicating the argument is not being processed or respected for FP8 models.

```bash
AR_LOG_LEVEL=TRACE auto_round --model /models/Qwen3-8B-FP8 --ignore_layers "mlp"
```

```bash
Loading checkpoint shards: 100%|████████████████████████| 2/2 [00:00<00:00, 16.29it/s]
2026-01-14 22:03:32 WARNING model.py L323: the support for fp8 model as input is experimental, please use with caution.
2026-01-14 22:03:32 INFO base.py L442: using torch.bfloat16 for quantization tuning
2026-01-14 22:03:32 INFO base.py L708: 'enable_torch_compile' is set to `False` by default. Enabling it can reduce tuning cost by 20%, but it might throw an exception.
2026-01-14 22:03:32 DEBUG replace_modules.py L152: Scanning for modules to replace
2026-01-14 22:03:32 DEBUG replace_modules.py L174: No modules found for replacement
2026-01-14 22:03:32 TRACE utils.py L889: not_to_quantized_layers: [] <--------------------
```

PR #1289, opened 2026-01-15
Robust FP8 layer detection for ignore_layers (#1283)

scopophobic:
FP8 layers were not detected by get_fp_layer_names, causing ignore_layers
to be ignored. This PR:
- Auto-detects FP8 layers
- Includes them in not_to_quantized_layers
- Ensures ignore_layers works for FP8 models
  

Signed-off-by: Adithyan Madhu <adithyanworkmail@gmail.com>

### DIFF
--- a/auto_round/compressors/utils.py
+++ b/auto_round/compressors/utils.py
@@ -865,17 +865,16 @@ def get_fp_layer_names(model: torch.nn.Module, ignore_layers: str):
         subcomponents of those layers.
     """
     from auto_round.utils import SUPPORTED_LAYER_TYPES
-    
+
     not_to_quantized_layers = []
 
-    for n,m in model.name_modules():
+    for n, m in model.name_modules():
         if is_fp8_linear(m):
             not_to_quantized_layers.append(n)
             logger.trace(f"Auto-detected FP8 layer to ignore : {n}")
-    
 
     if ignore_layers:
-        ignore_list = ignore_layers.replace(" ","").split(",")
+        ignore_list = ignore_layers.replace(" ", "").split(",")
         for fp_layer in ignore_list:
             if not fp_layer:
                 continue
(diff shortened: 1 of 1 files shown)


auto_round/compressors/utils.py
```diff
@@ -866,8 +866,30 @@ def get_fp_layer_names(model: torch.nn.Module, ignore_layers: str):
...
-        return []
+    not_to_quantized_layers = []
+
+    for n, m in model.name_modules():
+        if is_fp8_linear(m):
+            not_to_quantized_layers.append(n)
+            logger.trace(f"Auto-detected FP8 layer to ignore : {n}")
+
+    if ignore_layers:
+        ignore_list = ignore_layers.replace(" ", "").split(",")
```
yiliu30, line 877: Hi @scopophobic, thanks for your interest in fix that issue! I think there might be a bit of misunderstanding.
We don’t want to skip all FP8 layers. The idea is that we start with an FP8 model and want to requantize it to another format, like W4A16. However, we don’t want certain layers—such as those inside the attention module—to be quantized to W4A16.
The fix here is aligned with what we’re aiming for. https://github.com/intel/auto-round/pull/1286

yiliu30: Hi @scopophobic   Would you be interested in working on the left part of this issue? https://github.com/intel/auto-round/issues/1283#issuecomment-3759471170

scopophobic: Hi @yiliu30, thanks a lot for the clarification, that helped resolve a misunderstanding I had 👍

I now understand that the goal is not to skip all FP8 layers, but to start from an FP8 model and re-quantize it (e.g., to W4A16), while keeping specific submodules (like attention) from being quantized.

I’m definitely interested in working on the remaining part of #1283. My current thought is to make FP8 detection more robust by moving away from class-name checks (like "FP8Linear") and instead relying on explicit FP8 characteristics (e.g., presence of FP8 scale metadata used during dequantization). This would allow supporting multiple FP8 layer implementations without brittle heuristics.

Does this approach sound aligned with what you had in mind for this issue?

yiliu30:
Close it in favor of https://github.com/intel/auto-round/pull/1348

PR #1297, opened 2026-01-19
Shorten CI time by splitting up time-consuming tasks

xin3he:


test/test_cpu/export/test_gguf/test_gguf_format.py
```diff
@@ -0,0 +1,116 @@
...
+        autoround = AutoRound(
+            self.model_name,
+            bits=bits,
+            group_size=group_size,
+            sym=sym,
+            iters=1,
+            data_type="int",
+            nsamples=1,
+            seqlen=8,
+        )
```
wenhuach21, line 56: @n1ck-guo  use disable_opt_rtn if possible and only use opt-rtn for limited scenarios

.azure-pipelines/scripts/ut/run_ut.sh
```diff
@@ -35,7 +35,7 @@ ut_log_name=${LOG_DIR}/ut.log
 # Split test files into 5 parts
 find ./test_cpu -name "test*.py" | sort > all_tests.txt
 total_lines=$(wc -l < all_tests.txt)
-NUM_CHUNKS=5
+NUM_CHUNKS=8
```
XuehaoSun, line 38: Do not change here.

xin3he: Okay, in my theory, chunks=8 might provide a better experience with well splitted UTs. Do you have interest to have a try? If no, we can close this PR I think.

xin3he:
close it

.azure-pipelines/scripts/ut/run_ut.sh
XuehaoSun: Because the parallel code is not here.

PR #1311, opened 2026-01-21
align act_max of experts for qwen3-vl and qwen3-next

xin3he:


auto_round/utils/model.py
```diff
@@ -613,6 +613,8 @@ def is_moe_layer(module: torch.nn.Module) -> bool:
             "DeepseekV3MoE".lower(),
             "Qwen2MoeSparseMoeBlock".lower(),
             "Qwen3MoeSparseMoeBlock".lower(),
+            "Qwen3NextSparseMoeBlock".lower(),
+            "Qwen3VLMoeTextSparseMoeBlock".lower(),
```
WeiweiZhang1, line 617: If this change is intended to support NVFP quantization for Qwen3VLMoE&Qwen3Next, please also verify VLLM generation, thanks!

xin3he: Sure, this PR is necessary since some experts are not calibrated and act_max is missing.

Issue #1247, opened 2026-01-08
format scheme checkers bug


wenhuach21:
https://github.com/intel/auto-round/blob/main/auto_round/formats.py#L436
1 --scheme W8A16 --bits 4
2 --scheme W4A16  --group_size 256

PR #1332, opened 2026-01-23
format add check for scheme args

n1ck-guo:
## Description

In the previous code, the format function only handles cases where the scheme is a string. It does not perform checks for scenarios like --scheme w2a16 --bits 4. This PR adds the relevant checks.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #1247 

### DIFF
--- a/auto_round/formats.py
+++ b/auto_round/formats.py
@@ -362,7 +362,7 @@ class LLMCompressorFormat(OutputFormat):
     @classmethod
     def check_scheme_args(cls: OutputFormat, scheme: QuantizationScheme) -> bool:
         error_logs = []
-        if scheme.bits not in [4, 8]:
+        if scheme.bits not in [4, 8, 16]:
             error_logs.append(f"bits={scheme.bits}")
         if not re.search("mxfp|fp|nvfp", scheme.data_type):
             error_logs.append(f"data_type={scheme.data_type}")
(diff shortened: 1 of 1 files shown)


auto_round/formats.py
```diff
@@ -355,6 +359,26 @@ def __init__(self, format, ar):
             self.output_format = f"llm_compressor:{format}"
             self.backend = None
 
+    @classmethod
+    def check_scheme_args(cls: OutputFormat, scheme: QuantizationScheme) -> bool:
+        error_logs = []
+        if scheme.bits not in [4, 8]:
```
wenhuach21: what is the scheme, the default scheme or the layer-wise scheme, if it's for layer-wise, I think most formats support 16 bits, so we'd better add 16 bits in this

auto_round/formats.py
```diff
@@ -355,6 +359,26 @@ def __init__(self, format, ar):
...
+            error_logs.append(f"bits={scheme.bits}")
+        if not re.search("mxfp|fp|nvfp", scheme.data_type):
+            error_logs.append(f"data_type={scheme.data_type}")
+        if scheme.data_type == "fp" and scheme.bits != 8:
+            error_logs.append(f"data_type={scheme.data_type}, bits={scheme.bits}")
+        if scheme.super_bits:
```
wenhuach21, line 371: Please note that a single format should support different data types across layers. Ensure that the refactor accommodates this scenario.

PR #1339, opened 2026-01-26
refactor init of compressor

n1ck-guo:
## Description

Move some functions that require delayed execution to `post_init` to avoid bugs caused by incorrect call order.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #

### DIFF
--- a/auto_round/utils/model.py
+++ b/auto_round/utils/model.py
@@ -25,6 +25,7 @@ from typing import Union
 import psutil
 import torch
 import transformers
+from packaging import version
 
 from auto_round import envs
 from auto_round.export.export_to_gguf.config import ModelType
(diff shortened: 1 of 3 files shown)


auto_round/compressors/base.py
```diff
@@ -865,6 +879,9 @@ def quantize_and_save(
         Raises:
             ValueError: If an unsupported format is specified.
         """
+        # post init
+        self._post_init()
```
xin3he, line 923: How about adding it in `quantize` function for develop mode?
I worry the `quantize` mode is still in use, such as UT.

auto_round/compressors/base.py
wenhuach21, line 923: How can developers know which initialization code belongs in __post_init__?

n1ck-guo: The `init` method should only contain basic attribute assignments. Complex operations, modifications to attributes, and function calls that depend on basic attributes should be performed in the `post_init` method.

PR #1349, opened 2026-01-27
support hadamard transform for mxfp4 with rtn or autoround method.

lkk12014402:
## Description

1. support hadamard transform for mxfp4

original linear：

$$
y = Wx
$$

**transform matrix** $$H$$（Hadamard should $$H^\top H = I$$，and $$H^{-1}=H^\top$$）：

$$
y = W x = (W H^\top) (H x)
$$

define：

*   $$W' = W H^\top$$（**rotated weight**）
*   $$x' = H x$$（**rotated activation**）

then：

$$
y = W' x'
$$

2. support do evaluation with huggingface/transformers

with huggingface/transformers

```
from transformers import AutoModelForCausalLM, AutoTokenizer

model_name = "./mxfp4_transformed_model"
model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype="auto")
model.to("cuda")
print(model)
tokenizer = AutoTokenizer.from_pretrained(model_name)
text = "There is a girl who likes adventure,"
print(tokenizer.decode(model.generate(**inputs, max_new_tokens=50)[0]))
```

with vllm

```
lm_eval --model hf    --model_args pretrained=./mxfp4_transformed_model    --tasks gsm8k     --batch_size 8
```

### DIFF
--- a/auto_round/autoround.py
+++ b/auto_round/autoround.py
@@ -117,7 +117,8 @@ class AutoRound:
             disable_opt_rtn (bool, optional): Disable RTN-mode optimization (iters=0) for fast quatnziation
                                               with lower accuracy. Defaults to None.
             low_cpu_mem_usage (bool, optional): Lower CPU memory mode. Defaults to False.
-            transform_config (dict, optional): transform matrix config like hadamard, like {"transform_class": "hadamard"}.
+            transform_config (dict, optional): transform matrix config like hadamard,
+                                               like {"transform_class": "hadamard"}.
 
             bits (int, optional): Weight quantization bits. Defaults to 4.
             group_size (int, optional): Weight quantization group size. Defaults to 128.
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -218,7 +218,8 @@ class BaseCompressor(object):
             minmax_lr (float, optional): Learning rate for min-max tuning; defaults to `lr`.
             low_gpu_mem_usage (bool, optional): Lower GPU memory mode. Defaults to False.
             low_cpu_mem_usage (bool, optional): Lower CPU memory mode. Defaults to False.
-            transform_config (dict, optional): transform matrix config like hadamard, like {"transform_class": "hadamard"}.
+            transform_config (dict, optional): transform matrix config like hadamard,
+                                               like {"transform_class": "hadamard"}.
             iters (int, optional): Optimization iterations. Defaults to 200.
             seqlen (int, optional): Calibration sequence length. Defaults to 2048.
             nsamples (int, optional): Number of calibration samples. Defaults to 128.


auto_round/experimental/qmodules/mx.py
```diff
@@ -145,7 +158,20 @@ def _get_float_scale(cls, scale_e8m0: torch.Tensor) -> torch.Tensor:
...
     @torch.inference_mode()
     def forward(self, input: torch.Tensor) -> torch.Tensor:
-        qdq_input = self.qdq_input(input)
+
+        if self.enable_transform:
+            from ..triton.mxfp4 import mxfp4_forward_kernel_wrapper
+
+            orig_shape = input.shape
+            x_flat = input.contiguous().flatten(end_dim=-2)
+            qdq_input, _ = mxfp4_forward_kernel_wrapper(
```
wenhuach21, line 167: How can the code ensure that the transformation is equivalent?

auto_round/compressors/base.py
```diff
@@ -213,6 +215,7 @@ def __init__(
             minmax_lr (float, optional): Learning rate for min-max tuning; defaults to `lr`.
             low_gpu_mem_usage (bool, optional): Lower GPU memory mode. Defaults to False.
             low_cpu_mem_usage (bool, optional): Lower CPU memory mode. Defaults to False.
+            transform_config (dict, optional): transform matrix config like hadamard, like {"transform_class": "hadamard"}.
```
wenhuach21: Please coordinate with Heng and Weiwei regarding the API (@n1ck-guo @WeiweiZhang1).
We should avoid fusing multiple algorithms into a single implementation, as we plan to support more algorithms in the future. We can take inspiration from the design used in LLMC.

auto_round/compressors/base.py
```diff
@@ -192,6 +193,7 @@ def __init__(
         disable_opt_rtn: bool | None = None,
         seed: int = 42,
         low_cpu_mem_usage: bool = True,
+        transform_config: dict = {},
```
wenhuach21, line 199: It’s better not to name it transform_config, as it may be confusing with Transformers.

auto_round/experimental/triton/mxfp4.py
```diff
@@ -0,0 +1,188 @@
...
+    configs=[
+        triton.Config({"BLOCK_SIZE": 32 * 32}),
+        triton.Config({"BLOCK_SIZE": 64 * 32}),
+        triton.Config({"BLOCK_SIZE": 128 * 32}),
+        triton.Config({"BLOCK_SIZE": 256 * 32}),
+        triton.Config({"BLOCK_SIZE": 512 * 32}),
+    ],
+    key=[],
+)
+@triton.jit
```
wenhuach21, line 32: I assume xpu does not support this, but it's not a big issue for now

auto_round/experimental/qmodules/mx.py
lkk12014402: pls see the transformation equation in the comments. the activation is rotated and the packed weight is also rotated. so the linear is equivalent

auto_round/experimental/triton/mxfp4.py
lkk12014402: xpu support this kernel. but I haven't tested the performance

auto_round/experimental/triton/mxfp4.py
```diff
@@ -0,0 +1,188 @@
...
+def mxfp4_forward_kernel(
```
wenhuach21, line 33: please add the source if the code is copied from another repo.
better add their license at the beginning of this file

auto_round/experimental/qmodules/mx.py
wenhuach21: ok, thanks. Should qkv or moe share a same H on activations, as they are fused in vllms

lkk12014402: only support one same H for all layers. we can support specific transform matrixs for each layer when we support other transform functions in this file `[auto_round/transforms/transforms.py](https://github.com/intel/auto-round/pull/1349/files#diff-c384e42585ac4e70e9217f23c870c46b8b66097f7d6810bea55779a55e38fe73)`

auto_round/compressors/base.py
lkk12014402: good suggestion. we can refer LLMC way like this https://github.com/vllm-project/llm-compressor/blob/main/examples/transform/quip_example.py#L26 to support more rotation methods

auto_round/autoround.py
```diff
@@ -114,6 +115,7 @@ def __new__(
             disable_opt_rtn (bool, optional): Disable RTN-mode optimization (iters=0) for fast quatnziation
                                               with lower accuracy. Defaults to None.
             low_cpu_mem_usage (bool, optional): Lower CPU memory mode. Defaults to False.
+            transform_config (dict, optional): transform matrix config like hadamard, like {"transform_class": "hadamard"}.
```
yiliu30: Please mark it as experimental feature, and clarify the limitation.

auto_round/wrapper.py
```diff
@@ -502,8 +524,17 @@ def __init__(self, orig_layer, enable_torch_compile=False, device="cpu"):
...
+        self.enable_transform = False
+        if getattr(self.orig_layer, "transform_config", False):
+            from .transforms.transforms import build_transform
+
+            self.in_transform = build_transform(**self.orig_layer.transform_config)
+            self.enable_transform = True
+
     def forward(self, x):
         act_max = self.orig_layer.act_max if hasattr(self.orig_layer, "act_max") else None
+        if self.enable_transform:
```
yiliu30, line 554: Nit: using a forward hook might make the implementation clearer from a readability perspective.

lkk12014402: I think so too. Let me look into refining the code.

wenhuach21: A cleaner API would look like: model = Hardarmd_transform(model), and then pass the transformed model to AutoRound. Please sync with Heng on this or other possibilities.

Issue #1238, opened 2026-01-08
support for mistralai/Devstral-2-123B-Instruct-2512


n1ck-guo:
1. mistralai/Devstral-2-123B-Instruct-2512 is per-tensor fp8 quantization, need change fp8 dequantize
2. On a100 with transformers >= 5.0, cannot save Devstral-2-123B-Instruct-2512. Related to Fp8Dequantize，Error as below

packages/transformers/core_model_loading.py", line 531, in reverse_transform
    kwargs["operations"] = [op.reverse_op for op in self.operations[::-1]]
                            ^^^^^^^^^^^^^
  File "/home/hengguo/miniforge3/envs/auto_round/lib/python3.12/site-packages/transformers/core_model_loading.py", line 102, in reverse_op
    raise NotImplementedError
NotImplementedError

wenhuach21:
Issue 1 is ours. Target 0.9.5
Issue2 is transformers'

yiliu30:
This issue was fixed by @SwekeR-463 in https://github.com/intel/auto-round/pull/1356 
@XuehaoSun @chensuyue  Please help quantize and relase model if needed.

Issue #1337, opened 2026-01-25
[Feature]:  Support Re‑Quantizing Per‑Tensor FP8 Models to Other Dtypes


yiliu30:
### Feature Description

As the title.

### Motivation and Use Case

- Input:
Models whose weights have already been quantized to FP8 with per‑tensor granularity, such as: https://huggingface.co/mistralai/Devstral-2-123B-Instruct-2512
Output:
Re‑quantized models in other dtypes, e.g. `W4A16`

- The relevant implementation starting point is here:
https://github.com/intel/auto-round/blob/6190869b7f0ff00dab98e786655ce2145c8a8891/auto_round/utils/model.py#L1070-L1075

### Definition of Done

- [ ] Allow taking `mistralai/Devstral-2-123B-Instruct-2512` as input, and `W4A16` as output.
- [ ] UT

PR #1356, opened 2026-01-27
feat: handle per-tensor FP8 dequantization for Devstral models

SwekeR-463:
Fix https://github.com/intel/auto-round/issues/1238

## Description

This PR adds support for dequantizing and re-quantizing FP8 models with per-tensor or per-channel granularity, such as Devstral-2-123B-Instruct-2512. Previously, the dequantization logic assumed block-wise quantization with a required block_size attribute.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #1337

### DIFF
--- a/test/test_cpu/core/test_fp8_re_quant.py
+++ b/test/test_cpu/core/test_fp8_re_quant.py
@@ -4,7 +4,7 @@ import torch
 import torch.nn as nn
 from transformers import AutoConfig
 
-from auto_round.utils.model import convert_fp8_layer_to_linear, _dequant_fp8_linear_weight
+from auto_round.utils.model import _dequant_fp8_linear_weight, convert_fp8_layer_to_linear
 
 
 class MockFP8Layer:


SwekeR-463:
And I think for the hardcoded fallback i can write the linear layer as per devstral.

yiliu30:
Hi @SwekeR-463, The refactor work has been landed. Would you mind refactor the code to align with it?
https://github.com/intel/auto-round/pull/1348
And you may need to upgrade the Transformers package to test this model.

auto_round/utils/model.py
```diff
@@ -1032,18 +1032,27 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 @with_thread_limits()
 def dequant_block_fp8_weight(
```
yiliu30: Thanks @SwekeR-463! I tested it locally and it works!
Could you help refine the function name a bit? Since it will be used for `FP8Linear`, maybe something like `_dequant_fp8_linear_weight` would be more appropriate.

SwekeR-463: > Thanks @SwekeR-463! I tested it locally and it works! Could you help refine the function name a bit? Since it will be used for `FP8Linear`, maybe something like `_dequant_fp8_linear_weight` would be more appropriate.

Yes done with this. :+1:

PR #1358, opened 2026-01-28
refactor auto_round api, using local symbol table to pass all args

n1ck-guo:
## Description

refactor auto_round api, using local symbol table to pass all args
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #

### DIFF
--- a/auto_round/autoround.py
+++ b/auto_round/autoround.py
@@ -52,6 +52,8 @@ class AutoRound:
         enable_torch_compile (bool): Whether to enable torch.compile for quant blocks/layers.
     """
 
+    SKIP_ARGS = ("local_args", "kwargs", "cls", "model_cls", "dynamic_compressor", "extra_config", "enable_adam")
+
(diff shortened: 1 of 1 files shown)


auto_round/autoround.py
```diff
@@ -454,23 +427,9 @@ def __init__(
         optimizer="AdamW",
         **kwargs,
     ):
+        local_args = {k: v for k, v in locals().items() if k not in ("local_args", "kwargs", "self")}
```
yiliu30, line 430: Nit: Looks like this line is the same across different Compressors. How about defining a function to retrieve it?

n1ck-guo: Better not, as it can lead to bugs caused by `locals()`.

Issue #1345, opened 2026-01-27
[Feature]: support general moe replacement


xin3he:
### Feature Description

Implementing a general way to replace gate_up_proj&dow_proj and inject for loop into the original forward.

### Motivation and Use Case

Currently we have [replace_module](https://github.com/intel/auto-round/blob/main/auto_round/modelling/replace_modules.py) but it requires manually change for each model.
Now, transformers==5.0.0 changes all MOE gate_up_proj&dow_proj as below:
```
 self.gate_up_proj = nn.Parameter(torch.empty(self.num_experts, 2 * self.intermediate_dim, self.hidden_dim)) 
 self.down_proj = nn.Parameter(torch.empty(self.num_experts, self.hidden_dim, self.intermediate_dim))
```
So we need a general method to replace them and inject for loop into the original forward.

### Definition of Done

- [ ] Feature implemented and working as described
- [ ] Unit tests added/updated
- [ ] Documentation updated

PR #1360, opened 2026-01-28
Support general MOE replacement for MOE models

lvliang-intel:
### Design Overview
This feature provides a generic, model-agnostic mechanism to enable quantization of Mixture-of-Experts (MOE) models by leveraging transformers 5.0+'s native MOE integration interface (@use_experts_implementation decorator and [ALL_EXPERTS_FUNCTIONS registry).

### Core Problem
MOE models (DeepSeek-V2, Llama4, Qwen-MoE, Mixtral, etc.) use fused 3D parameter tensors for expert weights in transformers 5.0. 

gate_up_proj: (num_experts, 2*intermediate_dim, hidden_dim)
down_proj: (num_experts, hidden_dim, intermediate_dim)

These fused tensors cannot be directly quantized as nn.Linear layers, which auto-round expects.

### Solution Architecture
Transformers 5.0+ MOE Interface: ALL_EXPERTS_FUNCTIONS registry , "linear_loop" (our custom implementation)
auto_round/modelling/moe_experts_impl.py:
linear_loop_experts_forward()  
          - Loops over experts using nn.Linear layers 
          - Supports quantized nn.Linear (NVFP4, INT4, etc.) 
_unfuse_experts_weights_inplace()
          - Converts 3D Parameters → nn.ModuleList[nn.Linear] 
          - Handles transposed vs non-transposed layouts
 prepare_model_for_moe_quantization() 
          - Main entry point  
          - Registers linear_loop, unfuses all experts modules 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

[](https://github.com/intel/auto-round/issues/1345)

Fixes or relates to #


auto_round/modeling/fused_moe/replace_modules.py
```diff
@@ -30,6 +30,31 @@
...
+        is_linear_loop_available,
+        prepare_model_for_moe_quantization,
+    )
+
+    if not is_linear_loop_available():
+        logger.warning("MOE handling requires transformers 5.0+. " "MOE modules will not be handled.")
+        return []
+
+    # Use transformers' experts interface
+    unfused = prepare_model_for_moe_quantization(model)
```
yiliu30, line 55: I'd like to confirm when this replacement happens. It’s after the model is loaded but before quantization, right?
One concern I have is the RAM usage. Right now, loading the model materializes all expert weights into RAM, and that’s just not doable for really big models.

lvliang-intel: Yes. Replacements are applied right after model load via update_module() and before any quantization runs. Quantization later calls materialize_model_(). So materialization happens inside quantization, not during replacement, right?

yiliu30: > Quantization later calls materialize_model_(). So materialization happens inside quantization, not during replacement, right?

This works in V4. 

Unfortunately, in V5 the materialization happens earlier, specifically inside `from_pretrained`. Because the expert weights get concatenated during loading. Please check the reproduction code here. https://github.com/huggingface/transformers/issues/43284#issuecomment-3797516794

wenhuach21: Then we could combine PR #1354 with this one. PR #1354 can be used for our already supported MoE, while this PR targets the new MoE that we don’t support yet.

PR #1370, opened 2026-01-30
fix cuda ut fail

n1ck-guo:
## Description

fix cuda ut fail

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #

### DIFF
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -509,7 +509,7 @@ class BaseCompressor(object):
             logger.info("AutoScheme is not yet supported for multimodal LLMs.")
             sys.exit(-1)
 
-        if is_quantized_input_module(model):
+        if is_quantized_input_module(self.model):
             logger.info("AutoScheme does not currently support quantized input models (e.g., FP8).")
             sys.exit(-1)
 


auto_round/compressors/base.py
```diff
@@ -507,20 +507,18 @@ def __init__(
             except (ImportError, ModuleNotFoundError):
                 logger.error("algorithm extension import error, fallback to default mode")
 
-    def _gen_auto_scheme(
-        self, model: torch.nn.Module, scheme: AutoScheme, dataset: str, device_map: Union[str, int, dict, torch.device]
```
wenhuach21, line 508: Why use self.model here? We could keep the model as a parameter instead, and later extract this logic into a separate function to make the code cleaner.

n1ck-guo: In function _gen_auto_scheme, some parameters are passed through function arguments, while others are accessed directly using `self`, which is very strange.

xin3he (approved):

PR #1376, opened 2026-01-30
support llm dynamic wint8aint8 export

mengniwang95:
## Description

support W8A8 llmc export

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

### DIFF
--- a/auto_round/utils/model.py
+++ b/auto_round/utils/model.py
@@ -625,6 +625,8 @@ def get_block_names(model, quant_vision=False):
             return [(name, module)]
         target_modules = []
         for n, m in module.named_children():
+            if hasattr(type(m), "__name__") and "NgramEmbedding" in type(m).__name__:
+                continue
             if hasattr(type(m), "__name__") and "ModuleList" in type(m).__name__:
                 target_modules.append((".".join(filter(None, (name, n))), m))
             else:


test/test_cpu/export/test_export.py
```diff
@@ -416,6 +416,24 @@ def test_export_format(self):
         assert format_list[1].output_format == "auto_gptq"
         assert format_list[1].get_backend_name() == "auto_gptq"
 
+        autoround = AutoRound(
```
yiliu30, line 419: How about adding a predefined scheme for that case?

auto_round/export/export_to_llmcompressor/export.py
```diff
@@ -18,33 +18,89 @@
...
-        scale (float): The scale factor used for quantization.
-
-    Returns:
-        torch.Tensor: The recovered floating-point weight tensor.
-    """
-    return (qdq_weight / scale).to(torch.int8)
+def construct_ct_scheme(layer):
+    from compressed_tensors.quantization import QuantizationArgs, QuantizationScheme  # pylint: disable=E0401
+
+    weights_args = QuantizationArgs(
```
yiliu30, line 38: How about importing the `INT8_W8A8` scheme as the base scheme and updating it if needed? 
 https://github.com/vllm-project/compressed-tensors/blob/0eaedebe7f4719cf927a00bc62b3dcffa9bbfdf9/src/compressed_tensors/quantization/quant_scheme.py#L231

test/test_cpu/export/test_export.py
```diff
@@ -450,3 +468,28 @@ def test_export_format_with_scheme(self, tiny_qwen_model_path):
...
+            group_size=-1,
+            sym=True,
+            act_bits=8,
+            act_data_type="int",
+            act_sym=True,
+            act_dynamic=True,
+            act_group_size=-1,
+        )
+        quantized_model_path = "./saved"
+        autoround.quantize_and_save(output_dir=quantized_model_path, format="llm_compressor")
```
yiliu30, line 475: It can be loaded by LLMC only, right? Please raise a warning if the user passes an auto‑round format.

wenhuach21: it could be implemented in formats. Heng refined formats code recently, you could sync with heng.

auto_round/formats.py
```diff
@@ -364,9 +371,9 @@ def check_scheme_args(cls: OutputFormat, scheme: QuantizationScheme) -> bool:
         error_logs = []
         if scheme.bits not in [4, 8, 16]:
             error_logs.append(f"bits={scheme.bits}")
-        if not re.search("mxfp|fp|nvfp", scheme.data_type):
+        if not re.search("mxfp|fp|nvfp|int", scheme.data_type):
             error_logs.append(f"data_type={scheme.data_type}")
-        if scheme.data_type == "fp" and scheme.bits != 8:
+        if scheme.data_type in ["fp", "int"] and scheme.bits != 8:
```
wenhuach21, line 381: @n1ck-guo The refactored code still has some gap with supporting new schemes. What I am assuming is each format should maintain its own list of supported schemes, so developers don’t need to modify other parts of the code and can simply add the corresponding packing logic.

test/test_cpu/export/test_export.py
mengniwang95: updated

auto_round/export/export_to_llmcompressor/export.py
mengniwang95: updated

yiliu30: We could import `INT8_W8A8`from the CT, but the current way is ok.

yiliu30 (approved):

auto_round/formats.py
mengniwang95: can we merge this PR first?

test/test_cpu/export/test_export.py
mengniwang95: have synced with heng

mengniwang95: @yiliu30 if user passes an auto‑round format it will be saved as fake format automatically and a warning will be raised

mengniwang95: but there is a bug when calling fake backend for saving since the model is meta....

auto_round/formats.py
wenhuach21: sure

test/test_cpu/export/test_export.py
mengniwang95: fix the meta device issue in this PR

Issue #1380, opened 2026-02-02
[Bug]: meituan-longcat/LongCat-Flash-Lite: RuntimeError: Expected tensor for argument #1 'indices' to have one of the following scalar types: Long, Int; but got CUDABFloat16Type instead


XuehaoSun:
### Problem Description

```
RuntimeError: Expected tensor for argument #1 'indices' to have one of the following scalar types: Long, Int; but got CUDABFloat16Type instead
```

### Reproduction Steps

Command: 
```
auto-round --model_name meituan-longcat/LongCat-Flash-Lite --iters 200 --bits 4
```

### Error Logs
log: [quant log](https://inteltf-jenk.sh.intel.com/job/AutoRound_new_model_test/262//artifact/logs/meituan-longcat/LongCat-Flash-Lite/1/quant.log)

```shell
Quantizing model.layers.13:  34%|███▍      | 13/38 [1:19:13<2:33:32, 368.50s/it][A[38;20m2026-02-01 19:00:13 INFO base.py L2996: ['mlp.router.classifier'] have not been quantized[0m
[38;20mquantized 784/785 layers in the block, loss iter 0: 0.107309 -> iter 168: 0.042881,'peak_ram': 227.76GB, 'peak_vram': 66.18GB[0m


Quantizing model.layers.13:  37%|███▋      | 14/38 [1:25:19<2:27:07, 367.82s/it][A

Quantizing model.ngram_embeddings.embedders.0:  37%|███▋      | 14/38 [1:25:21<2:27:07, 367.82s/it][ATraceback (most recent call last):
  File "/home/uttest/miniforge3/envs/autoround_test/bin/auto-round", line 10, in <module>
    sys.exit(run())
             ^^^^^
  File "/home/uttest/miniforge3/envs/autoround_test/lib/python3.12/site-packages/auto_round/__main__.py", line 783, in run
    start()
  File "/home/uttest/miniforge3/envs/autoround_test/lib/python3.12/site-packages/auto_round/__main__.py", line 516, in start
    tune(args)
  File "/home/uttest/miniforge3/envs/autoround_test/lib/python3.12/site-packages/auto_round/__main__.py", line 722, in tune
    model, folders = autoround.quantize_and_save(export_dir, format=args.format)  # pylint: disable=E1101
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1388, opened 2026-02-03
Support quant for `meituan-longcat/LongCat-Flash-Lite`

Kaihui-intel:
## Description

https://github.com/intel/auto-round/issues/1380

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #

### DIFF
--- a/auto_round/utils/model.py
+++ b/auto_round/utils/model.py
@@ -625,6 +625,8 @@ def get_block_names(model, quant_vision=False):
             return [(name, module)]
         target_modules = []
         for n, m in module.named_children():
+            if hasattr(type(m), "__name__") and "NgramEmbedding" in type(m).__name__:
+                continue
             if hasattr(type(m), "__name__") and "ModuleList" in type(m).__name__:
                 target_modules.append((".".join(filter(None, (name, n))), m))
             else:


auto_round/utils/model.py
```diff
@@ -625,6 +625,8 @@ def _search_block(name, module):
             return [(name, module)]
         target_modules = []
         for n, m in module.named_children():
+            if hasattr(type(m), "__name__") and "NgramEmbedding" in type(m).__name__:
```
wenhuach21, line 628: @WeiweiZhang1  @mengniwang95 
is there better way to handle this. Using name is not very robust

xin3he (approved):

auto_round/utils/model.py
mengniwang95: is there any corner case that type(module) doesn't have __name__ attribute?

wenhuach21: What I mean is that the model owner may rename modules, or different models may use different module names.

Kaihui-intel: There is currently no better way, merge this pr first.

wenhuach21: Is it sufficient to check whether the module list includes all embedding or non-linear/conv layers?

wenhuach21: Please do not merge immediately after responding. Allow some time for reviewers to follow up.

Kaihui-intel: > Is it sufficient to check whether the module list includes all embedding or non-linear/conv layers?

Not enough, we still need to process the inputs for `post_projs` in `NgramEmbedding`.

wenhuach21: I mean for each block list generated by this function, we have a double check.  If the block list has no linear, we remove it from the results. It should be at least equal to your current solution

Issue #1451, opened 2026-02-12
[Bug]: cannot quantize models larger than available VRAM using v0.10.0 unless low_gpu_mem_usage = True


davedgd:
### Problem Description

With release V0.9.7, I was able to begin quantization of google/gemma-3-27b-it on a 48GB GPU (RTX 6000 Ada) despite the model not fitting entirely into VRAM. Now, the exact same call in v0.10.0 fails when the model is moving to CPU:

```python
2026-02-12 15:00:32 INFO autoround.py L165: using MLLM mode for multimodal model.
Loading weights: 100%|████████████████| 1247/1247 [00:00<00:00, 2167.15it/s, Materializing param=model.vision_tower.vision_model.post_layernorm.weight]
2026-02-12 15:01:03 INFO base.py L486: using torch.bfloat16 for quantization tuning
2026-02-12 15:01:03 WARNING formats.py L154: some layers are skipped quantization (shape not divisible by 32).
2026-02-12 15:01:03 INFO base.py L1739: start to cache block inputs
Traceback (most recent call last):
  File "/home/daved/project/code/convert_autoround.py", line 43, in <module>
    ar.quantize_and_save(output_dir = quant_path, format = "auto_round")
  File "/home/daved/miniforge3/envs/autoround/lib/python3.12/site-packages/auto_round/compressors/base.py", line 949, in quantize_and_save
    model, _ = self.quantize()
               ^^^^^^^^^^^^^^^
  File "/home/daved/miniforge3/envs/autoround/lib/python3.12/site-packages/auto_round/compressors/base.py", line 1740, in quantize
    all_inputs = self.try_cache_inter_data_gpucpu(all_first_block_names, self.nsamples, layer_names=layer_names)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/daved/miniforge3/envs/autoround/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/daved/miniforge3/envs/autoround/lib/python3.12/site-packages/auto_round/compressors/base.py", line 2209, in try_cache_inter_data_gpucpu
    new_max_memory[device] = max_memory[device] * 0.9
[...]

PR #1457, opened 2026-02-21
Fix KeyError when GPU is missing from accelerate max_memory

lvliang-intel:
## Description

**Issue1451** https://github.com/intel/auto-round/issues/1451

**Root Cause:**

When GPU VRAM is nearly exhausted, accelerate's get_max_memory() silently drops the GPU from its return dict (e.g., returns {'cpu': ...} instead of {0: ..., 'cpu': ...}). The code then tries to access max_memory[0], causing a KeyError crash.

**Solution:**

Skip unavailable GPU devices with a continue guard before accessing max_memory[device]. When no GPU survives, raise torch.OutOfMemoryError to trigger the existing CPU fallback handler, avoiding unnecessary get_balanced_memory/dispatch_model overhead. Also normalize bare device strings (e.g., "cuda" → "cuda:0") in parse_available_devices to prevent hardcoding device = 0.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

https://github.com/intel/auto-round/issues/1451

### DIFF
--- a/auto_round/compressors/mllm/compressor.py
+++ b/auto_round/compressors/mllm/compressor.py
@@ -67,7 +67,7 @@ def _only_text_test(model, tokenizer, device, model_type):
 
     # Estimate model size and check if it fits in GPU memory.
     # When the model is too large, we skip the GPU transfer and test on CPU only.
-    use_gpu = True if device != cpu else False
+    use_gpu = True if device != "cpu" else False
     if device != "cpu" and torch.cuda.is_available():
         model_bytes = sum(p.numel() * p.element_size() for p in model.parameters())
         dev_idx_str = device.split(":")[-1] if ":" in device else "0"


auto_round/compressors/base.py
```diff
@@ -2184,10 +2184,29 @@ def try_cache_inter_data_gpucpu(self, block_names, nsamples, layer_names=None, l
                                     device = 0
                                 else:
                                     raise ValueError(f"Unsupported device {device} in device_map: {self.device_map}")
+                                if device not in max_memory:
+                                    if device != "cpu":
+                                        logger.warning(
```
wenhuach21: I don't think users care about this info. It would be better change it from warning to commments

auto_round/compressors/mllm/compressor.py
```diff
@@ -65,20 +65,47 @@ def _only_text_test(model, tokenizer, device, model_type):
...
-        model = model.to("cpu")
-        inputs = inputs.to("cpu")
+    # Estimate model size and check if it fits in GPU memory.
+    # When the model is too large, we skip the GPU transfer and test on CPU only.
+    use_gpu = True
+    if device != "cpu" and torch.cuda.is_available():
+        try:
+            model_bytes = sum(p.numel() * p.element_size() for p in model.parameters())
+            dev_idx = int(device.split(":")[-1]) if ":" in device else 0
+            free_bytes, _ = torch.cuda.mem_get_info(dev_idx)
```
wenhuach21: Could we replace the try with an if, or at least limit the scope of the try block to only wrap mem_get_info, since I believe the exception originates there?

auto_round/compressors/mllm/compressor.py
```diff
@@ -65,20 +65,47 @@ def _only_text_test(model, tokenizer, device, model_type):
...
+            pass
+
+    if use_gpu:
         try:
+            inputs = inputs.to(device)
+            model = model.to(device)
             model(**inputs)
-        except:
+            return True
+        except RuntimeError as e:
```
n1ck-guo: It would be better to add a check to determine if it is OOM.

auto_round/compressors/mllm/compressor.py
```diff
@@ -65,20 +65,47 @@ def _only_text_test(model, tokenizer, device, model_type):
...
-        inputs = inputs.to(device)
-        model = model.to(device)
-        model(**inputs)
-        return True
-    except RuntimeError as e:
-        model = model.to("cpu")
-        inputs = inputs.to("cpu")
+    # Estimate model size and check if it fits in GPU memory.
+    # When the model is too large, we skip the GPU transfer and test on CPU only.
+    use_gpu = True
```
wenhuach21: use_gpu=True if device !=cpu else False?

PR #1473, opened 2026-02-27
Enable CUDA CI

XuehaoSun:
## Description

Enable CUDA CI

## TODO

- [x] Skip `gptqmodel`, `auto-gptq` test
- [x] Fix absolute path (like `/models/xxx`)
- [x] Skip tests that require a lot of hardware resources
- [x] Separate the vLLM-related unit tests @xin3he 
- [ ] `triton` issue
- [x] CUDA compatibility

### DIFF
--- a/test/test_cuda/integrations/test_sglang.py
+++ b/test/test_cuda/integrations/test_sglang.py
@@ -107,6 +107,7 @@ class TestAutoRound:
 
         shutil.rmtree(self.save_dir, ignore_errors=True)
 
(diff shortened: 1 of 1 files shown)


.azure-pipelines/unit-test-cuda.yml
```diff
@@ -0,0 +1,191 @@
+trigger: none
+
```
chensuyue, line 2: Need to setup pipeline trigger logical, only Repo members could trigger with comments like `test_cuda`.

PR #1511, opened 2026-03-08
Fix #1284: preserve FP8 format for layers specified in ignore_layers

### LINKED ISSUE #1284: [feat]  For FP8 models, keep layers specified in `ignore_layers` in their original FP8 format
For Deepseek v32 and similar FP8 models, it is preferable to keep layers specified in `ignore_layers` (such as indexer or attn) in their original FP8 format, rather than dequantizing them to BF16.

Expected behavior:

```
AR_LOG_LEVEL=TRACE auto_round --model /models/Qwen3-8B-FP8 --ignore_layers "attn"
```
All layers within attention (matching "attn") should remain in FP8 format, not be dequantized to BF16 or float.

Depends on https://github.com/intel/auto-round/issues/1283

cc @wenhuach21 @thuang6 @xin3he

LuciferDono:
## Description

For FP8 models (e.g. Qwen3-8B-FP8, DeepSeek-V3), layers specified in `ignore_layers` are currently dequantized to BF16 even though the user explicitly wants them kept unchanged. This PR fixes two issues:

1. **Detection gap (supersedes #1283):** `get_fp_layer_names()` only checked `SUPPORTED_LAYER_TYPES` (nn.Linear, Conv1D), missing FP8-specific layer types like `FP8Linear`. Now also checks `INNER_SUPPORTED_LAYER_TYPES`.

2. **Preservation:** `convert_module_to_hp_if_necessary()` blindly converted all quantized layers to high precision. Now accepts an `ignore_layers` parameter to skip specified layers, keeping them in their original FP8 format.

## Type of Change
- [x] Bug fix
- [x] New feature

## Related Issues

Fixes #1284
Supersedes the detection fix from #1283

## Changes

- **`auto_round/compressors/utils.py`** — `get_fp_layer_names()` now detects FP8Linear via `INNER_SUPPORTED_LAYER_TYPES`
- **`auto_round/utils/weight_handler.py`** — `convert_module_to_hp_if_necessary()` gains backward-compatible `ignore_layers` parameter (default `None`)
- **`auto_round/compressors/base.py`** — Resolved ignore layer names stored after `configure_layer_config()` and threaded through all model-level conversion call sites
- **`test/test_cpu/utils/test_fp8_ignore_layers.py`** — 6 unit tests covering detection, skip-on-convert, and end-to-end flow with mock FP8Linear layers

### DIFF
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -41,6 +41,7 @@ from auto_round.compressors.utils import (
     check_need_act_calibration,
     check_skippable_keywords,
     collect_best_params,
+    get_fp_layer_names,
     get_shared_keys,
     infer_bits_by_data_type,
     init_cache,
@@ -1406,8 +1407,10 @@ class BaseCompressor(object):
                             shard_writer(self, name=n)
                         m.to("meta")
 
-        # Convert remaining fp8
-        convert_module_to_hp_if_necessary(self.model, self.amp_dtype, self.device)
+        # Convert remaining fp8, but keep ignored layers in their original format
+        convert_module_to_hp_if_necessary(
+            self.model, self.amp_dtype, self.device, ignore_layers=self._resolved_ignore_layers
+        )
         if self.is_immediate_saving:
             shard_writer(self, is_finalize=True)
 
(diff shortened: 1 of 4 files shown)


LuciferDono:
> @LuciferDono Have you ever validated it with vLLM? The quantization_config of fp8 would be overwritten by AutoRound, will it cause a failure to load fp8 tensors?

nope, haven't validated with vLLM directly; don't have a GPU setup for it. The unit tests cover the skip logic, but you're right that the quantization_config interaction needs validation. happy to adjust the code if you can point out what would break, or if someone with a GPU env could test the loading path

PR #1515, opened 2026-03-09
support hadamard transform for mxfp4 with rtn or autoround method

lkk12014402:
## Description

1. support hadamard transform for mxfp4

original linear：

$$
y = Wx
$$

**transform matrix** $$H$$（Hadamard should $$H^\top H = I$$，and $$H^{-1}=H^\top$$）：

$$
y = W x = (W H^\top) (H x)
$$

define：

*   $$W' = W H^\top$$（**rotated weight**）
*   $$x' = H x$$（**rotated activation**）

then：

$$
y = W' x'
$$

2. support do evaluation with huggingface/transformers

with huggingface/transformers

```
from transformers import AutoModelForCausalLM, AutoTokenizer

model_name = "./mxfp4_transformed_model"
model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype="auto")
model.to("cuda")
print(model)
tokenizer = AutoTokenizer.from_pretrained(model_name)
text = "There is a girl who likes adventure,"
print(tokenizer.decode(model.generate(**inputs, max_new_tokens=50)[0]))
```

with vllm

```
lm_eval --model hf    --model_args pretrained=./mxfp4_transformed_model    --tasks gsm8k     --batch_size 8
```

### DIFF
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -197,7 +197,6 @@ class BaseCompressor(object):
         disable_opt_rtn: bool | None = None,
         seed: int = 42,
         low_cpu_mem_usage: bool = True,
-        transform_config: dict | None = None,
         **kwargs,
     ):
         """Initialize AutoRound with quantization and tuning configuration.
@@ -546,7 +545,7 @@ class BaseCompressor(object):
             except (ImportError, ModuleNotFoundError):
                 logger.error("algorithm extension import error, fallback to default mode")
 
-        self.transform_config = {} if transform_config is None else dict(transform_config)
+        self.transform_config = kwargs.pop("transform_config", {})
 
     def _gen_auto_scheme(self) -> dict[str, dict]:
         if self.mllm:
--- a/auto_round/experimental/qmodules/mx.py
+++ b/auto_round/experimental/qmodules/mx.py
@@ -206,7 +206,7 @@ class TransformMXFP4QuantLinear(MXFP4QuantLinear):
         super().__init__(*args, **kwargs)
         self.enable_transform = True
         self.register_buffer(
-            "forward_hadamard_transform",
+            "transform_matrix",
             torch.empty(
                 self.group_size,
                 self.group_size,
--- a/auto_round/experimental/transform/apply.py
+++ b/auto_round/experimental/transform/apply.py
@@ -51,10 +51,10 @@ def _apply_to_module(
     """
 
     # create transform as submodule
-    transform_name = "forward_hadamard_transform"
+    transform_name = "transform_matrix"
 
     if config.location == "input":
-        from .triton.utils import is_triton_available
+        from auto_round.experimental.transform.triton.utils import is_triton_kernel_available
 
         # activation needs transpose
         inp_transform = build_transform(
@@ -69,7 +69,7 @@ def _apply_to_module(
         else:
             transform_weight = None
 
-        if is_triton_available():
+        if is_triton_kernel_available():
             from auto_round.experimental.transform.triton.mxfp4 import mxfp4_forward_kernel_wrapper
 
             def input_hook(self, args):
@@ -80,7 +80,7 @@ def _apply_to_module(
                 qdq_input, _ = mxfp4_forward_kernel_wrapper(
                     x_flat,
                     (
-                        transform_weight if transform_weight is not None else self.forward_hadamard_transform.T
+                        transform_weight if transform_weight is not None else self.transform_matrix.T
                     ),  # this matrix from w_transform, needs transpose
                 )
                 return qdq_input.reshape(orig_shape)
@@ -101,8 +101,8 @@ def _apply_to_module(
                     input = input.view(-1, transform_weight.shape[0])
                     return _multihead_matmul(input, transform_weight.to(input.device)).view(ori_shape)
                 else:
-                    input = input.view(-1, self.forward_hadamard_transform.shape[0])
-                    return _multihead_matmul(input, self.forward_hadamard_transform.T).view(ori_shape)
+                    input = input.view(-1, self.transform_matrix.shape[0])
+                    return _multihead_matmul(input, self.transform_matrix.T).view(ori_shape)
 
             # for fused transform + quantization kernel
             module.pre_dequantized_input = False
--- a/auto_round/experimental/transform/triton/mxfp4.py
+++ b/auto_round/experimental/transform/triton/mxfp4.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Refer code here:
+#   https://github.com/IST-DASLab/FP-Quant/blob/master/inference_lib/src/fp_quant/module/triton/mxfp4.py
+
 import torch
 import triton  # pylint: disable=E0401
 import triton.language as tl  # pylint: disable=E0401
--- a/auto_round/experimental/transform/triton/utils.py
+++ b/auto_round/experimental/transform/triton/utils.py
@@ -4,7 +4,7 @@
 import torch
 
 
-def is_triton_available() -> bool:
+def is_triton_kernel_available() -> bool:
     """
     Best-effort check for whether Triton kernel path can be used.
     """
--- a/auto_round/experimental/transform/patch_modules.py
+++ b/auto_round/experimental/transform/patch_modules.py
@@ -202,8 +202,8 @@ def patch_quantlinear():
             self.input_global_scale = input_global_scale.to(torch.float32).to(device).reshape([1])
 
         # add transform weight
-        transform = getattr(linear, "forward_hadamard_transform")
-        self.register_buffer("forward_hadamard_transform", transform.weight.to(device))
+        transform = getattr(linear, "transform_matrix")
+        self.register_buffer("transform_matrix", transform.weight.to(device))
         return
 
     QuantLinear.pack = _pack_patched


auto_round/experimental/transform/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+        return hadamard_transform(x.view(-1, self.group_size), scale=self.scale).view(x_shape)
+
+    def get_transform_matrix(self, device: torch.device = None, dtype: torch.dtype = None):
+        return hadamard_transform(
+            torch.eye(self.group_size, device=device, dtype=dtype), scale=1 / math.sqrt(self.group_size)
+        )
+
+
+TRANSFORMS = {
+    "identity": IdentityTransform,
```
wenhuach21, line 122: what is the identity for?
Could this API support awq transformation? If not, just naming it as  Hadamard rather than a general name  is better?

auto_round/autoround.py
```diff
@@ -89,6 +89,7 @@ def __new__(
         enable_alg_ext: bool = False,
         disable_opt_rtn: bool | None = None,
         low_cpu_mem_usage: bool = True,
+        transform_config: dict = {},
         **kwargs,
     ) -> BaseCompressor:
```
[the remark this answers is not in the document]
wenhuach21: good catch, need to fix this issue

auto_round/autoround.py
```diff
@@ -89,6 +89,7 @@ def __new__(
         enable_alg_ext: bool = False,
         disable_opt_rtn: bool | None = None,
         low_cpu_mem_usage: bool = True,
+        transform_config: dict = {},
```
wenhuach21: @n1ck-guo  does this API align with the future design to support more algorithms

auto_round/experimental/transform/transform_config.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Configuration of transforms to be applied to a model. This config is to be
+    serialized within a model's `config.json` file
+    """
+
+    transform_block_size: int = Field(default=32)
+
+    transform_type: str = Field(default="hadamard")
+
+    location: str = Field(default="weight")
```
wenhuach21: this should not be exposed to users

auto_round/experimental/transform/transform_config.py
```diff
@@ -0,0 +1,19 @@
+# # Copyright (C) 2026 Intel Corporation
+# # SPDX-License-Identifier: Apache-2.0
+
+from pydantic import BaseModel, Field
+
+__all__ = ["TransformConfig"]
+
+
+class TransformConfig(BaseModel):
```
wenhuach21, line 9: is a dataclass enough

auto_round/experimental/transform/triton/mxfp4.py
```diff
@@ -0,0 +1,189 @@
...
+
+from random import randint
+
+import torch
+import triton
+import triton.language as tl
+
+
+@triton.autotune(
+    configs=[
```
wenhuach21, line 24: Currently, there is no code supporting other schemes, and the implementation is not general enough to support them easily. In addition, there is no check or exception to inform users that only MXFP4 is supported. It is strongly recommended to add PyTorch support for other schemes.

auto_round/experimental/transform/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+
+import inspect
+import math
+from typing import Any, Callable, Dict
+
+import torch
+import torch.nn as nn
+from fast_hadamard_transform import hadamard_transform
+
```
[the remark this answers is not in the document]
wenhuach21: nice catch

auto_round/experimental/transform/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+        self.group_size = group_size
+        self.scale = 1 / math.sqrt(self.group_size)
+
+    def forward(self, x: torch.Tensor):
+        # Hadamard transform is it own inverse
+        x_shape = x.shape
+        return hadamard_transform(x.view(-1, self.group_size), scale=self.scale).view(x_shape)
+
+    def get_transform_matrix(self, device: torch.device = None, dtype: torch.dtype = None):
+        return hadamard_transform(
```
wenhuach21: Is there any randomness in this matrix? If so, how do you guarantee that the same matrix is used during both tuning and inference? If not, is it still a good choice for the algorithm?

auto_round/experimental/transform/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import inspect
+import math
+from typing import Any, Callable, Dict
```
wenhuach21, line 17: use default dict

auto_round/experimental/transform/transforms.py
lkk12014402: They’re not the same thing: the “transform” here is more like a rotation / basis-mixing transform (often orthogonal, invertible, and efficiently structured), whereas AWQ’s “transformation” usually refers to activation-aware per-channel scaling / re-parameterization.

they can be combined: in practice you can first apply a rotation (e.g., Hadamard or an orthogonal rotation) to make the distribution more uniform, and then apply AWQ-style scaling/clipping to further reduce quantization error.

auto_round/experimental/transform/transform_config.py
lkk12014402: BaseModel is better for this case

auto_round/experimental/transform/triton/mxfp4.py
lkk12014402: yes, I will support more combination solutions further.

auto_round/experimental/transform/transforms.py
lkk12014402: not any randomness in this matrix now. will consider this when support more transform variants

auto_round/experimental/transform/transforms.py
lkk12014402: what's your points for using default dict?

auto_round/experimental/transform/transforms.py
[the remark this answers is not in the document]
lkk12014402: I know this issue. I will remove the `fast_hadamard_transform` dependency and implement them with python code.

auto_round/experimental/transform/transform_config.py
lkk12014402: add `exclude=True`

auto_round/experimental/transform/transforms.py
wenhuach21: I meant the code’s generatability. Given the name, this API should support AWQ transformations as well as other types of transformations. If it doesn’t, it should be given a more specific name.

auto_round/experimental/transform/transforms.py
wenhuach21: def func(a:dict)

auto_round/inference/convert_model.py
```diff
@@ -650,6 +653,18 @@ def convert_hf_model(model: nn.Module, target_device: str = "cpu") -> tuple[nn.M
     layer_configs = get_layer_config(model, quantization_config)
     used_backends = _replace_by_quant_layers(model, layer_configs, backend, target_device, packing_format)
 
+    if getattr(quantization_config, "transform_config", None) is not None:
+        from ..experimental.transform.apply import apply_transform
```
wenhuach21: Prefer absolute imports over relative imports whenever possible to improve readability and avoid ambiguity.

auto_round/experimental/transform/apply.py
```diff
@@ -0,0 +1,96 @@
...
+        module.pre_dequantized_input = True
+
+        module.register_forward_pre_hook(input_hook, prepend=True)
+
+    elif config.location == "weight":
+        # eagerly apply transformation to weight
+        # fuse transform into weight
+        assert hasattr(module, "weight")
+        with torch.no_grad():
+            getattr(module, "weight").copy_(transform(module.weight.to("cuda")).to(module.weight.device))
```
wenhuach21: be careful of devices without cuda

auto_round/experimental/transform/utils.py
```diff
@@ -0,0 +1,77 @@
+# # Copyright (C) 2026 Intel Corporation
+# # SPDX-License-Identifier: Apache-2.0
+
+import transformers
+
+from ...wrapper import WrapperLinear
+
+
+def patch_wrapperlinear_qdq_weight_to_apply_transform(transform_attr: str = "forward_hadamard"):
```
wenhuach21: too long

auto_round/experimental/transform/utils.py
```diff
@@ -0,0 +1,77 @@
...
+            return orig_qdq_weight(self, value, min_scale, max_scale)
+
+        min_scale.data.clamp_(0, 1.0)
+        max_scale.data.clamp_(0, 1.0)
+
+        weight = self.orig_layer.weight
+        if weight.device.type == "meta":
+            weight = self.orig_layer.get_weight().to(self.device)
+
+        is_conv1d = type(self.orig_layer) == transformers.pytorch_utils.Conv1D
```
wenhuach21: This design is not very robust, since further refinements or bug fix to the original `qdq` function should be also applied to  this function as well. It would be better to maximize code reuse, for example by adding a `transform` argument to the original function.

auto_round/experimental/transform/triton/mxfp4.py
```diff
@@ -0,0 +1,188 @@
+# Copyright (c) 2026 Intel Corporation
```
yiliu30, line 1: Please add `__init__` for `triton` subfolder.

auto_round/experimental/transform/triton/mxfp4.py
```diff
@@ -0,0 +1,189 @@
...
+        x_ptr=x,
+        hadamard_matrix_ptr=hadamard_matrix,
+        output_ptr=output,
+        clip_mask_ptr=clip_mask,
+        n_elements=n_elements,
+        hadamard_dim=hadamard_matrix.shape[-1],
+        group_size=32,
+        gaussian_scale=gaussian_scale,
+        quest=quest,
+    )
```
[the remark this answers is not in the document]
yiliu30: @lkk12014402 please help address this concern.

auto_round/experimental/transform/transform_config.py
yiliu30: Please add `pydantic` to requirements if you prefer it.

auto_round/inference/convert_model.py
```diff
@@ -650,6 +653,18 @@ def convert_hf_model(model: nn.Module, target_device: str = "cpu") -> tuple[nn.M
...
+        from ..experimental.transform.transform_config import TransformConfig
+
+        transform_config = quantization_config.transform_config
+        act_transform_config = TransformConfig(
+            transform_block_size=transform_config["transform_block_size"],
+            transform_type=transform_config["transform_type"],
+            location="input",
+        )  # apply to activation
+        model = apply_transform(model, act_transform_config, desc="Register pre forward hood for transform")
```
yiliu30:
```suggestion
        model = apply_transform(model, act_transform_config, desc="Register pre forward hook for transform")
```

auto_round/experimental/qmodules/mx.py
```diff
@@ -145,10 +145,14 @@ def _get_float_scale(cls, scale_e8m0: torch.Tensor) -> torch.Tensor:
 
     @torch.inference_mode()
     def forward(self, input: torch.Tensor) -> torch.Tensor:
-        qdq_input = self.qdq_input(input)
+        if not self.pre_dequantized_input:
```
yiliu30, line 149: What does the `pre_dequantized_input` refer to? 
Please add the new attr in `__init__` if needed.

auto_round/experimental/transform/utils/hadamard.py
```diff
@@ -0,0 +1,150 @@
+# # Copyright (C) 2026 Intel Corporation
+# # SPDX-License-Identifier: Apache-2.0
+
+import math
+from pathlib import Path
+
+import torch
+from safetensors import safe_open
+
+REPO_PATH = Path(__file__).parent / "hadamards.safetensors"
```
yiliu30, line 10: Is it possible for us to generate it from code instead of loading it from disk?

auto_round/experimental/transform/utils/hadamard.py
```diff
@@ -0,0 +1,151 @@
+# # Copyright (C) 2026 Intel Corporation
+# # SPDX-License-Identifier: Apache-2.0
+
+import math
+from pathlib import Path
+
+import torch
+from safetensors import safe_open
```
wenhuach21, line 8: this is not a sym lib

requirements.txt
```diff
@@ -6,4 +6,5 @@ py-cpuinfo
 threadpoolctl
 torch
 tqdm
-transformers>=4.38
\ No newline at end of file
+transformers>=4.38
+pydantic==2.12.2
```
wenhuach21: better not add version restriction which may cause llmc upstream issue

auto_round/compressors/base.py
```diff
@@ -198,6 +199,7 @@ def __init__(
         disable_opt_rtn: bool | None = None,
         seed: int = 42,
         low_cpu_mem_usage: bool = True,
+        transform_config: dict | None = None,
```
wenhuach21: This API should be refined.  it should at least support str, like hadamard or a specific config like HadamadConifg which should be class not a dict.

test/test_cuda/transform/test_mxfp4_transform.py
```diff
@@ -0,0 +1,73 @@
...
+        # Yield to hand control to the test methods
+        yield
+
+        # ===== TEARDOWN (teardown_class) =====
+        print("[Teardown] Running after all tests in class")
+        shutil.rmtree("./saved", ignore_errors=True)
+        shutil.rmtree("runs", ignore_errors=True)
+    """
+
+    def test_transform_mxfp4_quant_infer(self):
```
wenhuach21, line 30: This PR introduces a relatively large feature. Please add more unit tests covering different models (e.g., MoE) , schemes (e.g., RTN and tuning) and so on

auto_round/experimental/qmodules/mx.py
```diff
@@ -193,6 +197,25 @@ def unpack_data(self, packed_data: torch.Tensor) -> torch.Tensor:
...
+    """
+    Quantized linear layer using the MXFP4 quantization scheme.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.weight_name = "weight_packed"
+        super().__init__(*args, **kwargs)
+        self.enable_transform = True
+        self.register_buffer(
+            "forward_hadamard_transform",
```
wenhuach21: is this name well aligned with the community, llmc/quark?

auto_round/experimental/transform/triton/mxfp4.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+    Refer code here: https://github.com/IST-DASLab/FP-Quant/blob/master/inference_lib/src/fp_quant/module/triton/mxfp4.py
+    Apply Hadamard transform + group-wise FP4 quantize/dequantize on x.
+
+    Note:
+    The output is still in the Hadamard-transformed space (no inverse Hadamard is applied).
+    """
+    # Pick a device — we require CUDA
+    device = x.device
+    if device.type != "cuda":
```
wenhuach21, line 154: do we code the logic not use kernel for xpu and cpu?

wenhuach21: xpu should support triton kernel, right? Please have a check.

lkk12014402: yes, pls take a look here https://github.com/intel/auto-round/pull/1515/changes#diff-58146b7de6d5689c0702080b9e749092dad9190a6b89aca21f9d359da2c62710R72

auto_round/experimental/transform/triton/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    try:
+        import triton  # noqa: F401
+    except Exception:
+        return False
+
+    if not torch.cuda.is_available():
+        return False
+
+    try:
+        from .mxfp4 import mxfp4_forward_kernel_wrapper  # noqa: F401
```
wenhuach21: There are still many relative imports. Please help refine them.

auto_round/experimental/transform/apply.py
```diff
@@ -0,0 +1,158 @@
...
+
+    if config.location == "input":
+        from .triton.utils import is_triton_available
+
+        # activation needs transpose
+        inp_transform = build_transform(
+            **config.dict(),
+            inverse=True,
+            device="cpu",
+            precision=module.dtype,
```
wenhuach21, line 64: how about changing precision to dtype?

wenhuach21:
We could add a news item on the homepage after this PR is merged and link it to some accuracy results. However, we should also warn users that this feature is not well supported by the serving frameworks yet.

auto_round/experimental/transform/triton/utils.py
lkk12014402: fixed

test/test_cuda/transform/test_mxfp4_transform.py
lkk12014402: will add more uts in another pr

requirements.txt
lkk12014402: remove version restriction

auto_round/experimental/transform/triton/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    try:
+        import triton  # pylint: disable=E0401
+    except Exception:
+        return False
+
+    if not torch.cuda.is_available():
+        return False
+
+    try:
+        from auto_round.experimental.transform.triton.mxfp4 import mxfp4_forward_kernel_wrapper  # pylint: disable=E0401
```
yiliu30, line 20: nit: This function is a bit confusing; better use two separate functions, one to check the `triton` and another for `mxfp4_forward_kernel_wrapper `.

auto_round/experimental/transform/utils/hadamards.safetensors
yiliu30, line 1: Could we avoid uploading files like this to the code repo?

auto_round/experimental/transform/utils/hadamard.py
lkk12014402: the code is for loading hardmard tensor

auto_round/experimental/qmodules/mx.py
lkk12014402: the name refer this https://github.com/IST-DASLab/FP-Quant/blob/master/inference_lib/src/fp_quant/module/linear.py#L138

wenhuach21: FP quant is an academic library. If neither LLMC nor Quark defines this term, I suggest using a clearer name such as input_hadamard_matrix.

wenhuach21: Or simply use hadamard_matrix, since there’s no need to distinguish it from the weight, the weight can just use its inverse.

lkk12014402: ok, no problem

lkk12014402: let me do this fix in another pr when I add document for hadamard transform? @wenhuach21

auto_round/experimental/transform/utils/hadamard.py
wenhuach21: I mean you need to check the import whether the lib has been installed or not

auto_round/experimental/qmodules/mx.py
wenhuach21: What's the matter? Is there any potential conflict with other PRs? I strongly suggest fixing it in this PR. If there’s a reason I’m not aware of, please do this change:

1 Change the PR title to [Step 1] xxxx.

2 Move transform_config from the API to kwargs to temporarily hide it.

3 attach some accuracy data to show the correctness

auto_round/experimental/transform/triton/mxfp4.py
```diff
@@ -0,0 +1,189 @@
...
+    output_ptr,
+    clip_mask_ptr,
+    n_elements: tl.constexpr,
+    hadamard_dim: tl.constexpr,
+    group_size: tl.constexpr,
+    gaussian_scale: tl.constexpr,
+    quest: tl.constexpr,
+    BLOCK_SIZE: tl.constexpr,
+):
+    offsets_hadamard = tl.arange(0, hadamard_dim * hadamard_dim)
```
wenhuach21, line 46: Additionally, make sure to mark the source or include the third-party license if the code is taken from another repository

auto_round/experimental/transform/triton/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    Best-effort check for whether Triton kernel path can be used.
+    """
+    try:
+        import triton  # pylint: disable=E0401
+    except Exception:
+        return False
+
+    if not torch.cuda.is_available():
+        return False
```
wenhuach21, line 17: The behavior does not match the function name, as it currently only considers CUDA. Consider renaming it to is_triton_kernel_available or another more appropriate name.

lkk12014402: update function name

auto_round/experimental/transform/triton/mxfp4.py
lkk12014402: the source of the code is declared at the beginning of the file

auto_round/experimental/qmodules/mx.py
lkk12014402: the code has been updated based on your comments.

auto_round/compressors/base.py
wenhuach21: Thanks for addressing my comments. The main remaining issue is this one. Unlike LLMC, this argument is currently mixed into AutoRound, which makes it unclear what the transform actually refers to.

Additionally, using a dict type is not very user-friendly, as users may not know what fields or values should be provided.

auto_round/experimental/transform/utils/hadamard.py
wenhuach21: If this library is already installed via PyTorch or Transformers, there is no need to add this check.

auto_round/experimental/transform/triton/mxfp4.py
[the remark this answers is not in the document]
hshen14: HMT can be applied to NVFP as well, so the group size could be 16. It seems AI suggestion looks reasonable.

yiliu30 (approved):
Others LGTM

auto_round/experimental/transform/triton/utils.py
lkk12014402: good suggestion, I'll modify it in next pr before release

auto_round/experimental/transform/utils/hadamards.safetensors
lkk12014402: the code refer to here https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/transform/utils/hadamard.py. The Hadamard matrices in a `.safetensors` file can provide precomputed, validated base blocks (K×K)** for non–power-of-two dimensions, so `random_hadamard_matrix(n)` can work when `n = K · 2^m`. And avoids complex or unavailable runtime constructions for those non–power-of-two Hadamard orders.

auto_round/experimental/transform/triton/mxfp4.py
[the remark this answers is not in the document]
lkk12014402: this is for mxfp4, so the group_size is 32

wenhuach21:
please resolve the API issue and then merge

PR #1553, opened 2026-03-16
fix wrong match of ignore_layers and use warning instead of error for mismatch

xin3he:
## Description

layers.1 now matches layers.1x, which is not expected.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #

### DIFF
--- a/auto_round/compressors/utils.py
+++ b/auto_round/compressors/utils.py
@@ -28,7 +28,7 @@ from torch.amp import autocast
 from auto_round.export.export_to_gguf.config import GGML_QUANT_SIZES, GGUF_CONFIG, GGUF_INNER_CONFIG, QK_K, ModelType
 from auto_round.logger import logger
 from auto_round.schemes import QuantizationScheme, get_gguf_scheme, preset_name_to_scheme
-from auto_round.utils import check_to_quantized
+from auto_round.utils import check_to_quantized, to_standard_regex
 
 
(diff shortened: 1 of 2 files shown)


auto_round/compressors/utils.py
```diff
@@ -421,7 +424,11 @@ def normalize_item(item: Union[str, dict, "QuantizationScheme"], layer_name: str
         safetensor_only_matched = [ln for ln in safetensor_only_names if regex.search(ln)]
         # skip it for mtp layers not loaded in transformers
         if not matched and not safetensor_only_matched:
-            raise ValueError(f"Invalid '{name}' in layer_config, no match found.")
+            # type(mlp.gate) is Qwen3VLMoeTextTopKRouter instead of Linear
+            logger.warning(
+                f"Layer name or regex '{name}' in layer_config does not match any supported layers. "
+                + "Please check for typos or update the regex pattern."
```
wenhuach21: I think ValueError is better

add "ignore it for now." at the end

xin3he: I also lean more towards using `ValueError`. However, currently, if a user passes the `mlp.gate` parameter directly from the Model Card, they will encounter an error. Do you consider this acceptable? Should we update this?

PR #1564, opened 2026-03-18
add rtn quant func

mengniwang95:
## Description

add rtn quant func

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring
- [ ] Other (please specify):

For Qwen2.5-1.5B-Instruct

Before:
device.py L1642: 'peak_ram': 4.32GB, 'peak_vram': 0.42GB

After:
device.py L1642: 'peak_ram': 4.32GB, 'peak_vram': 0.38GB

### DIFF
--- a/auto_round/data_type/utils.py
+++ b/auto_round/data_type/utils.py
@@ -133,9 +133,6 @@ def get_quant_func(
     def pad_bits(data_type):
         return data_type + str(bits)
 
-    if enable_rtn:
-        dtype = "rtn_" + dtype
-
     if not disable_opt_rtn:
         opt_rtn_data_type = "opt_rtn_" + dtype
         data_types = [


auto_round/data_type/fp8.py
```diff
@@ -70,6 +69,47 @@ def quant_block_fp_sym(tensor, max_scale=1.0, tensor_max=None, group_size=(128,
     return qdq_res, scale, None
 
 
+@register_dtype(("block_rtn_fp8_sym", "block_rtn_fp8", "block_rtn_fp8_e4m3"))
```
wenhuach21: rtn should be placed first right? otherwise, the code could not find this function

auto_round/compressors/base.py
```diff
@@ -1273,6 +1274,7 @@ def _quantize_layer_via_rtn(self, name: str, dtype: torch.dtype = None, to_cpu=T
                         enable_norm_bias_tuning=False,
                         enable_round_tuning=False,
                         enable_torch_compile=self.enable_torch_compile,
+                        enable_rtn=self.iters == 0,
```
wenhuach21, line 1277: I haven’t seen any routing code that directs to opt_rtn; the original code only contains the rtn routine

auto_round/data_type/fp8.py
mengniwang95: https://github.com/intel/auto-round/blob/c66eacdf00c558dfe4bd4e85b43d5eca1995d9fa/auto_round/data_type/utils.py#L136
`rtn_` is added first, then `block_` is added, so the dtype_name should be `block_rtn_xxx`

wenhuach21: then please change the order and follow origin name style to keep it as simple as possible

mengniwang95: updated, please check

PR #1571, opened 2026-03-19
fix deadlock issue when do quantization on hpu.

lkk12014402:
## Description

fix deadlock issue when do quantization on hpu.

### DIFF
--- a/auto_round/__main__.py
+++ b/auto_round/__main__.py
@@ -740,7 +740,11 @@ def tune(args):
                     suffix = f"a{autoround.act_bits}"
             else:
                 suffix = f"g{autoround.group_size}"
-        export_dir = os.path.join(args.output_dir, model_name.split("/")[-1] + f"-w{autoround.bits}{suffix}")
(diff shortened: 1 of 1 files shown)


auto_round/export/export_to_llmcompressor/export_to_static_fp.py
```diff
@@ -154,17 +155,22 @@ def save_quantized_as_static_fp(
     if not torch.cuda.is_available() and not torch.xpu.is_available():
         max_workers = 2  ## 2 with cuda packing will cause hang occasionally
```
wenhuach21, line 156: just set max_works=1 on hpu?

lkk12014402: not work

wenhuach21: ok, there are many places contain similar code in exporting folder, please help fix them as well

wenhuach21: May I know why this was marked as resolved? There are several formats that contain similar code.

Issue #1577, opened 2026-03-19
nit: This function is a bit confusing; better use two separate functions, one to check the `triton` and another for `mxfp4_forward_kernel_wrapper `.


chensuyue:
nit: This function is a bit confusing; better use two separate functions, one to check the `triton` and another for `mxfp4_forward_kernel_wrapper `.

_Originally posted by @yiliu30 in https://github.com/intel/auto-round/pull/1515#discussion_r2938908723_

PR #1607, opened 2026-03-24
refine transform config api.

lkk12014402:
https://github.com/intel/auto-round/issues/1577
## Description

refine transform config api, which supports `str|dict|TransformConfig|None`


auto_round/compressors/base.py
```diff
@@ -201,6 +205,7 @@ def __init__(
         disable_opt_rtn: bool | None = None,
         seed: int = 42,
         low_cpu_mem_usage: bool = True,
+        transform_config: str | dict | TransformConfig | None = None,
```
wenhuach21: 1 As mentioned multiple times, transform_config is not a clear name. It’s ambiguous what transformation it refers to. If this configuration is only for Hadamard, it should be renamed to hadamard_config.
2 When will support for real random matrices be added?
3 Please add a warning that this is an experimental feature and currently only supported for inference with the transformers backend.
4 do we have more accuracy data?

auto_round/compressors/base.py
```diff
@@ -56,6 +57,9 @@
 )
 from auto_round.data_type import QUANT_FUNC_WITH_DTYPE
 from auto_round.data_type.utils import reshape_pad_tensor_by_group_size, update_block_global_scale_if_needed
+
+_normalize_transform_config = transform_helper._normalize_transform_config
```
wenhuach21: If you prefix it with _, it typically implies that it is intended for internal use. However, that does not seem to be the case here.

wenhuach21:
As also mentioned before, these two args in transform_config should not be exposed to users. 
~~~python

  location: str = Field(default="weight", exclude=True)

  # apply transform inside modules for nvfp4, autoround tuning etc.
  need_calibration: bool = Field(default=False, exclude=True)
~~~

auto_round/compressors/base.py
lkk12014402: fixed

auto_round/compressors/base.py
lkk12014402: 1. renamed to hadamard_config
2. real random matrices doesn't have enough time to test for this release
3. will add this in the doc
4. will update `qwen3-8b` accuracy data in the pr for doc

docs/step_by_step.md
```diff
@@ -621,6 +622,114 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
...
+
+The Hadamard transform uses a special square matrix with entries +1 or -1, where all rows are mutually orthogonal. This property makes it useful for "spreading" information across the weight/activation matrix, which can help reduce quantization error.
+
+#### Implementation
+
+AutoRound provides two types of Hadamard transforms:
+
+1. **Deterministic Hadamard Transform** (`hadamard`): Uses Sylvester's construction to create a deterministic Hadamard matrix. The size must be a power of 2.
+
+2. **Random Hadamard Transform** (`random_hadamard`): Uses known Hadamard matrices from N. J. A. Sloane's Library of Hadamard Matrices. Supports non-power-of-2 sizes and deterministic seeding.
```
wenhuach21, line 639: Better refine these terms, because it's difficult to name the real random rotation for each layer in the future

docs/step_by_step.md
```diff
@@ -621,6 +622,114 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
...
+
+```python
+from auto_round.experimental.transform.utils.hadamard import (
+    deterministic_hadamard_matrix,
+    random_hadamard_matrix,
+    is_pow2,
+)
+```
+
+- `deterministic_hadamard_matrix(size, dtype, device)`: Creates a deterministic Hadamard matrix using Sylvester's construction
```
wenhuach21: I guess users does not care about the datails in Key Functions and Usage in Transforms

docs/step_by_step.md
```diff
@@ -621,6 +622,114 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
...
+from auto_round.experimental.transform.hadamard_config import HadamardConfig
+
+hadamard_config = HadamardConfig()
+
+model = apply_hadamard_transform(
+    model,
+    hadamard_config,
+)
+
+ar = AutoRound(model, scheme="MXFP4", iters=0, hadamard_config=hadamard_config)
```
wenhuach21: The API is weird,  no need to let users call apply_hadamard_transform explicitly. could you refine it to this way
~~~python
ar = AutoRound("qen3-0.6B", scheme="MXFP4", hadamard_config="default")
~~~
or 
~~~python
ar = AutoRound("qen3-0.6B", scheme="MXFP4", hadamard_config=hadamard_config)
~~~

docs/step_by_step.md
```diff
@@ -621,6 +622,114 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
...
+| Class | Description |
+|-------|-------------|
+| `HadamardTransform` | Applies deterministic Hadamard transform to weights |
+| `RandomHadamardTransform` | Applies random Hadamard transform with optional seeding |
+
+#### Parameters
+
+| Parameter | Description |
+|-----------|-------------|
+| `transform_block_size` | Size of the transformation block (default: 32) |
```
wenhuach21: also please refine this API. I think the major args users care are transform_block_size and  seed
Others should be determined by AR configuration, eg. module_type should be aligned with AR supported layers, device should use ar.device or cpu when low_gpu_mem_usage is enabled  if the matrix is large.

lkk12014402: updated

docs/step_by_step.md
lkk12014402: updated the api as our discussion offline

auto_round/compressors/base.py
```diff
@@ -552,7 +556,17 @@ def __init__(
             except (ImportError, ModuleNotFoundError):
                 logger.error("algorithm extension import error, fallback to default mode")
 
-        self.transform_config = kwargs.pop("transform_config", {})
+        # apply hadamard transform
+        self.hadamard_config = normalize_hadamard_config(hadamard_config)
```
wenhuach21: move this line to 562

auto_round/experimental/transform/hadamards.py
```diff
@@ -118,13 +118,13 @@ def _create_weight(
         return nn.Parameter(data, requires_grad=False)
 
 
-TRANSFORMS = {
+HADAMARDS = {
     "identity": IdentityTransform,
```
wenhuach21: what's the identify for? could we delete it?

auto_round/experimental/transform/helper.py
```diff
@@ -0,0 +1,97 @@
+# Copyright (C) 2026 Intel Corporation
```
wenhuach21: how about renameing this file to utils.py

auto_round/experimental/transform/helper.py
```diff
@@ -0,0 +1,97 @@
...
+
+    try:
+        from auto_round.experimental.transform.triton.mxfp4 import mxfp4_forward_kernel_wrapper  # pylint: disable=E0401
+    except Exception:
+        return False
+
+    return True
+
+
+def normalize_hadamard_config(hadamard_config: Any) -> dict[str, Any]:
```
wenhuach21: Any->None|dict|xxx|xxx

auto_round/inference/backend.py
```diff
@@ -775,10 +775,10 @@ def dynamic_import_inference_linear(backend, config):
     if "torch_mxfp8" in backend:
         return ar_qmodules.MXFP8QuantLinear
     if "torch_mxfp4" in backend:
-        transform_config = getattr(config, "transform_config", None)
-        if transform_config is not None and transform_config:
-            if transform_config["transform_type"] == "random_hadamard":
-                return ar_qmodules.TransformMXFP4QuantLinear
+        hadamard_config = getattr(config, "hadamard_config", None)
```
wenhuach21, line 778: MXFP8 is also claimed to be supported, but the same implementation has not been applied to the torch_mxfp8 backend.

docs/step_by_step.md
```diff
@@ -621,6 +622,67 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
   Include the flag `--adam`. Note that AdamW is less effective than sign gradient descent in many scenarios we tested.
 
 
+### Hadamard Transform
```
wenhuach21, line 625: Hadamard Transform[Experimental Feature]

docs/step_by_step.md
```diff
@@ -621,6 +622,67 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
...
+
+```python
+from auto_round import AutoRound
+
+# Load a model (supports FP8/BF16/FP16/FP32)
+model_name_or_path = ""
+output_dir = ""
+
+# hadamard_config="default": block_size=32, hadamard_type="hadamard"
+ar = AutoRound(model_name_or_path, scheme="MXFP4", iters=0, hadamard_config="default")
```
wenhuach21: remove iters=0

docs/step_by_step.md
```diff
@@ -621,6 +622,67 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
...
+2. **Random Hadamard Transform** (`random_hadamard`): Uses known Hadamard matrices from N. J. A. Sloane's Library of Hadamard Matrices. Supports non-power-of-2 sizes and deterministic seeding.
+
+
+#### Quantization with Hadamard Transform
+
+```python
+from auto_round import AutoRound
+
+# Load a model (supports FP8/BF16/FP16/FP32)
+model_name_or_path = ""
```
wenhuach21: give an example

docs/step_by_step.md
```diff
@@ -621,6 +622,67 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
...
+|-------|-------------|
+| `HadamardTransform` | Applies deterministic Hadamard transform to weights |
+| `RandomHadamardTransform` | Applies random Hadamard transform with optional seeding |
+
+#### Parameters
+
+| Parameter | Description |
+|-----------|-------------|
+| `block_size` | Size of the transformation block (default: 32) |
+| `device` | Device to create the transform on |
```
wenhuach21: refine this to align with the latest code

docs/step_by_step.md
```diff
@@ -621,6 +622,67 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
...
+| `precision` | Data type for the transform matrix |
+| `location` | Location to apply transform ("weight") |
+| `module_type` | Module type (default: `torch.nn.Linear`) |
+| `inverse` | Whether to apply inverse transform |
+| `seed` | Random seed (for RandomHadamardTransform) |
+| `generator` | PyTorch generator for random values |
+
+#### Notes
+
+- The hadamard transform is its own inverse: H @ H = I
```
wenhuach21: if you think this is important to users,  wrap it with details as it's too long. Otherwise delete this scection

docs/step_by_step.md
wenhuach21, line 625: please update the Chinese doc as well

docs/step_by_step.md
```diff
@@ -621,6 +622,67 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
   Include the flag `--adam`. Note that AdamW is less effective than sign gradient descent in many scenarios we tested.
 
 
+### Hadamard Transform
+
+AutoRound supports Hadamard transform as an optional weight/activation transformation technique, which can improve quantization accuracy by rotating the weight/activation matrix. This is particularly useful for certain quantization scenarios.
+
+#### Overview
+
+The Hadamard transform uses a special square matrix with entries +1 or -1, where all rows are mutually orthogonal. This property makes it useful for "spreading" information across the weight/activation matrix, which can help reduce quantization error.
```
wenhuach21: This is too technical. It would be better to describe the scenarios where users should use this feature. For example, the Hadamard transform is known to help suppress activation outliers, so it can typically improve accuracy when act_bits < 8  with the cost of introducing a little overhead at inference. Besides, currently the serving framework's support for this feature is not very well

wenhuach21 (approved):

auto_round/inference/backend.py
lkk12014402: I haven't tested the mxfp8 scheme, remove it. will add it when I verify the scheme

auto_round/experimental/transform/helper.py
lkk12014402: move the file content to `auto_round/experimental/utils.py`

docs/step_by_step.md
lkk12014402: updated

docs/step_by_step.md
lkk12014402: updated

docs/step_by_step.md
lkk12014402: no change here

docs/step_by_step.md
lkk12014402: deleted

docs/step_by_step.md
lkk12014402: updated

PR #1611, opened 2026-03-25
fix fp8 accuracy drop issue

mengniwang95:
## Description

fp8 quant_func only use weight.max() for scale calculation when tuning

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

acc before:
|Tasks|Version|     Filter    |n-shot|  Metric   |  |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7286|± |0.0122|
|     |      |strict-match    |     5|exact_match|↑  |0.7316|± |0.0122|

acc after:
|Tasks|Version|     Filter    |n-shot|  Metric   |  |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7589|± |0.0118|
|     |      |strict-match    |     5|exact_match|↑  |0.7619|± |0.0117|

### DIFF
--- a/auto_round/compressors/diffusion/compressor.py
+++ b/auto_round/compressors/diffusion/compressor.py
@@ -326,6 +326,7 @@ class DiffusionCompressor(BaseCompressor):
 
         if self.pipe.device != self.model.device:
             self.pipe.to(self.model.device)
+        self.pipe.to(self.model.dtype)
         with tqdm(range(1, total + 1), desc="cache block inputs") as pbar:
             for ids, prompts in self.dataloader:
                 if isinstance(prompts, tuple):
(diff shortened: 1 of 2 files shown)


auto_round/compressors/base.py
```diff
@@ -473,6 +473,9 @@ def __init__(
             )
             disable_opt_rtn = False
 
+        if self.iters > 0 and is_block_wfp8(self):
+            logger.warning("RTN is recommended since it shows comparable accuracy to tuning for block-wise fp8 quantization.")
```
wenhuach21: Use --iters 0 --disable_opt_rtn to enable RTN mode.

wenhuach21: RTN is recommended, as it achieves accuracy comparable to tuning for block-wise FP8 quantization while being significantly faster.

PR #1634, opened 2026-03-27
[Draft] Support TurboQuant KV-cache quantization

lvliang-intel:
## Description

We have implemented a working TurboQuant KV-cache prototype in AutoRound with both algorithm-side and runtime-side support.
We used two references for this work: the [TurboQuant paper] (https://arxiv.org/abs/2504.19874) (ICLR 2026, Google Research) and the vLLM TurboQuant PR https://github.com/vllm-project/vllm/pull/38280, and they influenced different parts of the implementation. 

What is implemented:
1 Core TurboQuant quantization pipeline
2 QJL residual correction
3 GPU acceleration
4 Runtime cache modes
  a. Pre-dequant mode: quantize then immediately dequantize and store bf16 KV tensors in a standard cache
  b. Packed mode: store bit-packed KV cache with explicit unpack + reconstruction on read
  Both modes are wired through the TurboQuant KV-cache runtime builder

Currently the test result for TurboQuant 4bit is good, but TurboQuant 3 bit and 2 bit still have problem.
TurboQuant decode is still outside attention, so HBM traffic is still too high.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #


wenhuach21:
You’re moving really fast! 

That said, I don’t think this feature brings clear benefits to AR at the moment. The main advantage seems to be eval accuracy with quantized kv cache. However, for Transformers, KV cache quantization at the prefill stage is difficult, which could introduce bias.

PR #1641, opened 2026-03-31
[research feature] Introduce INT4 support at the algorithm level

wenhuach21:
## Description

- [ ] delay rotation to save ram/vram
- [ ] inference
- [ ] automodel and automodelforcacuallm should be different


auto_round/compressors/base.py
```diff
@@ -557,18 +557,33 @@ def __init__(
...
+            # from auto_round.experimental.utils import normalize_hadamard_config
+
+            #
+            # _hcfg = normalize_hadamard_config(hadamard_config)
+            # if not isinstance(_hcfg, HadamardConfig):
+            #     _hcfg = HadamardConfig(**_hcfg)
+            # group_size = _hcfg.block_size  # e.g. 32 for MXFP4
+
+            self.model.to("cuda")
+            apply_hadamard_rotation(self.model, group_size=self.group_size, allow_online_hadamard=False)
```
wenhuach21:
```suggestion
            apply_hadamard_rotation(self.model, group_size=self.group_size, allow_online_hadamard=True)
```

wenhuach21:
please have a review

auto_round/compressors/utils.py
```diff
@@ -107,14 +107,24 @@ def is_wint_a16(ar: Callable) -> bool:
...
     if is_wint8aint8(ar_or_format):
         return True
     return False
 
 
+def is_wint4aint4(ar_or_scheme: Union[str, Callable]):
+    if isinstance(ar_or_scheme, str):
+        return "int4" in ar_or_scheme.lower()
+    elif (
+        "int4" in ar_or_scheme.act_data_type or ("int" in ar_or_scheme.act_data_type and ar_or_scheme.act_bits == 4)
```
hshen14, line 122: It assumes act_data_type is always non-None. Maybe we can add assertion here.

docs/environments.md
```diff
@@ -69,6 +69,26 @@ export AR_DISABLE_OFFLOAD=1
 export AR_DISABLE_DATASET_SUBPROCESS=true
 ```
 
+### AR_ACT_SCALE
+- **Description**: Controls the scaling factor applied to activation min/max values during activation quantization. A value less than 1.0 shrinks the clipping range, which can reduce outlier impact.
+- **Default**: `1.0`
+- **Valid Values**: Any float, e.g. `0.8`, `0.9`, `1.0`
```
hshen14: Inconsistent with CN version? 1.0 vs. 0.9

Also the example 0.9 vs. 0.8

auto_round/inference/convert_model.py
```diff
@@ -680,19 +680,19 @@ def convert_hf_model(model: nn.Module, target_device: str = "cpu") -> tuple[nn.M
...
         # apply forward hook
-        act_hadamard_config = HadamardConfig(
-            block_size=hadamard_config["block_size"],
-            hadamard_type=hadamard_config["hadamard_type"],
+        act_rotation_config = RotationConfig(
+            block_size=rotation_config["block_size"],
+            hadamard_type=rotation_config["hadamard_type"],
         )  # apply to activation
-        model = apply_hadamard_transform(
+        model = apply_rotation_transform(
```
hshen14: should use tuple (model, _)?

auto_round/compressors/utils.py
wenhuach21: That's true , `act_data_type` will be overridden by the weight data type if it’s not explicitly set. I can refine this behavior.

lkk12014402 (approved):
LGTM

lkk12014402:
By the way, is the implementation of Quarot consistent with community implementations like llm-compressor? Have we compared the accuracy?

wenhuach21:
> By the way, is the implementation of Quarot consistent with community implementations like llm-compressor? Have we compared the accuracy?

I’m not entirely sure. The Quarot accuracy is evaluated using their own code, while we focus specifically on the Hadamard transformation, where I believe our implementation is consistent with theirs. One key difference is that they use a random Hadamard transformation, whereas we use a fixed Hadamard, which we’ve found to perform better.

wenhuach21:
[0374bd2](https://github.com/intel/auto-round/pull/1641/commits/0374bd20d3e1b1510e3f7350720989f4a25510f4) accuracy good
56505e8
03c861b has issues

PR #1642, opened 2026-03-31
support WOQ model input, such as kimi2.5

xin3he:
## Description

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ ar.quantize_and_save(output_dir="./qmodel", format="auto_round")
 | **auto_round**     | W4A16(Recommended), W2A16, W3A16, W8A16, W2A16G64, W2A16G32, `MXFP4`, `MXFP8`, `MXFP4_RCEIL`, `MXFP8_RCEIL`, `NVFP4`, `FPW8A16`, `FP8_STATIC`, `BF16`                     |
 | **auto_awq**       | W4A16(Recommended), BF16                                                                                                                                                  |
 | **auto_gptq**      | W4A16(Recommended), W2A16, W3A16, W8A16, W2A16G64, W2A16G32,BF16                                                                                                          |
-| **llm_compressor** | NVFP4(Recommended), `MXFP4`, `MXFP8`, `FPW8A16`, `FP8_STATIC`                                                                                                             |
+| **llm_compressor** | NVFP4(Recommended), `MXFP4`, `MXFP8`, `FPW8A16`, `FP8_STATIC`, `FP8_BLOCK`, `INT8_W8A8`, `W4A16`, `W8A16`                                                                   |
 | **gguf**           | GGUF:Q4_K_M(Recommended), GGUF:Q2_K_S, GGUF:Q3_K_S, GGUF:Q3_K_M, GGUF:Q3_K_L, GGUF:Q4_K_S, GGUF:Q5_K_S, GGUF:Q5_K_M, GGUF:Q6_K, GGUF:Q4_0, GGUF:Q4_1, GGUF:Q5_0, GGUF:Q5_1,GGUF:Q8_0 |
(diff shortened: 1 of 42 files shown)


auto_round/utils/weight_handler.py
```diff
@@ -180,6 +182,15 @@ def detect_layer(self, module: torch.nn.Module) -> bool:
         """
         pass
 
+    def attach_weight_shape(self, module: torch.nn.Module):
```
yiliu30, line 185: This part is a bit hard to follow. Why do we need a new attribute to track the weight shape?

xin3he: AutoRound will check the weight shape at the beginning, such as divisible_by_32.
The quantized model won't have weight attribution, the weight is packed to weight_packed.

Therefore, we need to attach meta-weights to the verification logic, as the dequantization process is performed block-by-block.

yiliu30: Could you explain why auto-round needs to check the shape and point me to the relevant code? 
Also, if needed, do we clear these attributes?

auto_round/utils/weight_handler.py
```diff
@@ -297,6 +308,7 @@ def check_and_mark_quantized_module(model: torch.nn.Module) -> Set[ModuleWeightT
     for weight_type, handler in _WEIGHT_TYPE_HANDLERS.items():
         # Check model itself first
         if handler.detect_layer(model):
+            handler.attach_weight_shape(model)
```
lvliang-intel, line 311: `attach_weight_shape` applied to all handlers unconditionally. The base class method assumes all modules have out_features / in_features attributes. For handlers like GGUFHandler or FP8Handler where the module might be a regular torch.nn.Linear (already has .weight), this is harmless but semantically unnecessary.  Add a guard check?

auto_round/utils/weight_handler.py
```diff
@@ -819,3 +867,68 @@ def convert_layer(
...
+# ----------------------------------------------------------------------------
+# WOQ Handler - CompressedLinear with weight-only quantization
+# ----------------------------------------------------------------------------
+
+
+@register_weight_type_handler(ModuleWeightType.WOQ)
+class WOQHandler(WeightTypeHandler):
+    """Handler for integer 4-bit weight-only quantized layers (Compressed Tensor only)."""
+
+    def detect_layer(self, module: torch.nn.Module) -> bool:
```
lvliang-intel, line 881: Duplicate logic. The two branches (CompressedLinear class check and quantization_scheme attribute check) share the same num_bits in [2, 4, 8] / type == "int" / input_activations is None condition. Consider extracting a helper _is_woq_scheme(q_scheme) to deduplicate.

auto_round/utils/weight_handler.py
xin3he: https://github.com/intel/auto-round/blob/f81e20d678289bb75533b3aefd61100e181b4678/auto_round/formats.py#L151
It's used to update layer_config at the beginning to avoid quantizing un-supported shape.

auto_round/utils/weight_handler.py
xin3he: It's not unconditional, `handler.detect_layer(model)` is the filter, only quantized module without weight will be attached.

auto_round/utils/weight_handler.py
xin3he: Thanks, it's because that compressed-tensors provide two versions, the old one is using CompressedLinear and the other is using Linear with hook.

PR #1685, opened 2026-04-15
fix: pass `dynamic=True` to torch.compile to stop dynamo recompilation

### LINKED ISSUE #1679: [Bug]: torch._dynamo hit config.recompile_limit (8) ... tensor 'v' size mismatch at index 0 ...
### Problem Description

When converting any model, if `enable_torch_compile` is set to True (with version v0.12.2), I receive a bunch of messages from `torch._dynamo` regarding an error. The conversion still completes, but the output is flooded with the compilation messages shown below.

```python
2026-04-14 02:40:04 INFO base.py L1835: caching done
Quantizing model.layers.0:   0%|                                                                        | 0/28 [00:00<?, ?it/s]/home/daved/miniforge3/envs/autoround/lib/python3.12/site-packages/torch/autograd/graph.py:869: UserWarning: Flash Attention defaults to a non-deterministic algorithm. To explicitly enable determinism call torch.use_deterministic_algorithms(True, warn_only=False). (Triggered internally at /pytorch/aten/src/ATen/native/transformers/cuda/attention_backward.cu:124.)
  return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
W0414 02:40:14.779000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8] torch._dynamo hit config.recompile_limit (8)
W0414 02:40:14.779000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8]    function: 'quant_tensor_sym' (/home/daved/miniforge3/envs/autoround/lib/python3.12/site-packages/auto_round/data_type/int.py:118)
W0414 02:40:14.779000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8]    last reason: 0/7: tensor 'v' size mismatch at index 0. expected 16384, actual 24576
W0414 02:40:14.779000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8] To log all recompilation reasons, use TORCH_LOGS="recompiles".
W0414 02:40:14.779000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8] To diagnose recompilation issues, see https://docs.pytorch.org/docs/main/user_guide/torch_compiler/compile/programming_model.recompilation.html
quantized 7/7 layers in the block, loss iter 0: 0.001117 -> iter 175: 0.000318,'peak_ram': 8.91GB, 'peak_vram': 3.43GB
Quantizing model.layers.1:   4%|██▎                                                             | 1/28 [00:11<04:58, 11.06s/it]W0414 02:40:22.531000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8] torch._dynamo hit config.recompile_limit (8)
W0414 02:40:22.531000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8]    function: 'quant_tensor_sym' (/home/daved/miniforge3/envs/autoround/lib/python3.12/site-packages/auto_round/data_type/int.py:118)
W0414 02:40:22.531000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8]    last reason: 0/7: tensor 'v' size mismatch at index 0. expected 16384, actual 24576
W0414 02:40:22.531000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8] To log all recompilation reasons, use TORCH_LOGS="recompiles".
W0414 02:40:22.531000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8] To diagnose recompilation issues, see https://docs.pytorch.org/docs/main/user_guide/torch_compiler/compile/programming_model.recompilation.html
quantized 7/7 layers in the block, loss iter 0: 0.001711 -> iter 128: 0.000567,'peak_ram': 8.94GB, 'peak_vram': 3.43GB
Quantizing model.layers.2:   7%|████▌                                                           | 2/28 [00:18<03:56,  9.09s/it]W0414 02:40:30.431000 88614 site-packages/torch/_dynamo/convert_frame.py:1743] [0/8] torch._dynamo hit config.recompile_limit (8)
```

### Reproduction Steps

1 - 4. Run the following command to see the error:

```
auto-round \
    --model Qwen/Qwen3-0.6B \
    --scheme "W4A16" \
    --format "auto_round" \
    --enable_torch_compile \
    --output_dir ./tmp_autoround
```

### Environment Information

- OS: Ubuntu 24.04.4
- Python Version: 3.12.13
- AutoRound Version: v0.12.2
- Hardware: RTX 6000 Ada
- Torch Version: 2.11.0

### Error Logs

```shell

```

### Additional Context

_No response_

rahulp7801:
Fixes #1679.

## Summary

`compile_func_on_cuda_or_cpu` in `auto_round/utils/device.py:97` calls `torch.compile(func)` with no shape-specialisation hint. When `quant_tensor_sym` is compiled, `torch._dynamo` traces it for the concrete shape of the `v` tensor (the per-weight rounding perturbation, initialised to the weight's shape at `wrapper.py:164`). Different layers have different weight shapes, so dynamo recompiles the function for each new shape. After 8 recompilations it logs the `config.recompile_limit` warning and falls back to eager mode. Passing `dynamic=True` to `torch.compile` tells dynamo to use symbolic shapes, producing a single shape-generic trace that never needs to recompile for size changes in `v` or the weight tensor.

## Files changed

- `auto_round/utils/device.py`

## Test plan

- **Call sites checked**: `compile_func` is called in `wrapper.py:175,182`, `alg_ext.py:373,1042,1046`, and `compressors/base.py:539`. All of these wrap quantisation functions (`quant_tensor_sym`, `quant_tensor_asym`, and `block_forward`) that receive tensors of varying shape across layers — so `dynamic=True` is the correct default for all of them.

<details><summary>Contribution guide notes</summary>

- **Sign-off required**: Every commit must include `Signed-off-by: Name <email>` (DCO, per CONTRIBUTING.md).

</details>

### DIFF
--- a/auto_round/inference/convert_model.py
+++ b/auto_round/inference/convert_model.py
@@ -179,6 +179,43 @@ def get_available_devices():
     return devices
 
 
+def _remap_paths_for_text_model(model, quant_block_list, extra_config):
+    """Remap quantization paths when a composite model checkpoint is loaded as its text sub-model.
+
+    Uses Transformers' conversion_mapping WeightRenaming rules (e.g. "model.language_model" -> "model")
+    to fix path mismatches. Returns (remapped_block_list, remapped_extra_config).
+    """
+    try:
+        from transformers.conversion_mapping import get_checkpoint_conversion_mapping
+    except ImportError:
+        return quant_block_list, extra_config
+
+    model_type = getattr(getattr(model, "config", None), "model_type", None)
+    if model_type is None:
+        return quant_block_list, extra_config
+
+    mapping = get_checkpoint_conversion_mapping(model_type)
+    renamings = [r for r in mapping if type(r).__name__ == "WeightRenaming"]
+    if not renamings:
+        return quant_block_list, extra_config
+
+    def remap(path):
+        for r in renamings:
+            for src, tgt in zip(r.source_patterns, r.target_patterns):
+                new_path = re.sub(src, tgt, path)
+                if new_path != path:
+                    return new_path
+        return path
+
+    new_block_list = [remap(b) for b in quant_block_list]
+    new_extra_config = {remap(k): v for k, v in extra_config.items()} if extra_config else extra_config
+
+    if new_block_list != quant_block_list:
+        logger.info(f"Remapped block_name_to_quantize: {quant_block_list} -> {new_block_list}")
+
+    return new_block_list, new_extra_config
+
+
 def get_layer_config(model, quantization_config):
     """
     get a layer-wise quantization configuration for a given model.
@@ -258,6 +295,17 @@ def get_layer_config(model, quantization_config):
 
     # Load extra configuration if available
     extra_config = getattr(quantization_config, "extra_config", {})
+
+    # When a composite model (e.g. VLM) is loaded as its text sub-model via AutoModelForCausalLM,
+    # block_name_to_quantize may still reference composite-level paths (e.g. "model.language_model.layers")
+    # while the actual module paths are "model.layers". Use conversion_mapping to remap if no layers matched.
+    if not layer_names and quant_block_list:
+        quant_block_list, extra_config = _remap_paths_for_text_model(model, quant_block_list, extra_config)
+        for n, m in model.named_modules():
+            if type(m) not in SUPPORTED_LAYER_TYPES:
+                continue
+            if check_start_with_block_name(n, quant_block_list):
+                layer_names.append(n)
 
     # Process GPTQ format: identify modules that should be quantized
     if getattr(quantization_config, "modules_in_block_to_quantize", None):


auto_round/utils/device.py
lvliang-intel, line 93: this torch.compie please also add 'dynamic=true'

wenhuach21:
Thanks for the detailed explanation in the PR, I finally understand it.

I asked chatgpt, and it mentioned that enabling `dynamic` can cause performance drops. Since the number of shapes within a block is usually limited, would it make sense to increase the recompile limit to better match the model? If the number of shapes becomes too large, say 20 or 30, then perhaps we could fall back to using `dynamic`.

rahulp7801:
It is a good point about the performance trade-off. But I think, increasing the recompile limit could lead to unbounded memory usage and compilation overhead in cases with many shape variations, which is harder to predict and debug. Using `dynamic=True` provides a predictable albeit slightly slower fallback that guarantees stability. We could revisit this if there are specific performance bottlenecks in practice but for now this approach looks like it gives us a reasonable default.

auto_round/utils/device.py
```diff
@@ -90,12 +90,12 @@ def _use_hpu_compile_mode():
...
 def compile_func_on_hpu(func):
     if _use_hpu_compile_mode():
-        return torch.compile(func, backend="hpu_backend")
+        return torch.compile(func, backend="hpu_backend", dynamic=True)
     return func
 
 
 def compile_func_on_cuda_or_cpu(func):
-    return torch.compile(func)
+    return torch.compile(func, dynamic=True)
```
yiliu30: Hi @rahulp7801, thanks for the PR!
Per the PyTorch documentation, this approach is mainly intended for testing purposes and isn’t recommended for production: https://docs.pytorch.org/docs/stable/user_guide/torch_compiler/torch.compiler_dynamic_shapes.html#torch-compile-dynamic-true-not-recommended
How about using the annotation-based approach instead? 
https://docs.pytorch.org/docs/stable/user_guide/torch_compiler/torch.compiler_dynamic_shapes.html#user-annotations

yiliu30 (changes requested):

auto_round/utils/device.py
rahulp7801: I decided to remove `dynamic=True` in my latest commit. Since these are generic wrapper functions without visibility into the actual functions being compiled or their input shapes, using the global `dynamic=True` flag is too broad and can cause unnecessary recompilations.

I think the better approach is to let callers handle dynamic shape requirements at the call sites using proper annotations like `torch._dynamo.mark_dynamic()` or `torch.export.Dim` when needed, which will give finer control and follows PyTorch's recommended patterns for production code.

Issue #1674, opened 2026-04-10
[Bug]: rename scheme INT8_W8A8 to INT8


wenhuach21:
### Problem Description

~

### Reproduction Steps

~

### Environment Information

~

### Error Logs

```shell
~
```

### Additional Context

~

thuang6:
Done

PR #1687, opened 2026-04-15
rename scheme INT8_W8A8 to INT8

thuang6:
## Description

Rename scheme INT8_W8A8 to INT8, including code, test, doc.  

Use INT8 in doc, but also keep backward compatibility of using old name.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring
- [ ] Other (please specify):

## Related Issues

Fixes or relates to #1674

### DIFF
--- a/auto_round/formats.py
+++ b/auto_round/formats.py
@@ -350,7 +350,18 @@ class FakeFormat(OutputFormat):
 
 @OutputFormat.register("llm_compressor")
 class LLMCompressorFormat(OutputFormat):
-    support_schemes = ["MXFP4", "MXFP8", "NVFP4", "FPW8A16", "FP8_STATIC", "INT8", "INT8_W8A8", "FP8_BLOCK", "W4A16", "W8A16"]
+    support_schemes = [
+        "MXFP4",
+        "MXFP8",
+        "NVFP4",
+        "FPW8A16",
+        "FP8_STATIC",
+        "INT8",
+        "INT8_W8A8",
+        "FP8_BLOCK",
+        "W4A16",
+        "W8A16",
+    ]
     format_name = "llm_compressor"
 
     def __init__(self, format, ar):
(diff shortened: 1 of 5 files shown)


auto_round/formats.py
```diff
@@ -350,7 +350,18 @@ def save_quantized(
...
 class LLMCompressorFormat(OutputFormat):
-    support_schemes = ["MXFP4", "MXFP8", "NVFP4", "FPW8A16", "FP8_STATIC", "INT8_W8A8", "FP8_BLOCK", "W4A16", "W8A16"]
+    support_schemes = [
+        "MXFP4",
+        "MXFP8",
+        "NVFP4",
+        "FPW8A16",
+        "FP8_STATIC",
+        "INT8",
+        "INT8_W8A8",
```
wenhuach21, line 360: dupliacated?

auto_round/formats.py
```diff
@@ -69,7 +69,7 @@ class AutoRoundExportFormat(str, Enum):
     NV_FP = "nv_fp"
     MX_FP_RCEIL = "mx_fp_rceil"
     NV_FP4_WITH_STATIC_GS = "nv_fp4_with_static_gs"
-    INT8_W8A8 = "int8_w8a8"
+    INT8 = "int8_w8a8"
```
wenhuach21, line 72: should change the value to INT8 as well?

auto_round/formats.py
thuang6: This is kept for backward compatibility since it was already in previous releases

auto_round/formats.py
wenhuach21: it’s probably better to deprecate it. If a user provides int8_w8a8, we can trigger a warning and automatically map it to INT8 instead

thuang6: make sense,  updated  with deprecated warning.  INT8_W8A8 has mapped to INT8 in PRESET_SCHEME.    Here value is used to map with compressed_tensor since its format uses this name

wenhuach21 (approved):

auto_round/formats.py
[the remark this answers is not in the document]
thuang6: this is intended to map INT8 scheme to INT8_W8A8 scheme in compressed-tensors

Issue #1783, opened 2026-05-07
[Bug]: TypeError: Gemma4TextDecoderLayer.forward() got multiple values for argument 'hidden_states'


3dluvr:
### Problem Description

When trying to quantize the google/gemma-4-31b model to INT8 W8A16, the process crashes complaining about multiple **hidden_states** argument being provided.

### Reproduction Steps

1. Quantization script provided below
2. Model: google/gemma-4-31b-it from HF
3. Dataset: NeelNanda/pile-10k from HF

### Environment Information

- Debian 13, headless
- Python 3.12, Cuda 13.0
- Auto-Round Nightly
- Epyc 7J13, 256GB RAM, 3 x RTX 3090, P2P Nvidia driver

### Error Logs

```shell
quantized 7/7 layers in the block, loss iter 0: 0.000013 -> iter 954: 0.000009
2026-05-06 18:48:07 INFO device.py L1802: 'peak_ram': 152.55GB, 'peak_vram': {'0': 21.62GB, '1': 23.08GB, '2': 21.72GB}
Quantizing model.language_model.layers.1:   2%|█▎                                                                             | 1/60 [19:58<19:38:13, 1198.20s/it]Traceback (most recent call last):
  File "/home/user/models/quant-gemma-mixed.py", line 156, in 
    main()
  File "/home/user/models/quant-gemma-mixed.py", line 152, in main
    ar.quantize_and_save(OUTPUT_DIR, format="auto_round:auto_gptq")
  File "/home/user/Envs/llm/lib/python3.12/site-packages/auto_round/compressors_new/base.py", line 1322, in quantize_and_save
    self.quantize()
  File "/home/user/Envs/llm/lib/python3.12/site-packages/auto_round/compressors_new/calib.py", line 1154, in quantize
    self._quantize_blocks(
  File "/home/user/Envs/llm/lib/python3.12/site-packages/auto_round/compressors_new/calib.py", line 977, in _quantize_blocks
    reference_output = self.quantizer._get_block_outputs(m, input_ids, input_others, bs)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/Envs/llm/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1791, opened 2026-05-09
Reduce RAM usage of quantizing VLM models and fix some issues of quantizing gemma4

lvliang-intel:
## Description

Fix VLM/diffusion quantization OOM issues and Gemma4 multi-GPU crash.

Fixed the below issues:
1 VLM: OOM due to redundant hidden_states caching
2 Diffusion: OOM in multi-device quantization
3 MLLM: Memory fragmentation in calibration loop

CUDA_VISIBLE_DEVICES=1,2 auto-round /mnt/disk3/lvl/gemma-4-31B-it --device_map "auto"    --data_type "int"     --group_size 128     --batch_size 2     --nsamples 512     --seqlen 2048     --iters 2000     --to_quant_block_names 'model.language_model.layers'     --output_dir gemma-3-27B-it-INT8-AutoRound     --scheme W8A16     --dataset NeelNanda/pile-10k     --format "auto_round:auto_gptq"

quantized 6/6 layers in the block, loss iter 0: 0.000566 -> iter 1714: 0.000053
2026-05-09 05:20:41 INFO device.py L1802: 'peak_ram': 70.72GB, 'peak_vram': {'0': 37.81GB, '1': 29.42GB}
Quantizing done: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 60/60 [6:22:13<00:00, 382.23s/it]
2026-05-09 05:20:47 INFO device.py L1802: 'peak_ram': 70.72GB, 'peak_vram': {'0': 37.81GB, '1': 29.42GB}
2026-05-09 05:20:52 INFO shard_writer.py L314: model has been saved to gemma-3-27B-it-INT8-AutoRound/gemma-4-31B-it-w8g128/
2026-05-09 05:20:52 INFO calib.py L1260: quantization tuning time 22938.128324747086
2026-05-09 05:20:52 INFO calib.py L1279: Summary: quantized 410/602 in the model, unquantized layers: lm_head, model.embed_vision.embedding_projection, model.vision_tower.encoder.layers.[0-26].mlp.down_proj.linear, model.vision_tower.encoder.layers.[0-26].mlp.gate_proj.linear, model.vision_tower.encoder.layers.[0-26].mlp.up_proj.linear, model.vision_tower.encoder.layers.[0-26].self_attn.k_proj.linear, model.vision_tower.encoder.layers.[0-26].self_attn.o_proj.linear, model.vision_tower.encoder.layers.[0-26].self_attn.q_proj.linear, model.vision_tower.encoder.layers.[0-26].self_attn.v_proj.linear, model.vision_tower.patch_embedder.input_proj
2026-05-09 05:20:52 INFO device.py L1802: 'peak_ram': 70.72GB, 'peak_vram': {'0': 37.81GB, '1': 29.42GB}

## Type of Change

Bug fix

https://github.com/intel/auto-round/issues/1783


auto_round/compressors/base.py
```diff
@@ -1665,6 +1665,11 @@ def _quantize_via_rtn_blockwise(self, all_to_quantized_module_names: list[str])
         # Clear hooks for multi-GPU setups
         if hasattr(self.model, "hf_device_map") and len(self.model.hf_device_map) > 1:
             accelerate.hooks.remove_hook_from_submodules(self.model)
+            # Re-apply special model patches (e.g., Gemma4 position embedding
```
hshen14: do we have test to cover this feature? it seems a regression since the comment says "re-apply special model patches".

auto_round/compressors_new/calib.py
```diff
@@ -429,31 +426,22 @@ def calib(self, nsamples, bs):
...
                         repeated = False
                         while j >= 0 and input_ids[i, j] == last_token:
                             repeated = True
                             new_attention_mask[i, j] = 0
                             j -= 1
-                        # If there was at least one repeat, set last token mask to 0
                         if repeated:
                             new_attention_mask[i, -1] = 0
 
-                # Workaround: some models treat an all-1 attention mask as equivalent to None and
```
hshen14: Will it impact some models? better to check with the code owners

auto_round/compressors/base.py
lvliang-intel: Good point. `remove_hook_from_submodules` is intentionally called here to strip the accelerate dispatch hooks added during calibration. However, it is a blunt accelerate API that removes all hooks indiscriminately, including the Gemma4 layer-forward patches attached by `_patch_gemma4_model` via AccelerateHook. The re-apply is a pragmatic workaround: rather than patching accelerate internals to selectively remove only the calibration hooks, we restore the Gemma4 patches in a single line after the sweep. An alternative would be to store a reference to the Gemma4 patches before removal and re-attach them, but that adds complexity for the same end result.

auto_round/compressors_new/calib.py
lvliang-intel: The surrounding trailing-repeat detection code was simplified, and the comment was removed in the same pass. The `workaround` itself (new_attention_mask[:, -1] = 0, line 445) is still in place — only the explanatory comment was dropped. I want to take it as a solution, not a workaround. Otherwise looks like the workaround everywhere in our code.

auto_round/compressors_new/calib.py
```diff
@@ -381,9 +381,6 @@ def calib(self, nsamples, bs):
         else:
             self.dataloader = self.dataset
         total_cnt = 0
-        if self.dataloader.__class__.__name__ == "BatchEncoding":
-            self.dataloader = [self.dataloader.data]
```
wenhuach21: We should keep it, and I guess it has no side affect to your PR.

auto_round/compressors_new/calib.py
wenhuach21: didn’t see any actual simplification,  only that the comments were removed. We keep those comments to indicate that this logic is intentional and should not be deleted by future developers.

auto_round/utils/device.py
```diff
@@ -1881,12 +1881,59 @@ def dispatch_model_by_all_available_devices(
...
+
+        # Calculate memory needed for non-target pipeline components
+        non_main_bytes = 0
+        for attr, component in pipe.components.items():
+            if attr == main_attr:
+                continue
+            if isinstance(component, torch.nn.Module):
+                non_main_bytes += sum(p.numel() * p.element_size() for p in component.parameters())
+        # Add 20% buffer for activations and overhead
+        non_main_reserved = int(non_main_bytes * 1.2) if non_main_bytes > 0 else 0
```
wenhuach21: better name a var to 1.2

wenhuach21: In Transformers, the max_memory in #1903 applies a 0.9 ratio as a safeguard to avoid the issue.

auto_round/special_model_handler.py
```diff
@@ -205,11 +211,7 @@ def _handle_special_model(model):
 
         model.forward = partial(_qwen3_omni_moe_forward, model)
     if hasattr(model, "config") and model_type == "gemma4":
-        import transformers
-        from packaging import version
-
-        if version.parse(transformers.__version__) < version.parse("5.6"):
```
wenhuach21, line 213: This is intentional, since this monkey patching is inherently high-risk if Transformers changes its internal implementation. We should at least log a warning to make the behavior explicit and warn users it has only been validated in transtormers<xxx

auto_round/compressors_new/calib.py
```diff
@@ -587,6 +575,13 @@ def forward(m, hidden_states=None, *positional_inputs, **kwargs):
                         and key == "hidden_states"
                     ):
                         continue
+                    # Skip non-hidden_states tensor args that are per-sample constants.
+                    # Only hidden_states varies per sample and must be cached.
+                    # However, shared_cache_keys tensors (e.g. position_ids) must still
+                    # be stored once — they are needed by special model patches.
+                    if key not in ("hidden_states",) and isinstance(kwargs[key], torch.Tensor):
```
wenhuach21: this change should have high risk I guess

auto_round/compressors_new/calib.py
```diff
@@ -587,6 +575,13 @@ def forward(m, hidden_states=None, *positional_inputs, **kwargs):
                         and key == "hidden_states"
                     ):
                         continue
+                    # Skip non-hidden_states tensor args that are per-sample constants.
+                    # Only hidden_states varies per sample and must be cached.
+                    # However, shared_cache_keys tensors (e.g. position_ids) must still
+                    # be stored once — they are needed by special model patches.
```
wenhuach21: a simpler fix is removing this https://github.com/intel/auto-round/blob/main/auto_round/special_model_handler.py#L750, and reverting most changes in this file

lvliang-intel: _PRE_DEFINED_FIXED_ATTR is necessary for transformers >= 5.6, otherwise it will raise "RuntimeError: The size of tensor a (512) must match the size of tensor b (256) at non-singleton dimension 3".

wenhuach21: OK, when variable_block_input is enabled, it is expected to cache all block inputs for each block except hidden_ids, for which only the first block should be cached. Please make sure this behavior has not changed. I don't know why your code work as it seems you only cache hidden_ids.

For transformers>=5.6, I suspect Heng’s patch has some issues as transformers made some change. I suggest updating the patch and only supporting this model with transformers>=5.6.

lvliang-intel: hidden_states: Only block 0's is cached (skipped for blocks 1-N for RAM saving). All other inputs (position_embeddings, position_ids, attention_mask, etc.): Cached per-block in self.inputs[block_name].

wenhuach21: Thanks, the RAM usage has been reduced to around 70GB, which I think is expected since the entire model is basically materialized. Since Gemma4 is an important model, would it be possible to refine the monkey-patch approach and disable the variable_block_inputs ? At least for transformers < 5.6, this should be doable and could save lots of ram futher.

auto_round/compressors_new/mllm_mixin.py
```diff
@@ -234,8 +236,13 @@ def calib(self, nsamples, bs):
                     mc.model(**data_new)
                 else:
                     mc.model(data_new)
+                # Clean up activation tensors between samples to reduce fragmentation.
+                gc.collect()
+                clear_memory(device_list=self.compress_context.device_list)
```
wenhuach21: clear_memory should have already conducted gc collect

auto_round/compressors_new/calib.py
lvliang-intel: The monkey-patch refined.

auto_round/utils/device.py
```diff
@@ -1918,12 +1918,59 @@ def dispatch_model_by_all_available_devices(
...
+            original = max_memory[primary_idx]
+            reserved = max(original - non_main_reserved, 0)
+            max_memory[primary_idx] = reserved
+            logger.info(
+                f"Diffusion dispatch: reserving {non_main_reserved / 1e9:.1f}GB on device {primary_idx} "
+                f"for non-target components ({original / 1e9:.1f}GB -> {reserved / 1e9:.1f}GB)"
+            )
+
+        new_max_memory = {}
+        for device in devices:
```
wenhuach21: @xin3he  please help review this change

auto_round/special_model_handler.py
```diff
@@ -210,6 +332,8 @@ def _handle_special_model(model):
 
         if version.parse(transformers.__version__) < version.parse("5.6"):
             _patch_gemma4_model(model)
+        else:
+            _attach_gemma4_rotary_emb(model)
```
wenhuach21, line 334: is it better to log a warning here as this monkey patch is a little risky, e.g.,
We apply a monkey patch to this model to reduce RAM usage,  however, the patch has only been validated with Transformers versions 5.6–5.8.

wenhuach21 (approved):

auto_round/utils/device.py
```diff
@@ -1924,38 +1924,64 @@ def dispatch_model_by_all_available_devices(
...
         no_split_modules = normalize_no_split_modules(getattr(main_model, "_no_split_modules", []))
+        # get_balanced_memory applies a 0.9 ratio internally to leave headroom.
+        max_memory = get_balanced_memory(main_model, max_memory=None, no_split_module_classes=no_split_modules)
+
+        # Reserve space on primary device for non-target components
+        primary_idx = int(primary_device.split(":")[-1]) if ":" in primary_device else 0
+        if primary_idx in max_memory and non_main_reserved > 0:
+            original = max_memory[primary_idx]
+            reserved = max(original - non_main_reserved, 0)
+            max_memory[primary_idx] = reserved
```
xin3he: Per my understanding, max_memory will limit the available memory of primary_device.
In your case, max_mamory = {'0': 37.81GB, '1': 29.42GB}, but actually the model needs {'0': 37.81GB, '1': **37.81**GB}。Then the (37.81-29.42)GB will be dispatched to CPU even the GPU is 80GB.
```
main_device_map = infer_auto_device_map(
            main_model, max_memory=new_max_memory, no_split_module_classes=no_split_modules
        )
```
If GPU memory is enough, we'd better dispatch all params to GPU. Could you please check whether CPU is used in main_device_map? You can resolve it if my understanding is wrong

lvliang-intel: Your understanding is correct. This was a bug in the previous implementation. But the main branch latest code already fixed the OOM issue. So I will reuse the main branch code here.

PR #1816, opened 2026-05-14
Fix forwarding of ExtraConfig overrides

dhruvil237:
## Description
This fixes a constructor-path bug where `ExtraConfig` scheme overrides were dropped before reaching `AutoRoundCompatible`.

When the old `AutoRound` API is used with `extra_config=ExtraConfig(...)`, `autoround.py` only merged fields that already existed in `local_args`. Fields that the new architecture expects through `kwargs`, such as `bits` and `group_size`, were silently lost. As a result, preset scheme defaults could override explicit user settings.

This change preserves the existing `local_args` behavior and also forwards non-local `ExtraConfig` values into the entrypoint kwargs, while still letting explicit kwargs win if both are provided.

A regression test was added to cover the failing path.

## Type of Change
Bug fix

## Related Issues
Fixes or relates to #

## Local Validation
- Reproduced on current `main` that `extra_config=ExtraConfig(group_size=32, bits=4)` reached `AutoRoundCompatible` without `group_size` or `bits`.
- Verified after the fix that the same path forwards `group_size=32` and `bits=4` correctly.
- `pytest` is not installed in this environment, so the repo test suite was not run here.


dhruvil237:
@wenhuach21 are there any more changes required in this PR?

wenhuach21:
> @wenhuach21 are there any more changes required in this PR?

Thanks for the pr！ I was not aware of this way before，I generally use AutoRound(scheme=“w4a16”，group_size=64).

@n1ck-guo 
Shall we keep the two ways or just one way？

n1ck-guo:
Thank you for helping us identify and resolve this issue. However, after fixing it, some other issues were exposed. We have made further modifications based on your changes, which are included in this pull request: https://github.com/intel/auto-round/pull/1817. Thanks again for your efforts in uncovering the problem, and you are very welcome to continue contributing.

PR #1862, opened 2026-05-27
Fix SDPA bug

luoyu-intel:
## Description

1. fix this issue: https://github.com/intel/sycl-tla/pull/813
2. add PV int8 support: V per 1-D block quantization (accuracy ready, performance is not ready)
## Type of Change

Bug fix & New feature


a32543254 (approved):
LGTM

Issue #1851, opened 2026-05-25
CUDA CI Failed on SM120


XuehaoSun:
1. test_fp8_format:
```
    @functools.cache
    def _load_finegrained_fp8_kernel() -> FineGrainedFP8:
        """
        Load the finegrained-fp8 Triton kernel once and return its entry points.
    
        Raises `ImportError` if the `kernels` package is missing, or the kernel or required
        symbols cannot be found.
        """
        if not is_kernels_available():
>           raise ImportError(
                "finegrained-fp8 kernel requires the `kernels` package. Install it with `pip install -U kernels`."
            )
E           ImportError: finegrained-fp8 kernel requires the `kernels` package. Install it with `pip install -U kernels`.
```
2. test_sglang:
```
[2026-05-25 05:29:23] Scheduler hit an exception: Traceback (most recent call last):
  File "/root/.venv/lib/python3.12/site-packages/sglang/srt/model_executor/cuda_graph_runner.py", line 736, in __init__
    self.capture()
  File "/root/.venv/lib/python3.12/site-packages/sglang/srt/model_executor/cuda_graph_runner.py", line 918, in capture
    _capture_one_stream()
  File "/root/.venv/lib/python3.12/site-packages/sglang/srt/model_executor/cuda_graph_runner.py", line 906, in _capture_one_stream
    ) = self.capture_one_batch_size(bs, forward, stream_idx)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.venv/lib/python3.12/site-packages/sglang/srt/model_executor/cuda_graph_runner.py", line 1187, in capture_one_batch_size
    run_once()
  File "/root/.venv/lib/python3.12/site-packages/sglang/srt/model_executor/cuda_graph_runner.py", line 1165, in run_once
    logits_output_or_pp_proxy_tensors = forward(
                                        ^^^^^^^^
  File "/root/.venv/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/root/.venv/lib/python3.12/site-packages/sglang/srt/models/qwen3.py", line 517, in forward
    hidden_states = self.model(
                    ^^^^^^^^^^^
  File "/root/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1779, in _wrapped_call_impl
[...]

PR #1909, opened 2026-06-10
fix: guard zero-division in GGUF quant kernels to avoid NaN block scales

### LINKED ISSUE #1908: [Bug] GGUF Q6_K export writes NaN block scales for all-zero blocks (0/0 in make_qx_quants_chunk) — NaN logits in llama.cpp
## Summary

The GGUF Q6_K export path writes **NaN block scales** (the fp16 `d` at the end of each 210-byte Q6_K block) for blocks whose source weights are **all zero** — e.g. padded/unused vocab rows in an embedding tensor. A single NaN scale poisons the whole forward pass: the GGUF loads fine, but llama.cpp inference returns NaN logits/PPL for any batch touching those rows.

## Environment

- auto-round **v0.13.0** (also present on current `main` — the affected code is unchanged)
- Observed with the CLI:

```bash
auto-round --model google/gemma-4-26B-A4B-it-qat-q4_0-unquantized \
  --format "gguf:q4_0" --iters 200 --nsamples 256 --seqlen 2048
```

## Observed behavior

The exported model's `token_embd.weight` (Q6_K, 2,883,584 blocks) contained **exactly 17 NaN `d` scales**. Every one was on a block whose quantized payload was uniformly zero (verified: payload min=max=0). Affected token rows included 107, 528, 532, 558, and a cluster at 236770–236832 — i.e. unused/padded vocab entries whose embedding rows are all zeros.

Patching the 17 scales to `0.0` in the GGUF fully repairs the model — it then scores normal perplexity. So the expected behavior is clear: **all-zero blocks should get `d = 0.0`** (exact reconstruction).

## Root cause

`make_qx_quants_chunk` in `auto_round/export/export_to_gguf/packing.py` (used for Q6_K scale search via `search_gguf_scale_min_sym`) divides without a zero guard:

https://github.com/intel/auto-round/blob/v0.13.0/auto_round/export/export_to_gguf/packing.py#L195-L197

```python
sumlx = (w * chunk * L).sum(dim=-1)
suml2 = (w * L * L).sum(dim=-1)
scales = sumlx / suml2          # 0/0 = NaN for an all-zero group (w = chunk**2 = 0, L = 0)
```

For an all-zero 16-element group, `w = chunk*chunk = 0` and `L = 0`, so this is `0/0 = NaN`. The NaN then slips through every downstream guard:

1. `quant_tensor_gguf_sym_dq` (`auto_round/data_type/gguf.py:818`): `torch.where(torch.abs(scale) < 1e-30, 0, scale)` — `abs(NaN) < 1e-30` is `False`, so the NaN passes through.
2. `double_quant_tensor_sym_rtn` (`auto_round/data_type/gguf.py:242-246`): `argmax` over the NaN-containing abs picks the NaN → `wmax = NaN` → `d_scale = NaN`.
3. `q6_k_quant_block` writes `d_scale` as the block's fp16 `d`.

The refinement loop inside `make_qx_quants_chunk` never repairs it either, because `replace_id = (suml2_tmp > 0) & ...` is `False` when `suml2_tmp` is NaN.

Notably, the original non-chunked `make_qx_quants` (same file, line 270) already has the correct guard — `scales = sumlx * get_reciprocal(suml2)`, where `get_reciprocal` returns 0 for (near-)zero inputs. The chunked rewrite dropped it. llama.cpp's own `make_qx_quants` likewise guards this case (`if (!amax) { scale = 0; ... }`).

## Minimal reproduction (no model needed)

```python
import torch
from auto_round.data_type.gguf import quant_tensor_gguf_sym_dq

t = torch.zeros(2, 256)        # block 0: all zeros (e.g. an unused vocab row)
t[1] = torch.randn(256)
qdq, scales, _ = quant_tensor_gguf_sym_dq(t, bits=6, scale_dtype=torch.float32)
print(scales["scale"][0].flatten()[:4])   # tensor([nan, nan, nan, nan])
print(scales["d_scale"].flatten())        # tensor([nan, ...])
print(torch.isnan(qdq[0]).any())          # True — qdq weights are NaN too
```

## Fix (verified)

One line in `make_qx_quants_chunk`, matching the existing non-chunked implementation:

```python
scales = sumlx * get_reciprocal(suml2)
```

With this change the reproduction above yields sub-scales `0.0`, `d_scale = 0.0`, and exact all-zero reconstruction. Happy to send a PR.

Two adjacent spots probably deserve the same audit:
- `make_qp_quants` (`packing.py:491`) ends with an unguarded `return sumlx / suml2, L` (imatrix super-block path).
- The NaN-transparent `torch.where(torch.abs(scale) < 1e-30, ...)` in `quant_tensor_gguf_sym_dq` could be hardened with `torch.nan_to_num` as defense in depth.

Entrpi:
Fixes #1908

## Problem

GGUF Q6_K export writes NaN fp16 `d` block scales for blocks whose source weights are all zero (e.g. padded/unused vocab rows in `token_embd.weight`). One NaN scale is enough to make llama.cpp return NaN logits/PPL for any batch touching those rows. Full analysis in #1908.

## Root cause

`make_qx_quants_chunk` divides without a zero guard:

```python
scales = sumlx / suml2   # 0/0 = NaN for an all-zero group (w = chunk**2 = 0, L = 0)
```

The NaN then slips through `torch.where(torch.abs(scale) < 1e-30, ...)` in `quant_tensor_gguf_sym_dq` (NaN compares false), propagates through `double_quant_tensor_sym_rtn` (argmax picks the NaN), and is written as the block's fp16 `d`.

The original non-chunked `make_qx_quants` already guards this with `sumlx * get_reciprocal(suml2)`; the chunked rewrite dropped the guard.

## Fix

- `make_qx_quants_chunk`: `scales = sumlx * get_reciprocal(suml2)` — matches the non-chunked implementation and llama.cpp's own `make_qx_quants` zero-amax handling. All-zero blocks now get `d = 0.0` (exact reconstruction).
- `make_qp_quants`: same guard applied to the identical unguarded `return sumlx / suml2, L`.

## Test

Added `TestGGUFZeroBlock` (no model download needed):
- `quant_tensor_gguf_sym_dq` on an all-zero 256-weight block → no NaN in sub-scales / `d_scale`, qdq reconstructs to exact zeros.
- `ggml_quant(..., "q6_k")` end-to-end → packed fp16 `d` is `0.0`, not NaN.

Both tests fail on `main` and pass with this change. Verified separately that patching the 17 NaN scales to 0.0 in an affected gemma-4-26B GGUF restores normal perplexity.

### DIFF
--- a/auto_round/export/export_to_gguf/packing.py
+++ b/auto_round/export/export_to_gguf/packing.py
@@ -194,7 +194,9 @@ def make_qx_quants_chunk(data, bits, rmse_type=0, qw=None, split_num=1, v=0):
         # Compute sumlx and suml2 using the pre-allocated L buffer
         sumlx = (w * chunk * L).sum(dim=-1)
         suml2 = (w * L * L).sum(dim=-1)
-        scales = sumlx / suml2
+        # Guard the all-zero group case (suml2 == 0): 0/0 would produce a NaN
+        # scale that ends up as a NaN fp16 `d` in the exported GGUF block.
+        scales = sumlx * get_reciprocal(suml2)
 
         if return_early:
             iscales_inv = (1 / iscales).reshape(iscales.shape[:2])
@@ -488,7 +490,7 @@ def make_qp_quants(nmax, data, quant_weights):
         if n_changed == 0:
             break
 
-    return sumlx / suml2, L
+    return sumlx * get_reciprocal(suml2), L
 
 
 @register_qtype("bf16")
--- a/test/test_cpu/export/test_gguf_format.py
+++ b/test/test_cpu/export/test_gguf_format.py
@@ -344,3 +344,33 @@ class TestGGUF:
         assert tensor_types["blk.0.attn_k.weight"] == "Q4_K"
         assert tensor_types["blk.0.ffn_down_exps.weight"] == "Q5_0"
         assert tensor_types["blk.0.ffn_up_exps.weight"] == "Q2_K"
+
+
+class TestGGUFZeroBlock:
+    """All-zero blocks (e.g. padded/unused vocab rows in an embedding tensor) must
+    quantize with scale d=0.0, not NaN. A single NaN fp16 `d` in an exported GGUF
+    makes llama.cpp return NaN logits for any batch touching those rows."""
+
+    def test_q6_k_all_zero_block_no_nan(self):
+        from auto_round.data_type.gguf import quant_tensor_gguf_sym_dq
+
+        tensor = torch.zeros(2, 256)
+        tensor[1] = torch.randn(256)
+        qdq, scales, _ = quant_tensor_gguf_sym_dq(tensor, bits=6, scale_dtype=torch.float32)
+        assert not torch.isnan(scales["scale"]).any(), "Q6_K sub-scales contain NaN for all-zero block"
+        assert not torch.isnan(scales["d_scale"]).any(), "Q6_K d_scale contains NaN for all-zero block"
+        assert not torch.isnan(qdq).any()
+        assert torch.equal(qdq[0], torch.zeros(256)), "all-zero block must reconstruct exactly to zeros"
+
+    def test_q6_k_all_zero_block_packs_zero_d(self):
+        import numpy as np
+
+        from auto_round.export.export_to_gguf.packing import ggml_quant
+
+        tensor = torch.zeros(2, 256)
+        tensor[1] = torch.randn(256)
+        packed = ggml_quant(tensor, "q6_k", device="cpu")
+        # Q6_K block layout: [ql(128), qh(64), scales(16), d(fp16)]
+        d = np.ascontiguousarray(packed.reshape(2, 210)[:, -2:]).view(np.float16)
+        assert not np.isnan(d).any(), f"packed Q6_K d scales contain NaN: {d}"
+        assert d[0] == 0.0


wenhuach21:
Nice catch! Thanks for the pr!

wenhuach21 (approved):

wenhuach21:
@Entrpi  Please be aware of this pr https://github.com/intel/auto-round/pull/1905 if you are using main branch for gguf

wenhuach21:
> @Entrpi Please be aware of this pr #1905 if you are using main branch for gguf

sorry, no impact to q40

wenhuach21:
TO merge this pr, you need to sign in DCO. just click the DCO in CI and you could see the follow message
~~~bash
DCO
There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.

To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

Here is how to fix the problem so that this code can be merged.

Rebase the branch
If you have a local git environment and meet the criteria below, one option is to rebase the branch and add your Signed-off-by lines in the new commits. Please note that if others have already begun work based upon the commits in this branch, this solution will rewrite history and may cause serious issues for collaborators ([described in the git documentation](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) under "The Perils of Rebasing").

You should only do this if:

You are the only author of the commits in this branch
You are absolutely certain nobody else is doing any work based upon this branch
There are no empty commits in the branch (for example, a DCO Remediation Commit which was added using --allow-empty)
To add your Signed-off-by line to every commit in this branch:

Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
In your local branch, run: git rebase HEAD~1 --signoff
Force push your changes to overwrite the branch: git push --force-with-lease origin fix-gguf-q6k-nan-scale
~~~

Entrpi:
Thanks, should be sorted now.

Entrpi:
The CUDA CI failure looks unrelated to this PR — could you re-trigger `Unit-Test-CUDA-AutoRound`?

Details from the GPU Test 1 log (build 65174):

- The failing test is `test_cuda/export/test_auto_round_format.py::TestAutoRound::test_fp8_block_fp8_format`, which fails inside transformers' FP8 inference path with:
  ```
  ImportError: finegrained-fp8 kernel requires the `kernels` package. Install it with `pip install -U kernels`.
  ```
  (raised from `transformers/integrations/finegrained_fp8.py` during `eval_generated_prompt`; there's also a `DeepGEMM unavailable ... falling back to Triton` warning just before it for the same reason.)
- `test_gguf_format` — the area this PR touches — **passed** on CUDA.
- The same test failed with the identical ImportError on #1907's run (build 65162), and the environment appears pod-dependent: build 65171 (#1905) passed while running concurrently with this PR's failing build 65174.

So it seems some CUDA test pods are missing the `kernels` pip package that recent transformers needs for finegrained-fp8 generation. Installing `kernels` in the CUDA test image (or adding it to `test/test_cuda/requirements.txt`) would make this deterministic.

chensuyue:
Yes, this is a known issue on some gpus, https://github.com/intel/auto-round/issues/1851.

PR #1920, opened 2026-06-13
refactor and support for multi algs fusion

n1ck-guo:
## Description
**What this PR does**
Introduces a composable QuantizationPipeline that separates pre-processing algorithms (e.g. AWQ) from the terminal block-quantizer (e.g. AutoRound/RTN), and lets users compose them declaratively via config lists.

**Key changes:**
- QuantizationPipeline (algorithms/quantization/pipeline.py): new orchestration layer — [preprocessors…] + block_quantizer. Replaces the implicit algorithm coupling in DataDrivenCompressor.
- BasePipelineMember / BaseWeightTransformer / BaseQuantizer (base.py): clean class hierarchy with unified lifecycle hooks (prepare_run, quantize_block, finalize_run).
- AWQConfig + AWQQuantizer refactored as a BaseWeightTransformer — pure weight-smoothing preprocessor, no quantization loop of its own.
- DiffusionMixin injected dynamically at pipeline construction time (is_diffusion=True) — no if is_diffusion branches in algorithm code.
- CLI (auto_round/cli/) rewritten to expose --alg_configs for composing pipelines from the command line.
-
**Usage: AWQ + AutoRound fusion**
```python
from auto_round.algorithms.quantization.awq.config import AWQConfig
from auto_round.algorithms.quantization.sign_round.config import
SignRoundConfig
ar = AutoRound(
    model_name,
    scheme="W4A16",
    [AWQConfig(), SignRoundConfig(iters=200)],
)
model, layer_config = ar.quantize()
```
Passing a list of configs activates the pipeline: AWQ smoothing runs first on each block, then AutoRound's SignSGD optimization runs on the smoothed weights. Passing a single config (old API) continues to work unchanged.
Compatibility
- Single-config API (AutoRound(model, ...)) is fully backward compatible.
- All existing CPU tests pass; pre-existing environment failures (missing auto-round-lib, device fixtures) are unrelated to this PR.

## Type of Change

New feature

## Related Issues

Fixes or relates to #


auto_round/algorithms/quantization/sign_round/quantizer.py
```diff
@@ -129,17 +124,7 @@ def _get_loss(
...
-        block: torch.nn.Module,
-        input_ids: Union[list[torch.Tensor], dict],
-        input_others: dict,
-        reference_output,
-        *,
-        loss_device: Union[str, torch.device],
-        mid_iter_mem_check: bool = False,
-        **kwargs,
-    ) -> dict:
+    def quantize_block(self, ctx) -> dict:
```
wenhuach21: please annotate ctx

auto_round/algorithms/quantization/sign_round/quantizer.py
```diff
@@ -255,9 +233,8 @@ def quantize_block(
 
             for batch_start in range(0, len(global_indices), batch_size):
                 indices = global_indices[batch_start : batch_start + batch_size]
-                current_output = self._get_current_output(reference_output, indices)
-                current_output = to_device(current_output, loss_device)
-                output_q = self._get_current_q_output(block, input_ids, input_others, indices, device, loss_device)
+                current_output = ctx.reference_batch(indices, device=loss_device)
+                output_q = ctx.forward_batch(indices, device=device, cache_device=loss_device)
```
wenhuach21: this function is a little confusing, I don't konw which module is forwarding from the API

auto_round/algorithms/quantization/sign_round/quantizer.py
```diff
@@ -255,9 +233,8 @@ def quantize_block(
 
             for batch_start in range(0, len(global_indices), batch_size):
                 indices = global_indices[batch_start : batch_start + batch_size]
-                current_output = self._get_current_output(reference_output, indices)
-                current_output = to_device(current_output, loss_device)
-                output_q = self._get_current_q_output(block, input_ids, input_others, indices, device, loss_device)
+                current_output = ctx.reference_batch(indices, device=loss_device)
```
wenhuach21: I couldn't tell what the function does from its name.

auto_round/algorithms/quantization/sign_round/quantizer.py
```diff
@@ -240,7 +218,7 @@ def quantize_block(
         # We assume the block input and output shape is same
         if self.gradient_accumulate_steps != 1 and not self.attention_mask:
             whole_indices = torch.arange(global_batch_size)
-            num_elm = self._get_current_num_elm(input_ids, whole_indices)
+            num_elm = ctx.count_active_elements(whole_indices)
```
wenhuach21: what's the meaning of active? Per the change, its number of elements. maybe just follow torch to use numel or something like that

auto_round/algorithms/quantization/base.py
```diff
@@ -40,19 +42,129 @@
...
+                    layer = layer.unwrapper({})
+                except Exception:
+                    raise
+        set_module(self.model, layer_name, layer)
+        self._immediate_pack_and_save_module(layer_name)
+
+    def _immediate_pack_and_save_module(self, module_name):
+        from auto_round.compressors.shard_writer import ShardWriter
+
+        shard_writer = ShardWriter.get_shard_writer()
```
wenhuach21, line 120: still coupled, better decouple them in the future

auto_round/compressors/base.py
```diff
@@ -107,6 +107,47 @@ class SerializedCompressorConfig:
 SERIALIZATION_KEYS = tuple(field.name for field in fields(SerializedCompressorConfig))
 
 
+def collect_user_scheme_overrides(configs: list) -> dict[str, Any]:
+    scheme_fields = {f.name for f in fields(QuantizationScheme)}
```
wenhuach21, line 111: annotate

auto_round/algorithms/quantization/sign_round/quantizer.py
n1ck-guo: I introduced ctx.io as a unified IO abstraction so block input selection, reference output caching, and mini-batch forward logic all go through the same path. Here it is forwarding the current block (ctx.block) on the sampled batch via the IO layer. I’ll add a short comment to clarify it.

wenhuach21 (commented):
It might be better to ask Xuehao to help trigger the pre-release tests for this PR. If the full test suite is too heavy, running a representative subset should be sufficient.

auto_round/algorithms/quantization/sign_round/quantizer.py
wenhuach21: I don't think this function should be part of ctx. It would be clearer to pass it explicitly, e.g.:

forward_batch(module, indices, ctx)

That makes the function's dependencies more obvious.  But it's up to you

n1ck-guo: This interface is designed for unified management of block inputs. This way, the quantizer doesn't need to design its own logic to manage inputs, devices, types, etc., but can use a unified input. The specific forward prcoess can be modified by inheriting from and overriding the `_resolve_block_forward` method of the base class. The current design may not be perfect and will be further optimized.

wenhuach21:
I would like to experiment with the LFQ paper (*Logit-aware Final-block Quantization for Boosting the Generation Quality of Low-Bit Quantized LLMs*). This method requires backpropagation through the LM-head loss, but our current API does not seem to fit this use case very well for several reasons.

From an algorithm-development perspective, I think our API should be flexible enough to support different quantization variants and research ideas. Ideally, it should expose reusable building blocks that can be easily composed to implement new algorithms, while avoiding unnecessary logic that is irrelevant to the algorithm itself.

wenhuach21 (approved):

XuehaoSun (approved):

PR #1928, opened 2026-06-17
Fix SIGSEGV in pack-weight unpack/GEMM on malformed blob

luoyu-intel:
## Summary

Add C++-level validation after BestLA blob deserialization to reject corrupted/metadata-tampered packed-weight blobs before they can cause out-of-bounds memory access (SIGSEGV).

## Background

A vulnerability was reported where a malformed packed weight blob (e.g., one byte flipped at offset 64) causes a segmentation fault in `auto_round_kernel.unpack_weight()`. The BestLA packed format embeds metadata (dimensions, block size, buffer offsets, etc.) in the blob itself. If those fields are corrupted, the deserialized pointers become invalid, leading to out-of-bounds reads.

### `cpu_wrapper.hpp` — CPU (BestLA) path
- Added `_validate_storage_weight()` that validates all deserialized metadata before `set_buffers()`:
  - `n, k > 0` and `≤ 16M`
  - `npad ≥ n, kpad ≥ k`
  - `block > 0` and `block ≤ k`
  - `n_block` consistent with `kpad / block`
  - Layout must be `NPack`
  - `misc_size + buf_size ≤ blob_count` (if available)
- Called in `BTLAGemmUnpack`, `BTLAWOQGemmForwardWorkspace`, `BTLAWOQGemmFp32Forward`.

### `xpu_wrapper.hpp` — XPU path
- `unpackq()` and `woq_gemm()` validate `blob_count ≥ get_packw_size()`.

### `ark.cpp` — Binding layer
- `unpack_weight()` and `woqgemm()` accept and forward `blob_numel`.

### `__init__.py` — Python API
- Added `_validate_packed_blob()` for parameter type validation.
- `_unpack_weight_core()` and `woqgemm()` pass `blob.numel()` / `B.numel()` to native calls.
- `_repack_quantized_weight_core()` validates input tensor shapes.

### DIFF
--- a/auto_round_extension/ark/auto_round_kernel/__init__.py
+++ b/auto_round_extension/ark/auto_round_kernel/__init__.py
@@ -343,19 +343,13 @@ def _validate_packed_blob(
         raise TypeError(f"blob must be a torch.Tensor, got {type(blob).__name__}")
 
     if blob.device.type not in ("cpu", "xpu"):
-        raise ValueError(
-            f"blob must reside on cpu or xpu, got {blob.device.type}"
-        )
+        raise ValueError(f"blob must reside on cpu or xpu, got {blob.device.type}")
 
     if blob.dtype != torch.int8:
-        raise ValueError(
-            f"blob must have dtype torch.int8, got {blob.dtype}"
-        )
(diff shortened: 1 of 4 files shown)


a32543254 (approved):
LGTM

PR #1933, opened 2026-06-17
fix saving hadamard rotation model without rotation config.

lkk12014402:
## Description

Fix regression that caused the rotation config to be lost when saving Hadamard transform models.


wenhuach21:
please refine the doc

auto_round/algorithms/transforms/quarot/apply.py
```diff
@@ -102,7 +102,7 @@ def apply_to_model(
             **kwargs:        Reserved for future use.
 
         Returns:
-            The mutated *model* with ``model.rotation_config`` set to the
+            The mutated *model* with ``model._rotation_config`` set to the
             normalised :class:`RotationConfig` dict.
```
wenhuach21, line 113: is moe supported? if not, please have a check and fallback in the code

lkk12014402: The Hadamard rotation feature performs rotation on all linear layers and is compatible with MoE architectures

lkk12014402:
> please refine the doc

mark the rotation feature as experimental

docs/step_by_step.md
```diff
@@ -859,7 +823,9 @@ autoround.save_quantized(format="auto_awq", output_dir="tmp_autoround")
   Include the flag `--adam`. Note that AdamW is less effective than sign gradient descent in many scenarios we tested.
 
 
-### Rotation
+### Rotation (Experimental)
+
+> ⚠️ **Experimental feature**: Rotation transform is still in experimental stage. Inference uses forward hooks, which may result in slower inference speed compared to native (non-rotated) models.
```
wenhuach21: add "only supported in transformers". I still prefer to mark it as a research feature as vllm/sglang does not support it.

lkk12014402: support vllm deployment now with vllm plugin for hadamard rotation and quarot rotation

wenhuach21: How should users use the vLLM plugin? Is it intended for production deployment with good performance, or is it still a research feature? Please think from a user’s perspective.

From a user’s point of view: if it can be deployed in production in most scenarios, it should be considered a production-ready feature. If it is only suitable for production in limited scenarios, it should be marked as an experimental feature. Otherwise, it should be treated as a research feature.

lkk12014402: The inference using the vLLM plugin depends on https://github.com/yiliu30/vllm-qdq-plugin. We plan to integrate this feature into AutoRound in the next release. Currently, however, the vLLM plugin is maintained in a separate repository.

wenhuach21: I assume this is still a research feature. While it may run successfully, I wouldn't expect near-production-level performance, similar to the current Transformers integration. In addition, users cannot achieve this with an out-of-the-box vLLM deployment, which further suggests that it is not yet production-ready.

I would also suggest keeping this plugin outside of AutoRound, as it is not really within AutoRound's scope.

lkk12014402: ok

lkk12014402: updated the description

xin3he (approved):

PR #1953, opened 2026-06-25
fix auto_scheme gguf format wrong bits if use tiny model

n1ck-guo:
## Description

## Type of Change

Bug fix

## Related Issues

Fixes or relates to #

### DIFF
--- a/auto_round/auto_scheme/delta_loss.py
+++ b/auto_round/auto_scheme/delta_loss.py
@@ -1201,7 +1201,7 @@ def _apply_head_trick(head_name, schemes, sorted_indices, target_bits, target_pa
     # assign it >= 6 bit schemes, then let DP globally optimize.          #
     #                                                                      #
     # Rules (only if user hasn't already fixed it):                        #
-    #   1. No option has bits >= 6      → keep all options (DP picks best)#
+    #   1. No option has bits >= 6      → prefer the lowest-loss option   #
     #   2. Exactly one option bits >= 6 → restrict to only that option    #
     #   3. Multiple options bits >= 6:                                      #
     #      - target_bits > 6  → restrict to only the highest-bit option   #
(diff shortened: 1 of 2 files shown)


auto_round/auto_scheme/delta_loss.py
```diff
@@ -1530,9 +1543,11 @@ def check_bf16_scheme(scheme):
     target_params_cnt = int(total_params * target_bits)
     sorted_indices = sorted(range(len(options_scores)), key=lambda i: options_scores[i])
     # Layers that are not fixed in fixed_layer_scheme
-    not_fixed_embedding_layers_names = [
-        name for name in embedding_layers_names if (name not in fixed_layer_scheme and name in quant_layer_names)
-    ]
+    not_fixed_embedding_layers_names = _get_not_fixed_embedding_layer_names(
```
wenhuach21: Thanks for the fix. I still see a potential issue.

For non-GGUF formats, the embedding layer should not be included in any bits calculation, since bit calculation is only meant for layers that will be quantized.

However, I noticed code like this:
~~~python
total_params = 0
for n, m in model.named_modules():
    if n in quant_layer_names + embedding_layers_names:
        n_param = m.weight.numel()
        if n_param == 0 and hasattr(m, "_cached_weight_numel"):
            n_param = m._cached_weight_numel
        total_params += n_param
~~~
Could you please ensure that embedding layers are excluded from bit calculation for all non-GGUF formats?

wenhuach21: The lm-head is similar. Its bits are only accounted for when the user explicitly chooses to quantize it.

wenhuach21: # Filter out embedding layers from the scoring set (they aren't linear
    # quantization targets in any of our schemes).
    embedding_layers_names = []
    for name in quant_layer_names:
        module = get_module(model, name)
        if isinstance(module, torch.nn.Embedding):
            embedding_layers_names.append(name)
    quant_layer_names = list(set(quant_layer_names) - set(embedding_layers_names))
    
   
no issues, as embedding_layers_names is a subset of quant_layer_names

n1ck-guo: done

auto_round/auto_scheme/delta_loss.py
```diff
@@ -1277,6 +1278,40 @@ def _apply_head_trick(head_name, schemes, sorted_indices, target_bits, target_pa
...
+            return True
+    return False
+
+
+def _get_bit_budget_layer_names(quant_layer_names, embedding_layers_names, schemes):
+    """Return layer names whose params contribute to the AutoScheme bit budget."""
+
+    layer_names = list(quant_layer_names)
+    if _is_gguf_auto_scheme(schemes):
+        layer_names += list(embedding_layers_names)
```
wenhuach21: as mentioned in the comments, there is no need to add this as embedding layer is a subset of quant_layer_names, so for no-gguf formats, embedding layers have been excluded

xin3he (approved):

XuehaoSun (approved):

PR #1956, opened 2026-06-26
rename the base rotation (deterministic/random hadamard) directory name.

lkk12014402:
## Description

| Directory | What the name implies | What the directory actually contains |
|---|---|---|
| `quarot/` | the QuaRot algorithm | ❌ **not** QuaRot — it is **per-Linear Hadamard** (registered as `"hadamard"`) |
| `spinquant/` | only SpinQuant | both SpinQuant **and** QuaRot (QuaRot = the non-trainable variant of SpinQuant) |

### After the rename

| Directory | Contents | Registered name |
|---|---|---|
| `hadamard/` | per-Linear Hadamard rotation | `"hadamard/random_hadamard"` |
| `spinquant/` | shared R1–R4 implementation for SpinQuant (trainable) + QuaRot (non-trainable) | `"spinquant"` / `"quarot"` |

The directory name `hadamard/` now matches its registered name `"hadamard"` and
is no longer misleading.


lkk12014402:
@n1ck-guo  please help review~

n1ck-guo:
LGTM  @wenhuach21 any comment?

wenhuach21:
looks good to me. 
1 From an algorithm perspective, however, Quarot is more accurate regardless of whether rotation is fused or not.

2 The rotation documentation should also be refined. It contains too much low-level detail that most users won't understand. It would be better to keep the main documentation concise and move the implementation details to a separate section (or collapse them under a "Details" section).

lkk12014402:
> looks good to me. 1 From an algorithm perspective, however, Quarot is more accurate regardless of whether rotation is fused or not.
> 
> 2 The rotation documentation should also be refined. It contains too much low-level detail that most users won't understand. It would be better to keep the main documentation concise and move the implementation details to a separate section (or collapse them under a "Details" section).

okay

xin3he (approved):

PR #1592, opened 2026-03-22
feat: add --dry-run estimation mode

mvanhorn:
## Summary

Adds a `--dry-run` flag to the CLI that estimates VRAM usage, output file size, and approximate quantization time without running the full quantization process.

- Loads only the model config via `AutoConfig.from_pretrained()` (no weights downloaded)
- Estimates peak VRAM from parameter count, dtype, batch size, and sequence length
- Estimates output file size from target bit width, parameter count, and group size overhead
- Estimates time from layer count, iterations, and calibration batch count
- Prints a formatted summary table and exits

## Motivation

Users quantizing large models (70B+) need to know resource requirements before committing GPU hours. This is relevant to #1551 (reduce quant cost) and #1584 (peak VRAM tracking).

## Example output

```
============================================================
  AutoRound Dry-Run Estimation
============================================================
  Model:              meta-llama/Llama-2-7b-hf
  Parameters:         6.61B
  Layers:             32
  Target bits:        4
  Group size:         128
  Model dtype:        float16
============================================================
  Estimated peak VRAM:    17.80 GB
  Estimated output size:  3.64 GB
  Estimated time:         3.4 hours
    (batch_size=8, seqlen=2048, nsamples=128, iters=200)
============================================================
  NOTE: These are rough estimates. Actual values depend on
  hardware, model architecture, and runtime conditions.
============================================================
```

## Changes

- **New:** `auto_round/estimation.py` - VRAM, disk, and time estimation functions
- **Modified:** `auto_round/__main__.py` - `--dry_run` / `--dry-run` CLI flag, short-circuits before model loading
- **New:** `test/test_cpu/core/test_estimation.py` - unit tests for all estimation functions

## Testing

All estimation unit tests pass (parameter counting, VRAM estimation, output size calculation, time estimation, format helpers). Tests use stub configs to avoid model downloads.

Fixes #1591

This contribution was developed with AI assistance (Claude Code).

PR #1958, opened 2026-06-26
feat: add --dry-run VRAM/size estimation mode

### LINKED ISSUE #1591: [Feature]: add --dry-run estimation mode
### Feature Description

Add a `--dry-run` flag to the CLI that estimates VRAM usage, output file size, and approximate quantization time without running the full quantization process. It loads the model config via `AutoConfig.from_pretrained()` (no weights) and prints a summary table, then exits.

### Motivation and Use Case

Users quantizing large models (70B+) need to know resource requirements before committing GPU hours. Currently the only way to find out is to start the process and watch it fail or succeed. A lightweight estimation step would save significant time and compute costs.

This is related to #1551 (reducing quant cost) and #1584 (VRAM tracking) - both involve understanding resource requirements for quantization.

### Alternatives Considered

- Manually calculating estimates from model card metadata - error-prone and tedious
- Running the full process with a tiny calibration set - still loads model weights and takes significant time/memory

### Definition of Done

- [x] `--dry-run` CLI flag prints VRAM, output size, and time estimates
- [x] Python API supports `dry_run=True` parameter in `AutoRound`
- [x] Works for common quantization schemes (W4A16, W2A16, W3A16, W8A16, MXFP4, etc.)
- [x] Unit tests for estimation functions
- [x] Exits cleanly without loading model weights

### Additional Context

The estimation approach:
- Peak VRAM: calculated from parameter count, dtype size, calibration batch size, and sequence length
- Output size: calculated from target bit width and parameter count
- Time: rough estimate from layer count, iterations, and calibration sample count

Estimation uses `AutoConfig.from_pretrained()` to load only the model architecture metadata.

### LINKED ISSUE #1689: [Bug]: dry-run estimation should fall back to text_config for Qwen3.5-MoE
## Problem

The dry-run estimation path reads model-shape fields only from the top-level config object.

Qwen3.5-MoE keeps the relevant fields under `text_config`, so the estimator cannot count parameters correctly for that model family.

## Affected code

- `auto_round/estimation.py:48-56`
- `auto_round/estimation.py:216-217`

## Repro

This reproduces with the Qwen3.5 MoE config shape:

```bash
python - <<'PY'
from transformers import AutoConfig

cfg = AutoConfig.from_pretrained("Qwen/Qwen3.5-35B-A3B", trust_remote_code=True)
print("top-level hidden_size:", getattr(cfg, "hidden_size", None))
print("nested hidden_size:", getattr(cfg.text_config, "hidden_size", None))
PY
```

Expected result:
- the estimator should fall back to `cfg.text_config` for fields such as `hidden_size`, `num_hidden_layers`, `intermediate_size`, and `vocab_size`

Actual result:
- those top-level fields are `None` for Qwen3.5-MoE-style configs, so parameter counting and the estimate result break or undercount

If you run the dry-run path against `Qwen/Qwen3.5-35B-A3B`, it cannot size the model correctly for this reason.

## Why this matters

Dry-run estimation is most useful on large MoE models where memory headroom is tight.

Qwen3.5-MoE is one of those models.

## Suggested fix

Add a small helper that reads each required field from the top-level config first and then falls back to `config.text_config` when present.

Use that helper in both `_count_parameters()` and `_build_estimate_result()`.

mvanhorn:
Re-submit of #1592, reworked per your feedback. The VRAM estimate is now built on auto-round's own block-wise memory model instead of a raw parameter count:

- Reuses `auto_round.utils.device.estimate_tuning_block_mem` and `get_moe_memory_ratio` (@xin3he's `card_0 = block_input_output + layer_activation + additional` formula).
- Models peak memory at decoder-block granularity (@wenhuach21).
- Excludes the block input/output cache when `low_gpu_mem_usage` is set, and includes it otherwise (@wenhuach21's caching point).
- MoE handled via `get_moe_memory_ratio`; layer/hidden-size discovery is robust across config field names and nested configs (`text_config` etc.) since `num_hidden_layers` doesn't cover every model (@xin3he).
- Loads `AutoConfig` only, no weights.

Unit tests cover helper reuse, the `low_gpu_mem_usage` path, MoE, and layer-count fallbacks (8 passing). Supersedes #1592 (couldn't reopen after the rebase).

### DIFF
--- a/auto_round/cli/main.py
+++ b/auto_round/cli/main.py
@@ -276,23 +276,9 @@ def tune(args):
     if "marlin" in args.format and args.asym is True:
         raise RuntimeError("marlin backend only supports sym quantization, please remove --asym")
 
-    from auto_round.utils import get_device_and_parallelism
-
-    device_str, use_auto_mapping = get_device_and_parallelism(args.device_map)
-
-    if args.enable_torch_compile:
-        logger.info(
-            "`torch.compile` is enabled to reduce tuning costs. "
-            "If it causes issues, you can disable it by removing `--enable_torch_compile` argument."
-        )
-
     model_name = args.model
     if model_name[-1] == "/":
         model_name = model_name[:-1]
-    logger.info(f"start to quantize {model_name}")
-
-    from auto_round.compressors.base import BaseCompressor
-    from auto_round.compressors.entry import AutoRound as PipelineAutoRound
 
     if "bloom" in model_name:
         args.low_gpu_mem_usage = False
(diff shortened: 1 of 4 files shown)


mvanhorn:
Pushed a fix for the two red checks:

- Code-Scan (pylint E0401): the optional `modelscope` import in `_load_auto_config` now carries `# pylint: disable=E0401`, matching how `calib_dataset.py` and `utils/model.py` handle modelscope (it's not in requirements). Scan is now 10.00/10 locally.
- DCO: signed off all commits on the branch.

wenhuach21:
Hi, thanks for the PR! It would be very useful if the estimation is both general and accurate.

Could you kindly share some evaluation results demonstrating its accuracy? It would be helpful to include results covering an LLM, a VLM, and an MoE model.

mvanhorn:
Happy to. Since --dry-run estimates VRAM/size analytically from the model config rather than measuring on hardware, I can put together a table comparing the estimate against the theoretical footprint (param count x bit-width, plus the group/zp/scale overhead the estimator accounts for) for a representative LLM, VLM, and MoE. That validates the estimation math end to end without needing a specific GPU.

Where I'd lean on you: confirming the estimate against actual measured peak VRAM on real hardware for those three model classes - you're better set up for that than I am. If a config-vs-theoretical table plus your measured spot-check works, I'll post the table. Which three models would you want represented (so the numbers match something you can cross-check)?

wenhuach21:
it's not easy to estimate the ram accurately. especially with torch compile. You could use rtn mode to quickly generate a model to know the size, for memory, you could leverage iters=1

PR #1979, opened 2026-07-01
fix: promote None-initialized kwarg to list in forward_capture

### LINKED ISSUE #1950: [Bug]: forward_capture AttributeError ('NoneType' has no attribute 'extend') during MoE calibration when a kwarg is None-then-Tensor
### Problem Description

When AutoRound calibrates a Mixture-of-Experts model (here GLM-5.2, arch `GlmMoeDsaForCausalLM`, 256 routed experts) driven via **llm-compressor**'s `AutoRoundModifier`, calibration crashes during the input-capture forward with:

```
File ".../auto_round/utils/model.py", line 2199, in forward
    return base_hook(m, hidden_states, *positional_inputs, **kwargs)
File ".../auto_round/calibration/hooks.py", line 126, in forward_capture
    state.inputs[name][key].extend(
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'extend'
```

### Root cause

In `auto_round/calibration/hooks.py` `forward_capture`, the **initialization** branch stores a raw value for a key that is `None` *or* a shared-cache key (paraphrasing 0.13.0):

```python
if key not in state.inputs[name].keys():            # initialization
    data = to_device(kwargs[key], device=torch.device("cpu"))
    if data is None or key in state.model_context.shared_cache_keys:
        state.inputs[name][key] = data              # <-- stores raw None for a NON-shared key
        continue
    ...
```

Later, on a subsequent sample where the **same key arrives as a Tensor**, the append branch runs and does:

```python
else:  # append cache inputs
    ...
    if state.quantizer.batch_size <= 1:
        state.inputs[name][key].append(new_data)
    else:
        state.inputs[name][key].extend(            # line 126 -> AttributeError: None has no .extend
            list(torch.split(new_data, 1, dim=state.quantizer.batch_dim)))
```

So the failure occurs whenever a forward kwarg is **`None` on the first calibration sample but a `Tensor` on a later sample** for a key that is *not* in `shared_cache_keys`. The init path stored a scalar `None` (not a per-sample list), and the append/extend path assumes a list. This is common for MoE / sparse-attention models where optional kwargs (e.g. a mask or routing-related tensor) are `None` for some sequences and present for others.

### Suggested fix

Treat a non-shared `None` key as per-sample list storage at init, OR guard the append branch. Minimal guard at the append site:

```python
else:  # append cache inputs
    ...
    if not isinstance(state.inputs[name].get(key), list):
        # key was first seen as None (non-shared); promote to per-sample list
        prev = state.inputs[name].get(key)
        state.inputs[name][key] = [] if prev is None else [prev]
    if state.quantizer.batch_size <= 1:
        state.inputs[name][key].append(new_data)
    else:
        state.inputs[name][key].extend(list(torch.split(new_data, 1, dim=state.quantizer.batch_dim)))
```

(Or, at init, store `[data]` instead of raw `data` when `data is None and key not in shared_cache_keys`.)

### Environment

- auto-round 0.13.0
- llm-compressor 0.12.0 (`AutoRoundModifier(targets=["Linear"], scheme="W4A16", iters=200)`, `moe_calibrate_all_experts=True`)
- transformers 5.10.1, torch 2.12.0, CUDA 12.8
- Model: GLM-5.2 (`GlmMoeDsaForCausalLM`, 256 routed experts), bf16 source

### Notes

Happy to open a PR with the fix if the maintainers prefer one of the two approaches above. The crash reproduces deterministically at the first MoE layer's calibration; it is independent of bit-width (seen at W4A16).

ramkrishs:
## Description

`forward_capture` in `auto_round/calibration/hooks.py` raised an `AttributeError`
during calibration when an optional kwarg (e.g. a sparse-attention mask in MoE
models) was `None` on the first calibration sample but a real `Tensor` on a later
one.

**Root cause:** The initialization branch stores `None` directly when a kwarg's
value is `None`:

```python
if data is None or key in state.model_context.shared_cache_keys:
    state.inputs[name][key] = data   # stores None
```

On the next sample, the append branch assumed the stored value was always a list
and called `.append()` or `.extend()` on it, raising:

```
AttributeError: 'NoneType' object has no attribute 'extend'
```

**Fix:** One guard before the append/extend calls promotes a stored `None` to an
empty list, covering both `batch_size <= 1` (`.append`) and `batch_size > 1`
(`.extend`) paths.

## Type of Change

- [x] Bug fix

## Related Issues

Fixes #1950

## Testing

Added three regression tests in `test/test_cpu/core/test_forward_capture_none_kwarg.py`:

| Test | What it covers |
|------|---------------|
| `test_none_then_tensor_kwarg_batch_size_1` | None → Tensor transition, `batch_size=1` (`.append` path) |
| `test_none_then_tensor_kwarg_batch_size_gt1` | None → Tensor transition, `batch_size>1` (`.extend` path) |
| `test_normal_tensor_kwarg_unaffected` | Kwargs that are always Tensors still accumulate correctly |

All three pass locally against the fix.

### DIFF
--- a/auto_round/calibration/hooks.py
+++ b/auto_round/calibration/hooks.py
@@ -119,6 +119,9 @@ def make_block_forward_func(state, name: str) -> Callable:
                                 state.inputs[name][key] = [stored, raw_new]
                         continue
                     new_data = to_device(new_data, device=torch.device("cpu"))
+                    # Guard against None-initialized kwargs that arrive as tensors on later samples (#1950).
+                    if state.inputs[name][key] is None:
+                        state.inputs[name][key] = []
                     if state.quantizer.batch_size <= 1:
                         state.inputs[name][key].append(new_data)
                     else:
--- a/test/test_cpu/core/test_forward_capture_none_kwarg.py
+++ b/test/test_cpu/core/test_forward_capture_none_kwarg.py
@@ -1,0 +1,137 @@
+"""Regression tests for https://github.com/intel/auto-round/issues/1950.
+
+``forward_capture`` raised ``AttributeError: 'NoneType' object has no attribute
+'extend'`` when an optional kwarg was ``None`` on the first calibration sample
+but a real ``Tensor`` on a subsequent one.  The fix promotes the stored ``None``
+to an empty list before appending.
+"""
+
+from functools import partial
+from types import SimpleNamespace
+
+import torch
+
+from auto_round.calibration.hooks import make_block_forward_func
+
+
+def _make_state(batch_size=1):
+    """Return a minimal state stub accepted by ``make_block_forward_func``.
+
+    Args:
+        batch_size (int): Number of samples per calibration batch.
+
+    Returns:
+        SimpleNamespace: State object with ``inputs``, ``quantizer``,
+            ``model_context``, ``has_variable_block_shape``,
+            ``blocks_requiring_input_ids``, and ``_should_stop_cache_forward``.
+    """
+    quantizer = SimpleNamespace(batch_size=batch_size, batch_dim=None)
+    model_context = SimpleNamespace(
+        shared_cache_keys=("position_ids", "cache_position", "position_embeddings", "cu_seqlens"),
+    )
+    return SimpleNamespace(
+        inputs={},
+        quantizer=quantizer,
+        model_context=model_context,
+        has_variable_block_shape=False,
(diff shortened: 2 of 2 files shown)


n1ck-guo (approved):
LGTM. Thanks for your effor.

PR #1982, opened 2026-07-02
fix: avoid silent weight loss when OffloadManager fails to save a block

xin3he:
We should exit instead of warning, otherwise the error log will mislead users.
image


xin3he:
s

lvliang-intel (approved):

PR #2035, opened 2026-07-09
Fix invalid group reference in revert_checkpoint_conversion_mapping

### LINKED ISSUE #2028: [Bug] revert_checkpoint_conversion_mapping crashes with re.error "invalid group reference" on transformers-5.x models
### Problem Description

Exporting a quantized model whose HF class defines a `checkpoint_conversion_mapping` (transformers 5.x) via `format="llm_compressor"` raises `re.error: invalid group reference 1` inside `revert_checkpoint_conversion_mapping`.

The function strips regex **capture groups** from the *source* pattern (`re.sub(r"\(.*\)", "", source_pattern)`) but leaves the corresponding **backreferences** (e.g. `\1`) in the *target* pattern. `re.subn(source, target, name)` then fails to compile the replacement template because the referenced group no longer exists.

### Reproduction Steps

```python
ar = AutoRound(model="<any Qwen3.5-MoE checkpoint>", scheme="W4A16")
ar.quantize_and_save(output_dir="out", format="llm_compressor")
```

Qwen3.5/3.6-MoE's transformers conversion mapping is e.g.
`{'^model\.language_model\.(.+)$': 'model.\\1', ...}` — the `(.+)` group is stripped from the source while `\1` remains in the target, triggering the error.

### Environment Information

- auto-round 0.14.1
- transformers 5.10.4
- torch 2.12.0+xpu
- Python 3.12.3

### Error Logs

```
File ".../auto_round/utils/common.py", line 1195, in revert_checkpoint_conversion_mapping
    name, n_replace = re.subn(source_pattern, target_pattern, name)
File ".../re/__init__.py", line 334, in _compile_template
    return _sre.template(pattern, _parser.parse_template(repl, pattern))
re.error: invalid group reference 1 at position 22
```

### Additional Context

- The function is imported into multiple modules (`compressors/base.py`, `compressors/shard_writer.py`). The `shard_writer` path only fires when a `layer_config` excludes some layers (mixed-precision quant), so a **uniform** quant can pass while a **mixed** one fails — worth patching in both call sites' shared helper.
- Suggested fix: keep source/target consistent — either do **not** strip the capture group from the source, or strip the backreference from the target whenever the group is stripped from the source.
- Worked around locally by monkeypatching the function to try the intact pattern first and fall back to stripping both sides.

A-S-Manoj:
## Description
`revert_checkpoint_conversion_mapping` unconditionally stripped the capture
group out of each source regex pattern, even when the corresponding target
pattern still referenced that group via a backreference (e.g. `\1`).

This is fine for mappings where the target is a plain string, but
transformers-5.x `checkpoint_conversion_mapping`s (e.g. Qwen3.5/3.6-MoE)
include entries like:

    {r"^model\.language_model\.(.+)$": r"model.\1"}

Stripping the group here leaves `\1` in the target dangling, and
`re.subn(source_pattern, target_pattern, name)` raises:

    re.error: invalid group reference 1 at position 7

The fix only strips the capture group when the target pattern does not
backreference it, so patterns that rely on the captured suffix keep working
while the existing bare-tensor-suffix handling is unaffected.

## Type of Change
Bug fix

## Related Issues
Fixes #2028


A-S-Manoj:
Thanks for triggering CI! `Unit-Test-CUDA-AutoRound (GPU Test 1 GpuTests)` failed, but the log shows it's an infra issue, not the code change:

    error: Request failed after 3 retries in 8.5s
      Caused by: Failed to fetch: `https://download.pytorch.org/whl/cu130/nvidia-nccl-cu13/`
      Caused by: HTTP status server error (503 Service Unavailable) for url (...)

This happened during dependency installation ("set up UT env..."), before any tests ran — `revert_checkpoint_conversion_mapping`/`common.py` never executed. GPU Test 0 and GPU Test 2 both passed in the same run.

Could you re-run `/azp run Unit-Test-CUDA-AutoRound` when you get a chance? Also happy to address any review feedback on the fix itself.

A-S-Manoj:
XPU failures are also infra, not this PR: torch==2.13.0+cu130 on the runner (no XPU support), so every test hits AssertionError: Torch not compiled with XPU enabled before the actual code runs. Everything else required (Code-Scan, Compatibility, Unit-Test-AutoRound, pre-commit) is green.
@yiliu30 mind re-running Unit-Test-XPU-AutoRound?

yiliu30:
Hi @A-S-Manoj, thanks for the contribution! I will retrigger the CI after https://github.com/intel/auto-round/pull/2043 lands.

A-S-Manoj:
Hi @yiliu30, looks like #2043 has been merged. Could you please retrigger the CI for this PR when you get a chance? Thanks!

yiliu30 (approved):
LGTM. thanks for the contribution!

A-S-Manoj:
Thank You! Glad to contribute.

PR #2061, opened 2026-07-19
Stream large checkpoints from disk during quantization

aquilarubra:
## Summary

Adds an opt-in `AR_DISK_STREAM_MODEL` env var that lets AutoRound quantize a checkpoint far larger than available RAM+VRAM, by never fully materializing it in CPU memory.

Today, `ModelContext._load_model()` unconditionally calls `llm_load_model(..., device="cpu")` (or `mllm_load_model` for multimodal checkpoints) whenever `model` is a string, fully loading the entire checkpoint onto CPU RAM before any block-wise tuning logic ever runs. Even the "regular" per-block offload/reload machinery (`OffloadManager`) only actually engaged for fused-MoE-patched architectures — a plain dense model got `low_cpu_mem_usage` silently forced back to `False`, since upstream's own assumption is "the whole model is already CPU-resident anyway, so per-block offload/reload buys nothing." That assumption breaks once the initial load itself is no longer full-residency.

When `AR_DISK_STREAM_MODEL=1`, `ModelContext` instead builds an all-meta skeleton (`accelerate.init_empty_weights()`, ~0 RAM) and materializes/frees one decoder block's real tensors — read directly from the checkpoint's safetensors shards — right before/after each place that block is actually used. Default (unset) behavior is byte-for-byte identical to upstream.

## What's in this PR

- `auto_round/envs.py`: new `AR_DISK_STREAM_MODEL` / `AR_RESUME_DIR` flags (the latter used by a follow-up resumability PR; unused by this PR alone).
- `auto_round/utils/disk_stream_util.py` (new): the streaming primitive — lazy, mmap-backed per-tensor reads from safetensors shards (never caches file handles, so touched pages don't stay RSS-resident), plus meta↔real materialize/free for a whole module.
- `auto_round/context/model.py`: builds the meta skeleton (text and multimodal paths) and materializes non-block params (embeddings/lm_head/final norm) right after the existing `unsupported_meta_device()` guard passes; re-ties tied output embeddings via `tie_weights()`.
- `auto_round/utils/offload.py`: makes the existing per-block offload/reload cycle actually work starting from a meta model (first-time materialization target-device fix, meta-aware save/reload, fused-3D-expert-tensor key splitting for on-disk layouts that don't match an already-unfused module tree).
- `auto_round/compressors/base.py`: propagates the checkpoint path to `OffloadManager` so it can materialize never-yet-offloaded blocks directly from disk.
- `auto_round/compressors/data_driven.py`: keeps `low_cpu_mem_usage` enabled for streamed dense models, reloads streamed blocks for GGUF export, and skips a `model.to("cpu")` call that can't work on an intentionally-still-meta model.

## Validation

Verified on synthetic Llama fixtures (8-layer/2.8GB, 20-layer/27.7GB): peak RAM stays flat regardless of checkpoint size (vs. proportional-to-size on the baseline), per-block tuning losses and final quantized weights are bit-identical to the unstreamed baseline. Also verified end-to-end at real scale against a 207GB checkpoint and against synthetic MoE/hybrid-attention/MTP-head fixtures — happy to share more detail if useful for review.

This is the first of a small stack of PRs splitting up a larger local patch set (disk streaming → resumability → AutoScheme streaming → calibration fixes); happy to adjust scope/splitting based on maintainer preference.

### DIFF
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -393,6 +393,12 @@ class BaseOrchestrator(object):
         # second AutoRound(...) call reuses the previous instance and silently keeps
         # stale values (e.g. low_cpu_mem_usage=True from a prior run).
         CompressContext.reset_context()
+        # When the model was built as a meta skeleton (AR_DISK_STREAM_MODEL=1),
+        # give the offloader the original checkpoint path so it can materialize
+        # each block on first touch directly from disk instead of assuming
+        # blocks already hold real weights (see OffloadManager._reload).
+        if self.model_context.disk_stream_model_dir is not None:
+            self._offloader.model_dir = self.model_context.disk_stream_model_dir
         # Alternatively, you can use CompressContext.create_context
         self.compress_context = CompressContext(
             low_cpu_mem_usage,
@@ -403,6 +409,11 @@ class BaseOrchestrator(object):
             static_attention_dtype=self.static_attention_dtype,
         )
         self.shard_writer = None
+        # Resumability state deferred from Orchestrator._quantize_data_driven() until
+        # quantize_and_save()'s save_quantized() call actually succeeds; see the
+        # comment in quantize() near "is_immediate_saving" for why clearing is
+        # deferred.
+        self._resume_states = None
 
         # Flag for post_init idempotency.  Set to False here so post_init() can be called
         # either via quantize_and_save() (preferred, outside inference_mode) or directly
@@ -1624,6 +1635,14 @@ class BaseOrchestrator(object):
         # Save the quantized model in the specified format_list
         model, folders = self.save_quantized(output_dir, inplace=inplace, return_folders=True, **kwargs)
         memory_monitor.log_summary()
+
+        # Only now -- after the full export (packing pass, config/tokenizer
+        # writes) has actually succeeded -- is it safe to drop the resume
+        # manifest. See the deferral comment in Orchestrator._quantize_data_driven().
+        if self._resume_states:
+            for rs in self._resume_states:
+                rs.clear()
+            self._resume_states = None
 
         return model, folders
 
--- a/auto_round/compressors/orchestrator.py
+++ b/auto_round/compressors/orchestrator.py
@@ -22,6 +22,7 @@ import torch
 from accelerate.big_modeling import dispatch_model
 from tqdm import tqdm
 
+from auto_round import envs
 from auto_round.calibration import CalibrationContext
 from auto_round.calibration.utils import (
     _update_inputs,
@@ -217,7 +218,19 @@ class CompressionOrchestrator(BaseOrchestrator):
                 modules = [get_module(model, n) for n in names]
                 m = WrapperMultiblock(modules)
 
-            if self.compress_context.low_cpu_mem_usage:
+            # Also reload when `AR_DISK_STREAM_MODEL` is set even if
+            # `low_cpu_mem_usage` has been forced False (e.g. GGUF export --
+            # see base.py's `_finalize_compress_context`, which disables
+            # `low_cpu_mem_usage` for gguf formats for reasons unrelated to disk
+            # streaming). Under streaming, a block starts on the meta device
+            # regardless of `low_cpu_mem_usage`, which only ever controlled whether
+            # to *free* it again after use -- without this, the block below is never
+            # materialized at all and `m.to(device)` crashes with "Cannot copy out
+            # of meta tensor". The block intentionally stays real afterward (no
+            # matching post-tune offload runs when `low_cpu_mem_usage` is False --
+            # see the `is_immediate_saving`-adjacent offload call further down),
+            # matching upstream's own choice not to cycle blocks for these formats.
+            if self.compress_context.low_cpu_mem_usage or envs.AR_DISK_STREAM_MODEL:
                 if nblocks == 1:
                     self._offloader.reload(model, n)
                 else:
@@ -550,6 +563,17 @@ class CompressionOrchestrator(BaseOrchestrator):
                 )
                 if not self._offloader.enabled:
                     self.compress_context.low_cpu_mem_usage = False
+            elif self.model_context._disk_stream_index is not None:
+                # Dense (non-MoE-patched) models normally get low_cpu_mem_usage
+                # disabled here because the per-block offload/reload dance is
+                # pointless when the whole model is already CPU-resident from
+                # the initial full load -- there's no memory to save. That
+                # assumption doesn't hold when the model started as a meta
+                # skeleton (AR_DISK_STREAM_MODEL=1): blocks are still on meta
+                # and must go through the same reload()-before/offload()-after
+                # cycle to get materialized from disk one at a time and freed
+                # again, so keep it enabled here.
+                pass
             else:
                 self.compress_context.low_cpu_mem_usage = False
         if len(all_blocks) > 1:
--- a/auto_round/context/model.py
+++ b/auto_round/context/model.py
@@ -14,6 +14,7 @@
 
 import gc
 import importlib
+import os
 from typing import Any, Callable, Optional, Union
 
 import torch
@@ -105,6 +106,12 @@ class ModelContext(BaseContext):
         self.need_calib = need_calib
         self.quant_nontext_module = quant_nontext_module
 
+        # Local addition (not upstream): remember the original string model name
+        # so OffloadManager can later materialize blocks directly from this
+        # checkpoint (see disk_stream_model_dir below and LOCAL_PATCHES.md).
+        self.disk_stream_model_dir = model if isinstance(model, str) else None
+        self._disk_stream_index = None
+
         # Load model and run basic initialization eagerly so the model is ready
         # by the time BaseCompressor.post_init() runs.
         self._load_model()
@@ -114,6 +121,8 @@ class ModelContext(BaseContext):
                 "AutoRound does not support parameters on meta device. "
                 "Please use more GPUs by setting `--device 0,1,2,3` or just place the model on CPU."
             )
+        if self._disk_stream_index is not None:
+            self._materialize_disk_stream_non_block_params()
         check_and_mark_quantized_module(self.model)
         self.model = self.model.eval()
         self.shared_cache_keys = get_shared_keys(self.model)
@@ -148,9 +157,42 @@ class ModelContext(BaseContext):
         if is_mllm_model(self.model, platform=self.platform):
             self.is_mllm = True
             if isinstance(self.model, str):
-                self.model, self.processor, self.tokenizer, self.image_processor = mllm_load_model(
-                    self.model, platform=self.platform, device="cpu", model_dtype=self.model_dtype
-                )
+                # Local addition (not upstream): multimodal checkpoints used to
+                # bypass disk streaming entirely -- mllm_load_model fully
+                # materializes the checkpoint on CPU, infeasible for a 100B+
+                # VLM (e.g. Ornith). Build the same meta skeleton the text
+                # path uses (build_meta_model resolves the multimodal class
+                # from config.architectures) and load the processor stack
+                # cheaply; non-block params INCLUDING the whole vision tower
+                # (small, and needed real for calibration forwards and RTN)
+                # are materialized right after the meta-device guard in
+                # __init__, exactly like the text path. Falls back to the
+                # full mllm_load_model on any failure.
+                loaded_via_meta = False
+                if envs.AR_DISK_STREAM_MODEL and os.path.isdir(self.model):
+                    try:
+                        self.model, self.tokenizer, self._disk_stream_index = self._build_disk_stream_model(
+                            self.disk_stream_model_dir
+                        )
+                        from transformers import AutoProcessor
+
+                        self.processor = AutoProcessor.from_pretrained(
+                            self.disk_stream_model_dir, trust_remote_code=self.trust_remote_code
+                        )
+                        self.image_processor = getattr(self.processor, "image_processor", None)
+                        loaded_via_meta = True
+                    except Exception:
+                        logger.warning(
+                            "AR_DISK_STREAM_MODEL requested but building a multimodal meta "
+                            "skeleton for %s failed; falling back to a full CPU load.",
+                            self.disk_stream_model_dir,
+                            exc_info=True,
+                        )
+                        self._disk_stream_index = None
+                if not loaded_via_meta:
+                    self.model, self.processor, self.tokenizer, self.image_processor = mllm_load_model(
+                        self.model, platform=self.platform, device="cpu", model_dtype=self.model_dtype
+                    )
         elif is_diffusion_model(self.model):
             self.is_diffusion = True
             self.pipe, self.model = diffusion_load_model(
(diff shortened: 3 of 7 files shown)


xin3he:
```AR_AUTO_SCHEME_NSAMPLES=8  AR_AUTO_SCHEME_BATCH_SIZE=8  auto-round /models/Qwen3-32B  --target_bits 4.5 --options MXFP4,MXFP8 --model_free --format llm_compressor  --shared_layers k_proj,v_proj,q_proj down_proj,up_proj  --low_cpu_mem_usage
2026-07-27 15:53:32 WARNING main.py L255: `low_cpu_mem_usage` is deprecated and is now enabled by default. To disable it, use `--disable_low_cpu_mem_usage`.
2026-07-27 15:53:32 INFO main.py L292: start to quantize /models/Qwen3-32B
2026-07-27 15:53:32 INFO model_free.py L2427: Model-free + AutoScheme: generating a per-layer scheme via delta-loss. The model is loaded temporarily for scoring, then released before shard-by-shard packing.
2026-07-27 15:53:32 INFO entry.py L756: Using LLM mode.
2026-07-27 15:53:32 WARNING logging.py L340: AutoRound entry received unsupported kwargs gradient_accumulate_steps. They will be ignored.
Loading weights: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████| 707/707 [00:00<00:00, 3769.96it/s]
[transformers] `loss_type=None` was set in the config but it is unrecognized. Using the default loss: `ForCausalLMLoss`.
2026-07-27 15:53:34 INFO gen_auto_scheme.py L104: force not using `low_cpu_mem_usage` in AutoScheme
2026-07-27 15:53:34 INFO gen_auto_scheme.py L237: AutoScheme option MXFP4 -> avg_bits=4.2500
2026-07-27 15:53:34 INFO gen_auto_scheme.py L237: AutoScheme option MXFP8 -> avg_bits=8.2500
2026-07-27 15:53:34 INFO gen_auto_scheme.py L132: Average bits range: [4.250, 8.250], target = 4.500
2026-07-27 15:53:35 INFO delta_loss.py L1613: AutoScheme steps(total)=256
2026-07-27 15:53:35 INFO delta_loss.py L1614: AutoScheme steps variables: scheme_num=2, block_num=64, nsamples=8, batch_size=8
2026-07-27 15:53:35 INFO delta_loss.py L1619: AutoScheme: nsamples/batch_size can be overridden via env vars AR_AUTO_SCHEME_NSAMPLES / AR_AUTO_SCHEME_BATCH_SIZE (e.g. `export AR_AUTO_SCHEME_NSAMPLES=1` for a quick run); see docs/environments.md for details.
2026-07-27 15:53:35 INFO delta_loss.py L1627: AutoScheme steps expanded(low_gpu): total_steps = scheme_num * block_num * 2(forward+backward) * n_batches = 2 * 64 * 2 * 1 = 256
Generating AutoScheme:   0%|                                                                                                                   | 0/256 [00:00<?, ?it/s]2026-07-27 15:53:35 INFO delta_loss.py L1697: AutoScheme transition: switch to scheme 1/2 (MXFP4)
2026-07-27 15:53:35 INFO calib_dataset.py L977: Preprocessing calibration dataset in a subprocess to avoid memory leaks...
Generating AutoScheme:  25%|██████████████████████████▌                                                                               | 64/256 [00:45<01:54,  1.67it/s]2026-07-27 15:54:20 INFO device.py L1450: 'peak_ram': 63.8GB, 'peak_vram': 12.84GB
/home/xinhe/auto-round-main/.venv/lib/python3.12/site-packages/torch/autograd/graph.py:869: UserWarning: Flash Attention defaults to a non-deterministic algorithm. To explicitly enable determinism call torch.use_deterministic_algorithms(True, warn_only=False). (Triggered internally at /pytorch/aten/src/ATen/native/transformers/cuda/attention_backward.cu:124.)
  return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
Generating AutoScheme:  50%|████████████████████████████████████████████████████▌                                                    | 128/256 [01:32<01:52,  1.14it/s]2026-07-27 15:55:07 INFO utils.py L574: AutoScheme non_block loss: none
2026-07-27 15:55:07 INFO device.py L1450: 'peak_ram': 63.8GB, 'peak_vram': 14.28GB
```
@aquilarubra I tried Qwen-32B and notice that the ram usage is 63.8GB, which is not reduced.
Could you please have a check?

aquilarubra:
Thanks for testing this against Qwen3-32B! I traced through this and I believe the streaming is working as designed, but wasn't enabled in your run — the repro command only sets `AR_AUTO_SCHEME_NSAMPLES`/`AR_AUTO_SCHEME_BATCH_SIZE`, not `AR_DISK_STREAM_MODEL=1` (which is opt-in, see `docs/environments.md`). Without it, `model_free` + `AutoScheme`'s fallback path does a normal full-precision load for scoring — for Qwen3-32B in bf16 that's ~64GB, which lines up with your 63.8GB almost exactly.

I confirmed the mechanism does reduce peak RAM when the env var is set (tested on Qwen3-0.6B: 3.53GB → 2.15GB for the AutoScheme scoring phase). Could you re-run with `export AR_DISK_STREAM_MODEL=1` and let me know if that resolves it?

xin3he:
Thanks @aquilarubra 

Another issue happened with this branch. 

Here is my command:
```
AR_DISK_STREAM_MODEL=1 AR_AUTO_SCHEME_NSAMPLES=8  AR_AUTO_SCHEME_BATCH_SIZE=8  auto-round /models/Qwen3-0.6B  --target_bits 4.5 --options MXFP4,MXFP8 --model_free --format llm_compressor  --shared_layers k_proj,v_proj,q_proj down_proj,up_proj
```
```
  File "/home/xinhe/auto-round-main/auto_round/compressors/base.py", line 736, in configure_layer_config
    self.layer_config = self._gen_auto_scheme()
                        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xinhe/auto-round-main/auto_round/compressors/base.py", line 689, in _gen_auto_scheme
    layer_config = self.scheme_generator.get_layer_config()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xinhe/auto-round-main/auto_round/auto_scheme/gen_auto_scheme.py", line 161, in get_layer_config
    layer_config = method_func(
                   ^^^^^^^^^^^^
  File "/home/xinhe/auto-round-main/auto_round/auto_scheme/delta_loss.py", line 2044, in gen_layer_config
    model.to("cpu")
  File "/home/xinhe/auto-round-main/.venv/lib/python3.12/site-packages/transformers/modeling_utils.py", line 3761, in to
    return super().to(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xinhe/auto-round-main/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1384, in to
    return self._apply(convert)
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/xinhe/auto-round-main/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 934, in _apply
    module._apply(fn)
  File "/home/xinhe/auto-round-main/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 934, in _apply
    module._apply(fn)
  File "/home/xinhe/auto-round-main/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 934, in _apply
    module._apply(fn)
  [Previous line repeated 2 more times]
  File "/home/xinhe/auto-round-main/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 965, in _apply
    param_applied = fn(param)
                    ^^^^^^^^^
  File "/home/xinhe/auto-round-main/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1377, in convert
    raise NotImplementedError(
NotImplementedError: Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device.
```

aquilarubra:
Tried to reproduce this and want to give an honest status rather than keep guessing.

**What I tested:** your exact command flags (`--target_bits 4.5 --options MXFP4,MXFP8 --model_free --format llm_compressor --shared_layers k_proj,v_proj,q_proj down_proj,up_proj`, `AR_DISK_STREAM_MODEL=1`) against Qwen3-0.6B — both as a single `model.safetensors` file and manually re-sharded into 4 files + `model.safetensors.index.json` (to mimic how a 32B checkpoint is actually laid out on disk, since that's the main structural difference from a small model). I also instrumented `materialize_module`/`free_module` directly (bypassing log-level filtering — that's why my earlier reply's "confirmation" wasn't fully reliable, I was reading `peak_ram` deltas that are small enough on a 0.6B model to be noise) and confirmed streaming genuinely engages: 112 materialize/free calls, exactly matching the scoring step count. No crash in either case, sharded or not.

**What I couldn't reproduce:** the `NotImplementedError: Cannot copy out of meta tensor` at `delta_loss.py:2044`. That line only runs when `disk_index` is `None` at that point in `gen_layer_config` — but your traceback shows the model *did* have live meta tensors at crash time, meaning streaming had engaged earlier and something caused the index to not be found later. I traced the index hand-off (`context/model.py` stashes it as `model._disk_stream_index`, `delta_loss.py` reads it back via `getattr`) and couldn't find a gap, but I can't rule out something specific to the real 32B checkpoint that a 0.6B model (even resharded) doesn't trigger.

Could you share:
1. The full log from process start (not just the traceback tail) — specifically whether a `"AR_DISK_STREAM_MODEL requested but building a meta skeleton ... failed"` warning appeared earlier in the run.
2. Which exact commit/branch you tested against (`/home/xinhe/auto-round-main` — want to make sure it's this PR's latest and not a different combination).

I'm also going to pull down Qwen3-32B myself and try to reproduce directly rather than keep guessing from a distance.

aquilarubra:
Update: I downloaded Qwen3-32B and ran your exact command against this PR's latest commit, adding `AR_DISK_STREAM_MODEL=1`:

```
AR_DISK_STREAM_MODEL=1 AR_AUTO_SCHEME_NSAMPLES=8 AR_AUTO_SCHEME_BATCH_SIZE=8 \
auto-round Qwen/Qwen3-32B --target_bits 4.5 --options MXFP4,MXFP8 --model_free \
--format llm_compressor --shared_layers k_proj,v_proj,q_proj down_proj,up_proj --low_cpu_mem_usage
```

Result: completed cleanly, no crash — `'peak_ram': 4.1GB, 'peak_vram': 14.31GB'` for the AutoScheme scoring phase (vs. your reported 63.8GB without the env var, which is close to the full unquantized model size). Same `scheme_num=2, block_num=64` as your log, so this is exercising the same 32B/64-block path.

Since I can't reproduce either the original RAM issue or the follow-up crash on this branch with the env var set, I suspect what you're running (`/home/xinhe/auto-round-main`) differs from this PR's current state in some way — maybe a different commit, or another change layered on top. Could you check that you're on this exact branch tip and try again with `AR_DISK_STREAM_MODEL=1`? Happy to keep digging if it still crashes for you.

wenhuach21:
@aquilarubra 
Really nice job! I didn’t expect RAM usage could be reduced to such an extent. I’d really appreciate it if you could keep refining our repo

wenhuach21:
I can reproduce the same issue. @xin3he, could you help investigate and fix it? Fabrizio wasn't able to reproduce it.
~~~bash
Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device
~~~

aquilarubra:
@wenhuach21 @xin3he Think I found the actual root cause, and it's not an env/library difference — it's a sequencing gap between this PR and #2063.

`#2061` never touches `auto_round/auto_scheme/delta_loss.py` (verified via `git diff origin/main...pr/disk-streaming-core --stat`). It builds an all-meta skeleton *unconditionally* whenever `AR_DISK_STREAM_MODEL=1` is set, regardless of what quantization mode runs afterward. AutoScheme's delta-loss scoring only learns to handle a meta model in **#2063**, which adds the `elif disk_index is not None: ...` guard around the exact `model.to("cpu")` call in the traceback — skipping it because the decoder blocks are intentionally still meta at that point. Without #2063 stacked on top, that guard doesn't exist, so stock AutoScheme code calls `.to("cpu")` on a genuinely-meta model and hits `NotImplementedError: Cannot copy out of meta tensor` — deterministically, every time `AR_DISK_STREAM_MODEL=1` is combined with AutoScheme (`--options ...`) and only #2061 is applied.

This lines up with everything observed: I only ever tested against the full stack (all 5 PRs combined, including #2063's guard), which is why I couldn't reproduce it despite matching flags exactly, including a real Qwen3-32B run. #2061's own PR description calls out that it's "first of a small stack... disk streaming → resumability → AutoScheme streaming" — AutoScheme + disk streaming specifically needs #2063.

Could you confirm whether you tested #2061 by itself, or with #2063 also checked out/merged in? If it was #2061 alone, could you re-test either:
1. with `fork/pr/autoscheme-streaming` (#2063) also stacked on top, or
2. the same command but without `--options MXFP4,MXFP8`/AutoScheme (plain disk streaming only)

to confirm the foundation PR itself is solid? If the crash still reproduces under either of those, that would point back to something environment-specific and I'll dig further with your exact `pip freeze` (torch/transformers/accelerate/safetensors versions) and full log.

xin3he:
> @wenhuach21 @xin3he Think I found the actual root cause, and it's not an env/library difference — it's a sequencing gap between this PR and #2063.
> 
> `#2061` never touches `auto_round/auto_scheme/delta_loss.py` (verified via `git diff origin/main...pr/disk-streaming-core --stat`). It builds an all-meta skeleton _unconditionally_ whenever `AR_DISK_STREAM_MODEL=1` is set, regardless of what quantization mode runs afterward. AutoScheme's delta-loss scoring only learns to handle a meta model in **#2063**, which adds the `elif disk_index is not None: ...` guard around the exact `model.to("cpu")` call in the traceback — skipping it because the decoder blocks are intentionally still meta at that point. Without #2063 stacked on top, that guard doesn't exist, so stock AutoScheme code calls `.to("cpu")` on a genuinely-meta model and hits `NotImplementedError: Cannot copy out of meta tensor` — deterministically, every time `AR_DISK_STREAM_MODEL=1` is combined with AutoScheme (`--options ...`) and only #2061 is applied.
> 
> This lines up with everything observed: I only ever tested against the full stack (all 5 PRs combined, including #2063's guard), which is why I couldn't reproduce it despite matching flags exactly, including a real Qwen3-32B run. #2061's own PR description calls out that it's "first of a small stack... disk streaming → resumability → AutoScheme streaming" — AutoScheme + disk streaming specifically needs #2063.
> 
> Could you confirm whether you tested #2061 by itself, or with #2063 also checked out/merged in? If it was #2061 alone, could you re-test either:
> 
> 1. with `fork/pr/autoscheme-streaming` ([Stream AutoScheme's per-layer sensitivity scoring block-by-block #2063](https://github.com/intel/auto-round/pull/2063)) also stacked on top, or
> 2. the same command but without `--options MXFP4,MXFP8`/AutoScheme (plain disk streaming only)
> 
> to confirm the foundation PR itself is solid? If the crash still reproduces under either of those, that would point back to something environment-specific and I'll dig further with your exact `pip freeze` (torch/transformers/accelerate/safetensors versions) and full log.

Yes, I tested #2061 by itself and #2063 works alone. I commented in #2063

xin3he (approved):
LGTM

xin3he:
@aquilarubra 
I prefer to merge #2063 directly, since it provides a solution for AutoScheme.

For normal tuning, we have enabled block-by-block loading and offloading, and I didn't observe a significant memory drop.
While, for Autoscheme, we need to execute model level forward which requires loading all params at once, and the disk streaming could help a lot.

aquilarubra:
@xin3he @wenhuach21 Sounds good — happy to go with merging #2063 directly.

One thing worth flagging: #2063's branch is stacked on top of #2061 (it needs the meta-skeleton scaffolding #2061 builds), so merging #2063 as-is will bring #2061's commits along with it too — there isn't a version of #2063 that doesn't include that foundation. That should be fine either way, just didn't want it to be a surprise in the diff.

The other thing: #2062 (resumability, AR_RESUME_DIR) and #2064 (targeted block re-quantization streaming) are both also stacked on top of #2061, so once #2063 merges (in whatever form), those two will need to be rebased onto the merged result before they're mergeable themselves. I'll take care of that.

Heads-up on timing: I'll be on vacation for the next 3 weeks starting now, so I won't be able to respond to review comments or rebase anything in that window. If #2063 merges while I'm out, #2061/#2062/#2064 may show conflicts in the meantime — that's expected and I'll clean it up as soon as I'm back. #2065 (mllm dataset fix) is independent of all this and can be merged on its own whenever it's ready.

Thanks for the reviews, and sorry for the radio silence coming up!

xin3he:
@aquilarubra Thank you for the update, wishing you a pleasant vacation. 

Here I attach some data for our future design/decision. I think the difference comes from the runtime splitting of fused MOE tensors.

`auto-round /models/Qwen3.6-35B-A3B/ --iters 1 --enable_torch_compile`
- main branch -> disk_stream: AR_DISK_STREAM_MODEL=1 
- 10/40 [04:17 -> 10/40 [06:00 # time usage after quantizing 10/40 blocks
- 'peak_ram': 26.13GB, 'peak_vram': 26.5GB  -> 'peak_ram': 25.37GB, 'peak_vram': 28.05GB

xin3he:
close it since #2063 is merged already

PR #2062, opened 2026-07-19
Add crash/resume support for tuning runs (AR_RESUME_DIR)

aquilarubra:
## Summary

Depends on #2061 (disk-streaming core) — this branch is stacked on top of it, so the diff below includes those commits until #2061 merges; only the last 5 commits are new here.

Adds an opt-in `AR_RESUME_DIR` env var: when set, the tuning loop (both the standard SignRound loop and the `--iters 0` RTN-with-imatrix path) checkpoints its progress to disk after each completed block, so a crash or kill mid-run can resume from the first not-yet-done block instead of restarting from block 0. Useful independently of disk streaming, but especially so for very large models where a full run spans hours.

## What's in this PR

- `auto_round/utils/resume.py` (new): `ResumeState`, tracking completed blocks plus cached chain tensors (`q_input`/`input_ids`) keyed by a signature hash over model/scheme/dataset/block-list, so a reused directory from an incompatible run is detected and ignored. Also `layer_config_fingerprint()`, folded into the signature so two `AutoScheme` runs with different `avg_bits` targets (which otherwise produce identical signatures) don't silently resume each other's state.
- `auto_round/compressors/data_driven.py`: wires `ResumeState` into both the standard tuning loop and the RTN-with-imatrix path (the latter needed a prerequisite fix first — it never wrote per-block output during its loop at all, only in one bulk pass at the end, which would have made resumability actively unsafe).
- `auto_round/compressors/shard_writer.py`: makes shard-based export resume-aware across process restarts (discovers a crashed prior process's already-flushed shards and continues numbering from there).
- `auto_round/export/export_to_autoround/export.py`: skips packing for a resumed, still-meta layer (its packed weights already live in a previously-flushed shard).
- `auto_round/compressors/base.py` (already in #2061): defers clearing the resume manifest until `save_quantized()` actually succeeds, not right after tuning finishes.

## Validation

Verified via real crash-and-resume tests (`kill -9` mid-run, fresh process resumes) across dense/MoE architectures, both `iters>0` and `iters=0`, both bare `.quantize()` and shard-export paths: resumed output is bit-identical to an uninterrupted control run in every case, confirmed by diffing full tensor sets. Also verified at real 207GB-checkpoint scale.

Happy to squash/reorganize commits if a different granularity is preferred for review.

### DIFF
--- a/auto_round/compressors/base.py
+++ b/auto_round/compressors/base.py
@@ -393,6 +393,12 @@ class BaseOrchestrator(object):
         # second AutoRound(...) call reuses the previous instance and silently keeps
         # stale values (e.g. low_cpu_mem_usage=True from a prior run).
         CompressContext.reset_context()
+        # When the model was built as a meta skeleton (AR_DISK_STREAM_MODEL=1),
+        # give the offloader the original checkpoint path so it can materialize
+        # each block on first touch directly from disk instead of assuming
+        # blocks already hold real weights (see OffloadManager._reload).
+        if self.model_context.disk_stream_model_dir is not None:
+            self._offloader.model_dir = self.model_context.disk_stream_model_dir
         # Alternatively, you can use CompressContext.create_context
         self.compress_context = CompressContext(
             low_cpu_mem_usage,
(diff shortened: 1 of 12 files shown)


xin3he:
Hi @aquilarubra FYI. here is a fix: https://github.com/intel/auto-round/pull/2220
Please resolve conflicts after #2220 is merged to reduce the workload.
It's a good feature!

aquilarubra:
I'll take a look during the weekend.

aquilarubra:
Closing — this content is fully superseded by #2220, which already merged the resumability/AR_RESUME_DIR support this PR was adding. I rebased locally to confirm: the diff against current `main` is empty, byte-identical. Nothing left to resubmit here.

PR #2063, opened 2026-07-19
Stream AutoScheme's per-layer sensitivity scoring block-by-block

aquilarubra:
## Summary

Depends on #2061 (disk-streaming core) — stacked on top of it, so the diff includes those commits until it merges; only the last commit is new here.

`AutoScheme`'s mixed-bit search scores every `(scheme, block)` pair to build its DP knapsack input — a second, independent place (besides the compressor's own tuning loop) that needs real weights one block at a time, and one that runs *before* the compressor's `OffloadManager` cycle even starts.

## What's in this PR

- `auto_round/auto_scheme/gen_auto_scheme.py`: removes `GenScheme.__init__`'s unconditional `low_cpu_mem_usage = False` override (added in `0c9c5b1d` to work around a bug in an old `OffloadManager`-based streaming path this patch set doesn't use).
- `auto_round/auto_scheme/delta_loss.py`: threads a `disk_index` parameter through AutoScheme's scoring call sites, materializing/freeing each decoder block around scoring instead of assuming it's already resident or moving it with `.to("cpu")` (which never actually frees RAM once a block started resident). Reads `disk_index` off the model object itself (set by `ModelContext` in #2061), so streaming works through the standard `AutoRound(model=path, scheme=AutoScheme(...))` API with no caller-visible changes. Also skips two unconditional `model.to("cpu")` calls that assume full CPU-residency and otherwise crash with "Cannot copy out of meta tensor" for a still-partially-meta model.

## Validation

Verified end-to-end (`AutoScheme(avg_bits=4.5, options=["W4A16","W8A16"])` + tuning) on synthetic fixtures up to 27.7GB: peak RAM stays flat through both scoring and tuning phases (vs. proportional-to-checkpoint-size on the baseline), and the resulting mixed-bit `layer_config` is byte-identical to the unstreamed baseline. Also verified at real 207GB scale.

### DIFF
--- a/auto_round/auto_scheme/delta_loss.py
+++ b/auto_round/auto_scheme/delta_loss.py
@@ -563,7 +563,7 @@ class MyCustomError(Exception):
 last_grad_input = None
 
 
-def prepare_model_low_gpu(model, block_inputs: dict = None, pbar=None, major_device="cpu"):
+def prepare_model_low_gpu(model, block_inputs: dict = None, pbar=None, major_device="cpu", disk_index=None):
     """Wrap every block's forward so that, for one calibration batch, it (1) moves itself to
     ``major_device`` on demand, (2) records its own inputs into ``block_inputs`` (on CPU) so
     they can be replayed later, and (3) moves itself back to CPU once done.
@@ -571,6 +571,12 @@ def prepare_model_low_gpu(model, block_inputs: dict = None, pbar=None, major_dev
     Called once per calibration batch before ``model_forward_low_gpu`` runs the actual
     forward+backward -- the recorded ``block_inputs`` are what let the backward pass be
     replayed manually, one block at a time, without keeping every block resident on GPU.
+
+    When ``disk_index`` is set (streaming mode -- the model is a meta-device skeleton,
+    see ``gen_layer_config``/``disk_stream_util.py``), each block's real weights are
+    materialized from the checkpoint right before its own forward and released back to
+    meta right after, instead of assuming the block already has real CPU-resident weights
+    to shuffle to GPU and back.
     """
     block_inputs.clear()
     for n, m in model.named_modules():
@@ -590,6 +596,10 @@ def prepare_model_low_gpu(model, block_inputs: dict = None, pbar=None, major_dev
             """Move the block to device, run its original forward, cache its (CPU) inputs
             for later replay, then move the block back to CPU.
             """
+            if disk_index is not None:
+                from auto_round.utils.disk_stream_util import materialize_module
+
+                materialize_module(module, module_name, disk_index, device=major_device)
             move_module_to_tuning_device(module, major_device=major_device)
             # for n,m in module.named_modules():
             #     if hasattr(m, "post_init_qdqw"):
@@ -608,7 +618,12 @@ def prepare_model_low_gpu(model, block_inputs: dict = None, pbar=None, major_dev
             }
             block_inputs[module_name] = input_info
 
-            module.to("cpu")
+            if disk_index is not None:
+                from auto_round.utils.disk_stream_util import free_module
+
+                free_module(module)
+            else:
+                module.to("cpu")
             memory_monitor.update(device_list=major_device)
             # clear_memory(device_list=major_device) #slow
             # memory_monitor.log_summary()
@@ -685,7 +700,7 @@ def model_forward(model, data, **forward_kwargs):
     return model(**prepared, **forward_kwargs), prepared
 
 
-def model_forward_low_gpu(model, dataloader, major_device="cuda", pbar=None, scheme_tag=None):
+def model_forward_low_gpu(model, dataloader, major_device="cuda", pbar=None, scheme_tag=None, disk_index=None):
     """Run one full scoring pass (all calibration batches) in low-GPU-memory mode.
 
     For each batch: capture per-block inputs via ``prepare_model_low_gpu``, run a forward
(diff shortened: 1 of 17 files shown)


wenhuach21:
Really appreciate this PR! It's exactly what we needed.

We're currently in a code freeze while we refactor the architecture, so we aren't merging new feature PRs at the moment. Once the refactoring is complete (ETA 1 week), we'll merge your PR

auto_round/auto_scheme/delta_loss.py
```diff
@@ -554,14 +554,21 @@ def __init__(self, message):
...
+def prepare_model_low_gpu(model, block_inputs: dict = None, pbar=None, major_device="cpu", disk_index=None):
     """Wrap every block's forward so that, for one calibration batch, it (1) moves itself to
     ``major_device`` on demand, (2) records its own inputs into ``block_inputs`` (on CPU) so
     they can be replayed later, and (3) moves itself back to CPU once done.
 
     Called once per calibration batch before ``model_forward_low_gpu`` runs the actual
     forward+backward -- the recorded ``block_inputs`` are what let the backward pass be
     replayed manually, one block at a time, without keeping every block resident on GPU.
+
+    Local addition: when ``disk_index`` is set (streaming mode -- the model is a
```
lvliang-intel: Nearly every hunk is commented "Local addition (not upstream)" and references a LOCAL_PATCHES.md and vendor/reap//vendor/auto_round/ paths that don't exist in this repo.

auto_round/auto_scheme/delta_loss.py
```diff
@@ -795,7 +829,20 @@ def backward_pre_hook(module, grad_input):
...
             del block_output, main_output, block_input_args, block_input_kwargs
-            block_module.to("cpu")
+            if disk_index is not None:
+                from auto_round.utils.disk_stream_util import free_module, total_resident_bytes
+
+                free_module(block_module)
+                import os as _os
+
+                if _os.environ.get("AR_STREAM_DEBUG"):
+                    print(
```
lvliang-intel: use logger instead of print

auto_round/auto_scheme/delta_loss.py
lvliang-intel, line 1: please add related UT cases and update document.

xin3he:
Hi @aquilarubra 
I will merge this feature recently, and it will cause lots of conflicts. Please take a look. https://github.com/intel/auto-round/pull/2083

We have observed that using the default `--low_gpu_mem_usage` parameter results in low GPU utilization when evaluating GLM-5.2/Hy3.
Therefore, we plan to leverage multi-processing to fully utilize GPU performance, though this will double memory/VRAM usage (depending on the chosen approach).

It would be very helpful if your change could reduce the memory usage based on that.

aquilarubra:
Thanks for the heads-up @xin3he — yes, I see the overlap in `delta_loss.py`/`gen_auto_scheme.py`. Happy to rebase once #2083 lands.

On the memory question: this PR's `AR_DISK_STREAM_MODEL` mode already keeps each process's resident footprint roughly flat (blocks are materialized from the checkpoint on demand and freed right after scoring), rather than proportional to model size. Looking at your benchmark table, that seems complementary to #2083's multi-process approach rather than competing with it — each worker process could run with disk-streaming enabled, so the memory cost scales with worker count × block size instead of worker count × full model size. I can help validate that combination once #2083 is in and rebase this PR on top of it.

wenhuach21 (approved):

wenhuach21:
@aquilarubra recently we have refactored the arch to better support more algorithms, could you kindly help fix the conflict? then we can merge your prs.
Thanks again for the prs!

xin3he:
Hi @aquilarubra 
Thank you for your contribution! This branch is good, I tried and the memory is truly reduced. Great job!
Previously I was using https://github.com/intel/auto-round/pull/2061 and it not correct. I believe we don't need 2061 if we got this PR merged, right?

FYI. #2083 is merged now and you can resolve the conflicts. 
Currently I disabled the autoscheme parallel due to huge ram usage. You can verify it by setting `export AR_ENABLE_AUTO_SCHEME_PARALLEL=1`.

Looking forward to getting this PR merged soon. 😀  I plan to set the `AR_ENABLE_AUTO_SCHEME_PARALLEL` as default with `AR_DISK_STREAM_MODEL=1`. We can expect a higher speed for autoscheme with both optimization. 
cc @wenhuach21 @lvliang-intel @n1ck-guo

wenhuach21:
> Hi @aquilarubra Thank you for your contribution! This branch is good, I tried and the memory is truly reduced. Great job! cc @wenhuach21
> 
> #2083 is merged now. Currently I disabled the autoscheme parallel due to huge ram usage. You can verified by setting `export AR_ENABLE_AUTO_SCHEME_PARALLEL=1`.
> 
> Looking forward to getting this PR merged soon. 😀

Great! How about the runtime? If the overhead is low, we could make this the default.

xin3he (approved):
Nice work!
@aquilarubra 
Please help resolve the conflicts, Please feel free to ask help if you have any issue.

xin3he:
@copilot resolve the merge conflicts in this pull request

aquilarubra:
@xin3he Thanks for reviewing — glad the memory reduction checks out. Quick note left on #2061 for context: this branch is stacked on it, so merging #2063 brings #2061's commits along too, and #2062/#2064 (also stacked on #2061) will need a rebase once this lands.

Also flagging: I'll be on vacation for the next 3 weeks starting now, so I won't be able to respond to review comments or handle the rebase during that window — will pick it back up once I'm back. #2065 is independent and fine to merge separately anytime.

PR #2064, opened 2026-07-19
Stream targeted block re-quantization's calibration pass

aquilarubra:
## Summary

Depends on #2061 (disk-streaming core) — stacked on top of it, so the diff includes those commits until it merges; only the last commit is new here.

`to_quant_block_names` lets a caller restrict tuning to a subset of decoder blocks — useful for cheaply re-quantizing just a couple of blocks in an already-produced checkpoint at higher precision instead of redoing a full multi-hour run. Combined with `AR_DISK_STREAM_MODEL`, this crashes: the "cache block inputs" calibration forward pass needs real weights in every block leading up to (and sometimes through) the target block(s), but nothing materializes blocks outside `quant_block_list` for this specific pass — they stay meta forever, and the forward silently propagates meta-ness until it collides with a genuinely-materialized module. Full (unrestricted) runs never hit this, since `quant_block_list` already covers every block in that case.

## What's in this PR

`auto_round/calibration/llm.py`: when disk streaming is active and any decoder block still has meta parameters at the point this calibration forward runs, wraps it with the existing `stream_block_forward` primitive (from #2061), scoped to just the still-meta blocks. Only activates when there's something left meta to fix, so it's a no-op for the normal full-quantization path.

Also adds an `AR_CALIB_STREAM_DEVICE` env-gated fast path: the default keeps this forward entirely on CPU (mixing a GPU-streamed block with CPU-resident hidden states crashes with a device mismatch), but a full CPU forward through every pre-target block of a 100B+ model is unusably slow for this specific targeted-requant use case. When set, every already-real tensor is moved to that device for the pass's duration and moved back afterward.

## Validation

Reproduced and fixed against a tiny hybrid-MoE fixture with an MTP head, with both a single restricted block and two adjacent ones, plus a control run of the full (unrestricted) path confirming no regression.


aquilarubra:
Rebased onto current `main` — the previous stack this was built on (#2061/#2063) is fully merged now, so I rebuilt this PR's unique diff on top of the merged result instead of a mechanical rebase. Most of the old diff (streaming helpers, resumability plumbing, parallel-scoring gating) turned out to already be upstream via #2063/#2220; what's left is the actual calibration-streaming feature in `auto_round/calibration/llm.py` plus docs/tests. Force-pushed to this branch.

docs/environments.md
```diff
@@ -231,6 +231,26 @@ export AR_DISK_STREAM_MODEL=1
 export AR_RESUME_DIR=/path/to/resume/state
 ```
 
+### AR_DISK_STREAM_MODEL
```
xin3he: We already have this variable; please clean up the relevant documentation.

docs/environments.md
```diff
@@ -231,6 +231,26 @@ export AR_DISK_STREAM_MODEL=1
...
+- **Description**: When enabled, `AutoRound(model=<path>, ...)` builds the model as a meta-device skeleton instead of fully materializing the checkpoint on CPU RAM up front, and streams each decoder block's real weights from the checkpoint's safetensors shards on demand -- materializing right before a block is used (calibration, tuning, or `AutoScheme` sensitivity scoring) and freeing it back to meta right after. This keeps peak CPU RAM roughly flat regardless of checkpoint size, instead of proportional to it. Non-block parameters (embeddings, `lm_head`, final norm) are still loaded up front, since they are typically small.
+- **Default**: `False`
+- **Valid Values**: `"1"`, `"true"`, `"yes"` (case-insensitive) for enabling; any other value for disabling
+- **Usage**: Enable this to quantize checkpoints larger than available CPU RAM + GPU VRAM combined. Only applies when `model` is a string (local directory) path; has no effect on already-loaded model objects.
+
+```bash
+export AR_DISK_STREAM_MODEL=1
+```
+
+### AR_RESUME_DIR
```
xin3he, line 234: We already have this variable; please clean up the relevant documentation.

xin3he (commented):
Nice catch, please clean up the document

docs/environments.md
aquilarubra: Good catch — the merge with main had duplicated the AR_DISK_STREAM_MODEL and AR_RESUME_DIR sections verbatim. Removed the duplicate copy, kept the richer one (mentions the parallel-scoring interaction). Pushed.

wenhuach21:
@Copilot move the cpu test to cuda test

wenhuach21:
@copilot move the cpu test to cuda test

wenhuach21:
@aquilarubra  Hi, the cpu ut is time out, we may need to move the new ut to cuda

auto_round/calibration/llm.py
```diff
@@ -81,16 +81,108 @@ def calibration(self, block_names, nsamples, layer_names=None, last_cache_name=N
...
+            # never hit this: quant_block_list already covers every block in
+            # that case, so there's nothing left outside it to leave meta.
+            # Fix: stream-materialize (then free) any block that's still meta
+            # for the duration of this one forward pass, reusing the same
+            # stream_block_forward primitive _streaming_eval_model() already
+            # uses in bin/el_quantize_autoround_mixed.py for the analogous
+            # held-out-loss-eval case.
+            stream_ctx = None
+            _moved_tensors = []
+            if envs.AR_DISK_STREAM_MODEL:
```
wenhuach21: For non-MoE models, Transformers already works this way, so no extra ops or code are needed. For MoE models, however, the current PR is not sufficient. The main branch already includes an attempt to address the MoE issue.

So I suppose there is no need for this PR. Please feel free to correct me if I am wrong.

PR #2065, opened 2026-07-19
Allow non-MLLM calibration datasets when not quantizing vision

aquilarubra:
## Summary

Standalone fix, no dependency on the other PRs in this stack.

`MLLMCalibrator` unconditionally routes to `get_mllm_dataloader` whenever the model is multimodal, which `KeyError`s on any dataset not in its `MLLM_DATASET` registry (its own `os.path.isfile(dataset) or dataset in MLLM_DATASET` guard still indexes `MLLM_DATASET` with the raw file path). But a local text dataset (file/dir/HF text set) is perfectly valid calibration for a VLM whose non-text modules are **not** being quantized (`quant_nontext_module=False`): the full-model forward runs text-only, and the loop's generic-dict branch already just feeds `model(**batch)` regardless of dataset shape.

## Change

When `quant_nontext_module` is `False` and the dataset isn't a known `MLLM_DATASET` entry, use the standard text `get_dataloader()` instead of falling through to `get_mllm_dataloader`.


aquilarubra:
Heads-up on the `Performance-Test-AutoRound` (`Qwen3_FP8_STATIC`) check on this PR — it's failed twice on retrigger, but on opposite sides each time:

- Run 1: Current = 22.83GB, Baseline = 20.51GB (+11.31%) → FAIL
- Run 2: Current = 20.55GB, Baseline = 23.58GB (-12.85%) → FAIL

The baseline number itself swings ~15% between runs of the identical job, and the sign of the "regression" flips. This PR only touches `calibration/mllm.py` (routing text-only datasets for non-vision calibration), which has no code path anywhere near FP8_STATIC quantization memory usage, so I don't think this reflects a real change from this diff — looks like VRAM measurement noise on the runner (possibly contention with other concurrent jobs) combined with a tolerance that's too tight to absorb it. Flagging in case it's useful for tightening up the perf-check baseline/tolerance.

aquilarubra:
Rebased onto current `main`. This PR only touches `calibration/mllm.py` and is independent of the disk-streaming/resumability stack, so the rebase was clean with no conflicts. Force-pushed to this branch.

auto_round/calibration/mllm.py
```diff
@@ -108,27 +108,49 @@ def calib(self, nsamples: int, bs: int) -> None:
...
-                bs=bs,
-                seed=self.seed,
-                nsamples=nsamples,
-                quant_nontext_module=self.quant_nontext_module,
-            )
-            if orig_bs != 1 and self.batch_size == 1:
-                self.is_only_supported_bs1 = True
+            from auto_round.compressors.mllm.dataset import MLLM_DATASET
+
+            if not self.quant_nontext_module and dataset not in MLLM_DATASET:
```
xin3he, line 113: @aquilarubra  If self.quant_nontext_module is False, How can we calibration the vision module without MLLM_DATASET?

aquilarubra: When quant_nontext_module=False, the vision/audio towers aren't being quantized at all, so there's nothing to calibrate them for — this branch only runs when we're staying entirely on the text path (full-model forward with a text-only batch, feeding model(**batch)). If quant_nontext_module=True we don't hit this branch: the outer elif above requires an MLLM_DATASET in that case, matching upstream behavior. So no code path here tries to calibrate vision without MLLM_DATASET while it's actually being quantized.

wenhuach21 (approved):

PR #1971, opened 2026-07-01
[AutoScheme] fix lm_head no_grad issue and support MOE model

xin3he:
## Description

1. Fix moe model meta tensor issue (materialize_model_)
2. Fix lm_head mix_score = 0 issue (no grad)
3. Set shared mix_score for inactive experts
4. Reduce `nsamples` and `seq_len` for MoE models to shorten execution time
  - Execution time for qwen3.6-35B-A3B reduced from 4 hours to 30 minutes.
5. Add debug log to check the inactive experts and average loss per block.
6. add Env variables: AR_AUTO_SCHEME_NSAMPLES, AR_AUTO_SCHEME_BATCH_SIZE

**Testing improvements:**

* Added `test_lm_head_mix_score_nonzero` to verify that the `lm_head` receives a non-zero `mix_score` across quantization schemes in the `low_gpu_mem_usage` path, ensuring correct score computation and increasing test coverage.
* verified with Qwen3.6-35B-A3B

BTW, enhanced log in debug mode.

#### meta-llama/Llama-3.2-1B-Instruct

<details>

command: ` AR_LOG_LEVEL=debug auto_round   --model_name  /models/Llama-3.2-1B-Instruct/   --avg_bits 6   --options "mxfp4,mxfp8"  --quant_lm_head`
```
2026-07-01 11:44:08 INFO delta_loss.py L1791: AutoScheme steps(total)=64                                        
2026-07-01 11:44:08 INFO delta_loss.py L1792: AutoScheme steps variables: scheme_num=2, block_num=16, nsamples=1
6, batch_size=8                                                                                                 
2026-07-01 11:44:08 INFO delta_loss.py L1799: AutoScheme steps expanded(low_gpu): total_steps = scheme_num * blo
ck_num * 2(forward+backward) * n_batches = 2 * 16 * 2 * 2 = 128                                                 
Generating AutoScheme:   0%|                                                           | 0/128 [00:00<?, ?it/s]2
026-07-01 11:44:08 INFO delta_loss.py L1836: AutoScheme transition: switch to scheme 1/2 (MXFP4)                
2026-07-01 11:44:08 INFO calib_dataset.py L977: Preprocessing calibration dataset in a subprocess to avoid memor
y leaks...                                                                                                      
Generating AutoScheme:  12%|??????????????                                           | 16/128 [00:16<00:36,  3.07it/s]2
026-07-01 11:44:25 INFO device.py L1450: 'peak_ram': 5.63GB, 'peak_vram': 3.92GB                                
[...]

PR #1846, opened 2026-05-22
fix bug of qwen and gguf export

n1ck-guo:
## Description
This PR updates AutoRound’s GGUF export integration to match the latest llama.cpp converter architecture.
**Why**
llama.cpp moved model-specific GGUF conversion logic out of convert_hf_to_gguf.py and into the new conversion/ package. AutoRound previously depended on symbols from convert_hf_to_gguf.py, which is now only a CLI wrapper and no longer sufficient for GGUF export.

**What Changed**
- Added a bundled llama.cpp conversion/ snapshot for out-of-the-box GGUF export.
- Added llama_cpp_conversion.py as the single adapter for GGUF conversion loading.
- Removed AutoRound’s dependency on convert_hf_to_gguf.py.
- Added support for LLAMA_CPP_ROOT to use a local llama.cpp checkout.
- Added optional AUTO_ROUND_GGUF_AUTO_UPDATE=1 support to dynamically download the minimal required conversion files when a model is unsupported by the bundled snapshot.
- Added a sync script to refresh the bundled conversion/ snapshot from a pinned llama.cpp commit.
- Excluded bundled upstream conversion files from formatting/lint/autofix hooks to keep them close to upstream.

**New Behavior**
By default, users can continue using GGUF export without extra setup. For newer models not supported by the bundled converter, users can either set LLAMA_CPP_ROOT or enable AUTO_ROUND_GGUF_AUTO_UPDATE=1 to try the latest llama.cpp conversion logic dynamically.

## Type of Change

Bug fix

## Related Issues

Fixes or relates to #

## Checklist Before Submitting

- [ ] My code has been tested locally.
- [ ] Documentation has been updated as needed.
- [ ] New or updated tests are included where applicable.
- [ ] The CUDA CI has passed. You can trigger it by commenting `/azp run Unit-Test-CUDA-AutoRound`.

PR #1374, opened 2026-01-30
Support general MOE replacement for MOE models (Transformers 5.0 compatible)

lvliang-intel:
## Description

Design Overview
This feature provides a generic, model-agnostic mechanism to enable quantization of Mixture-of-Experts (MOE) models by leveraging transformers 5.0+'s native MOE integration interface (@use_experts_implementation decorator and [ALL_EXPERTS_FUNCTIONS registry).

Core Problem
MOE models (DeepSeek-V2, Llama4, Qwen-MoE, Mixtral, etc.) use fused 3D parameter tensors for expert weights in transformers 5.0.

gate_up_proj: (num_experts, 2*intermediate_dim, hidden_dim)
down_proj: (num_experts, hidden_dim, intermediate_dim)

These fused tensors cannot be directly quantized as nn.Linear layers, which auto-round expects.

Solution Architecture
Transformers 5.0+ MOE Interface: ALL_EXPERTS_FUNCTIONS registry , "linear_loop" (our custom implementation)
auto_round/modelling/moe_experts_impl.py:
linear_loop_experts_forward()
- Loops over experts using nn.Linear layers
- Supports quantized nn.Linear (NVFP4, INT4, etc.)
_unfuse_experts_weights_inplace()
- Converts 3D Parameters → nn.ModuleList[nn.Linear]
- Handles transposed vs non-transposed layouts
prepare_model_for_moe_quantization()
- Main entry point
- Registers linear_loop, unfuses all experts modules

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Other (please specify):

## Related Issues

[](https://github.com/intel/auto-round/issues/1345)

Fixes or relates to #

## Checklist Before Submitting

- [x] My code has been tested locally.
- [ ] Documentation has been updated as needed.
- [x] New or updated tests are included where applicable.

PR #2097, opened 2026-07-28
Fix AutoRound gradients for Kimi-VL routed experts

lesj0610:
## Purpose

`moonshotai/Kimi-VL-A3B-Thinking-2506` is listed as a supported model, but its routed experts cannot be gradient-tuned by SignRound when `iters > 0`. The model remains quantizable with RTN, while iterative `W4A16_MIXED` tuning silently leaves the routed-expert tuning parameters without gradients (or fails at backward when no other differentiable path is present). This matters because the mixed recipe intentionally assigns INT4 only to the routed experts.

## Root cause

AutoRound correctly keeps the model in eval mode during block tuning. Kimi-VL's remote `DeepseekV3MoE.forward` therefore selects `moe_infer`, which is decorated with `@torch.no_grad()`. The routed experts execute inside that decorator, so their AutoRound tuning parameters are disconnected from autograd. The shared-expert path is outside `moe_infer`, which can make a block loss require gradients while still leaving every routed expert untuned.

Switching the block to training mode is not valid: Kimi-VL's `noaux_tc` routing path asserts inference mode and training mode would change routing semantics. Existing unfused/fused MoE patches do not cover this class because the top-level model type is `kimi_vl`, and this remote implementation uses a `ModuleList` rather than the Transformers 5 fused-expert interface.

## Design

This change adds an instance-local context manager around `SignRoundQuantizer.quantize_block`:

- It activates only for `model_type == "kimi_vl"` and `iters > 0`.
- It recognizes `DeepseekV3MoE` through the class MRO and temporarily binds the callable exposed by `moe_infer.__wrapped__` to that instance.
- It preserves eval mode and the original routing implementation; it only bypasses the outer `torch.no_grad` wrapper during optimization.
- It validates all targets before mutation, is reentrant, supports multiple targets, and restores the exact previous instance state on normal and exceptional exits.
- It is a no-op for RTN (`iters == 0`), other model types, and unrelated MoE classes.
- It explicitly rejects `ep_size > 1`: the Kimi-VL expert-parallel `all_to_all` path does not preserve routed-expert gradients. Quantize with `ep_size=1` or use `iters=0` (RTN).

`prepare_run()` is validation-only: before block iteration it scans the model to reject unrecognized grad-mode decorators and `ep_size > 1`, then logs the validated target count. It caches neither target descriptors nor runtime state. Every `quantize_block()` invocation—including the normal orchestrator path and the public/LLM-Compressor path—lazily scans only the current block and opens exactly one block-scoped context, so the public path behaves identically without `prepare_run()`.

The runtime patch is scoped to block tuning instead of globally modifying remote model source. It also avoids `.train()` and avoids reimplementing Kimi's routing logic. AdamRound and SignRoundV2 inherit the same `quantize_block` implementation, so they receive the same scoped behavior.

## `W4A16_MIXED` and MLA preservation

A meta-device inventory of the actual local Kimi-VL model applied the current `W4A16_MIXED` recipe with `group_size=32`:

- 5,292 `Linear` modules in total.
- 4,992 INT4 modules, exactly `26 MoE layers × 64 routed experts × 3 projections`; all use group size 32 and integer weight quantization.
- 299 modules remain at 16 bits: 78 shared-expert projections, 108 MLA attention projections, 108 vision linears, 2 multimodal-projector linears, and 3 dense layer-0 MLP projections.
- `language_model.lm_head` is excluded from `layer_config`, and the vision patch-embedding convolution is also untouched.

Safetensors metadata was inspected without loading tensor values. All corresponding checkpoint tensors are BF16: 4,992 routed weights, 78 shared weights, 108 MLA projection weights, 329 vision tensors, 6 projector tensors, and the `lm_head` weight. The patch does not alter this recipe or MLA execution.

## Before-fix reproduction

On upstream `main` at `54b4b84cdfcbd985a00264048a288a224e2a76c0`, a CPU reproducer using Kimi's eval-mode MoE path produced `loss.requires_grad == False`; `backward()` failed with:

```text
element 0 of tensors does not require grad and does not have a grad_fn
```

The remote `moe_infer` method was confirmed to carry the `torch.no_grad` wrapper, while the top-level config reports `model_type="kimi_vl"`.

## Unit and regression tests

Executed locally with CUDA hidden and Hugging Face/Transformers offline mode:

- `test/test_cpu/models/test_kimi_vl.py`: **22 passed**. The tests cover the Kimi-VL `moe_infer` dataflow, real `quantize_block` loss/backward/step behavior, active/inactive expert gradient and update behavior, output parity, eval-mode preservation, normal and exceptional restoration, the context manager's explicit reentrancy contract, lazy current-block scoping with sibling-block exclusion, the public/LLM-Compressor path without `prepare_run()`, `prepare_run()` rejection of unrecognized decorators and `ep_size > 1` before block iteration, multiple targets, exact-MRO inclusion and exclusion, the original scatter-reduce failure mode, inherited quantizers, and the mixed recipe.
- `test/test_cpu/models/test_moe_experts_interface.py`: **7 passed**.
- `test/test_cpu/models/test_vlm_ram_reduction.py`: **14 passed**.
- `pre-commit run --all-files` passed every configured hook, including Black, isort, Ruff, Bandit, codespell, and license checks.
- Python compilation and `git diff --check` passed.

## Actual Kimi GPU smoke test

The first actual MoE decoder block (`language_model.model.layers.1`) was tested on one RTX 3090 24 GB with PyTorch 2.11.0+cu130, Transformers 5.10.2, and the editable AutoRound checkout.

Settings: `W4A16_MIXED`, weight INT4, group size 32, activation 16-bit, `iters=1`, sequence length 32, one sample, batch size 1, seed 42, `quant_nontext_module=False`. The actual remote `DeepseekV3MoE` implementation and 197 block tensors were strict-loaded from the local checkpoint (571,080,768 parameters). No model files were modified.

Results:

- 49 routed experts were active and 15 were inactive only because this deliberately tiny single 32-token smoke sample did not cover all 64 experts; the inactive-expert count does not apply to the actual calibration configuration.
- The patched output exactly matched the decorated inference output (`max_abs_diff=0`) and the model remained in eval mode.
- The lazy `quantize_block()` path scanned the current block, found one `DeepseekV3MoE` target, and recorded one active patch at context depth 1.
- Quantization loss required gradients; loss was `0.007146065589040518`.
- All 441 active tuning parameters had non-`None`, finite, nonzero gradients, and all 441 update probes changed.
- All 135 inactive tuning parameters had `grad is None`, and all 135 full tensors remained unchanged.
- 192 routed-expert linears were quantized; all 3 shared-expert linears remained unquantized.
- Peak CUDA allocated memory: 7,578,559,488 bytes (about 7.06 GiB).
- Peak CUDA reserved memory: 7,604,273,152 bytes (about 7.08 GiB).
- Block quantization time: 2.840883 seconds; internal smoke run: 11.762815 seconds; external wall time: 16.80 seconds.
- External maximum resident set size: 4,491,320 KiB.
- The original config, remote-code, index, and checkpoint files retained their pre-test fingerprints.

Transformers 5.10.2 no longer exposes `transformers.utils.import_utils.is_torch_fx_available`, which the Transformers-4-era remote code imports. The smoke harness added an in-process compatibility alias returning `True`; neither AutoRound nor the checkpoint files were modified.

A separate full-model meta instantiation verified all module names and recipe assignments in 13.09 seconds with 2,828,012 KiB maximum RSS. It did not load full weights or run full-model quantization.

## Relationship to prior work

- #518 documents the initial Kimi-VL support surface but does not address routed-expert autograd.
- #1326 added the mixed scheme coverage, but its relevant path uses `iters=0`, so it cannot detect missing routed-expert gradients.
- #1367 patches the built-in `deepseek_v3` model type; Kimi-VL reports `kimi_vl`, so that patch does not apply.
- #1374 covers the Transformers 5 fused-expert interface; Kimi-VL's remote MoE uses a `ModuleList` and has no matching fused-expert interface/decorator.
- #1532 enables `trust_remote_code` while checking unfused MoE configurations, which is a prerequisite but not a Kimi-VL autograd fix.
- #1642 supports pre-quantized WOQ Kimi K2.5-class inputs, a different flow.
- #1846 adds Kimi/MLA-aware GGUF serialization, not block-tuning gradients.
- #1971 improves AutoScheme MoE scoring and inactive-expert score handling with RTN-style scoring; it does not restore SignRound gradients through Kimi's decorated inference method.

## Validation not completed

A complete CPU test suite was not practical in this environment:

- A related combined CPU selection reached 26 passing tests and then failed in `test/test_cpu/utils/test_alg_ext.py::TestAlgExt::test_alg_ext` because the installed `gguf` package lacks `MODEL_ARCH.DFLASH`; bundled llama.cpp conversion then reported `Qwen3ForCausalLM` unsupported. This path does not import or execute the changed Kimi/SignRound code.
- The existing remote-model `test_w4a16_mixed_mllm` fixture stopped making progress while fetching Qwen2.5-VL data. It was interrupted after 19:06; the preceding selected tests reported `8 passed, 26 deselected`. No generated model fixture remains in the worktree.

Full Kimi-VL checkpoint quantization, export/reload, vLLM/Marlin loading, end-to-end multimodal generation with real image inputs, and quality benchmarks were not run. No claim is made here about accuracy, reasoning quality, hallucination rate, or end-to-end compression quality. Maintainers should separately validate this patch with a Transformers combination supported by the repository and run a representative Kimi-VL calibration/evaluation/export cycle before declaring quality parity.

## Artifacts and remaining checks

This PR contains only two source files and one CPU regression-test file. It contains no model weights, quantized outputs, caches, logs, or generated artifacts.

The context is designed for AutoRound's serial block-tuning flow and is reentrant. Simultaneous multi-thread tuning of the same module instance was not validated. Maintainers should also confirm the remote-code compatibility matrix for the repository's supported Transformers versions and run full-model quality/export validation on the target checkpoint.

### DIFF
--- a/auto_round/algorithms/quantization/sign_round/quantizer.py
+++ b/auto_round/algorithms/quantization/sign_round/quantizer.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import copy
 from contextlib import nullcontext
+from functools import wraps
 from typing import TYPE_CHECKING, Any, Callable, Optional, Union
 
 import torch
@@ -27,6 +28,7 @@ from auto_round.compressors.utils import (
     collect_best_params,
 )
 from auto_round.logger import logger
+from auto_round.modeling.kimi_vl import enable_kimi_vl_moe_grad, prepare_kimi_vl_moe_grad
 from auto_round.utils import (
     htcore,
     is_hpex_available,
@@ -39,7 +41,18 @@ from auto_round.utils.distributed import setup_ddp_if_needed_
 from auto_round.wrapper import WrapperLinear, unwrapper_block, unwrapper_layer, wrapper_block
 
 if TYPE_CHECKING:
-    from auto_round.algorithms.composer import BlockContext
+    from auto_round.algorithms.composer import AlgorithmComposer, BlockContext
+
+
+def _with_kimi_vl_moe_grad(func: Callable) -> Callable:
+    """Scope the Kimi-VL routed-expert gradient patch to block tuning."""
+
+    @wraps(func)
+    def wrapper(self, block, *args, **kwargs):
+        with enable_kimi_vl_moe_grad(self.model, block, self.iters):
+            return func(self, block, *args, **kwargs)
+
+    return wrapper
 
 
 @register_pipeline_member(SignRoundConfig)
@@ -63,6 +76,17 @@ class SignRoundQuantizer(BaseQuantizer):
 
         self.optimizer = self._get_optimizer(optimizer=config.optimizer)
         self.wrapper_block = wrapper_block
+
+    def prepare_run(self, composer: "AlgorithmComposer" = None) -> None:
+        """Pre-validate Kimi-VL routed experts before block iteration."""
+        super().prepare_run(composer=composer)
+        target_count = prepare_kimi_vl_moe_grad(self.model, self.iters)
+        if target_count is not None:
+            logger.info(
+                "Kimi-VL routed-expert gradient support is enabled for %d DeepseekV3MoE module(s) "
+                "during block tuning.",
+                target_count,
+            )
 
     def dispatch_block(self, block, input_ids, input_others):
         """Multi-GPU aware block dispatch for SignRound tuning.
@@ -149,6 +173,7 @@ class SignRoundQuantizer(BaseQuantizer):
         else:
             return inputs.shape[self.calibration_context.batch_dim]
 
+    @_with_kimi_vl_moe_grad
     def quantize_block(
         self,
         block,
--- a/auto_round/modeling/kimi_vl.py
+++ b/auto_round/modeling/kimi_vl.py
@@ -1,0 +1,191 @@
+# Copyright (c) 2026 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from contextlib import contextmanager
+from dataclasses import dataclass
+from types import MethodType
+from typing import Iterable, Iterator
+
+import torch
+
+_MISSING = object()
+_STATE_ATTR = "_auto_round_kimi_vl_moe_grad_state"
+
+
+@dataclass(frozen=True)
+class _KimiVLMoEGradTarget:
+    module_name: str
+    module: torch.nn.Module
+
+
+class _KimiVLMoEGradState:
+    def __init__(self, previous_method: object, previous_state: object):
+        self.previous_method = previous_method
+        self.previous_state = previous_state
+        self.depth = 1
+
+
+def _is_kimi_vl_moe_grad_enabled(model: torch.nn.Module | None, iters: int) -> bool:
+    config = getattr(model, "config", None)
+    return iters > 0 and getattr(config, "model_type", None) == "kimi_vl"
+
+
+def _find_kimi_vl_moe_grad_targets(root: torch.nn.Module) -> tuple[_KimiVLMoEGradTarget, ...]:
+    targets = []
+    for module_name, module in root.named_modules():
+        if not any(base.__name__ == "DeepseekV3MoE" for base in type(module).__mro__):
+            continue
+        targets.append(_KimiVLMoEGradTarget(module_name=module_name, module=module))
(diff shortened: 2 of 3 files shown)


wenhuach21:
@chensuyue  We need someone to take a look at this issue, as this PR is currently inactive. https://github.com/intel/auto-round/issues/2282

xin3he:
Hi @lesj0610 Thanks for this contribution, this change is too hacky.

It's recommanded to provide a hook interface at the model level.
Expose a "differentiable inference" interface from the model itself:
```
# In the Kimi-VL modeling
class DeepseekV3MoE(nn.Module):
    def forward_with_grad(self, hidden_states, ...):
        """support gradient"""
        return self._moe_infer_impl(hidden_states, ...)
    
    def forward(self, hidden_states, ...):
        if torch.is_grad_enabled():
            return self.forward_with_grad(hidden_states, ...)
        else:
            with torch.no_grad():
                return self._moe_infer_impl(hidden_states, ...)
```

If you prefer to continue this work, please let me know recently~

PR #2103, opened 2026-07-30
fix: skip exporting fp8 attention q_max

yiliu30:
## Summary
- skip serializing the transient FP8 attention `q_max` calibration tensor during immediate shard writes
- keep `q_scale`, `k_scale`, and `v_scale` in exported checkpoints
- add export regression checks for both `llm_compressor` and `auto_round` formats

## Root cause
`clean_up_hooked_attention()` removes `q_max` from the in-memory model, but immediate saving writes shard files before that cleanup runs. As a result, `q_max` leaked into exported safetensors even though it is calibration-only.

## Validation
- manual repro with `nm-testing/tinysmokeqwen3` + `static_attention_dtype=fp8` + `format=llm_compressor`
  - `qmax_count = 0`
  - `q_scale_count = 6`
  - `k_scale_count = 6`
  - `v_scale_count = 6`
- manual repro with the same model + `format=auto_round`
  - `qmax_count = 0`
  - `q_scale_count = 6`
  - `k_scale_count = 6`
  - `v_scale_count = 6`


chensuyue (approved):
XPU test issue not related to this PR.

Issue #1986, opened 2026-07-02
[Feature]: adapt to code calibration dataset for model name with code


wenhuach21:
### Feature Description

~

### Motivation and Use Case

~

changwangss:
pr merged, add another task, verify the accuracy qwen3.6 27B and Gemma4-31B  on some code agent relatives task with different ratio between opencode and github-clean-code  datasets.

PR #2107, opened 2026-07-30
Automatically select calibration datasets for code models

changwangss:
## Description

Automatically select code-oriented calibration datasets when quantizing a pure-text code model without an explicitly configured calibration dataset.

- Add `is_code_model()` following the existing model-detection style.
  - Detect code-specialized models from the model name and existing config metadata without additional Hub requests.
  - Avoid false positives such as `encoder`, `decoder`, `codec`, and unrelated parent directories named `code`.

- Select an exact-size code calibration dataset:
  - `datasets <= 3.6.0`: OpenCodeInstruct 50% and GitHub Code Clean 50%.
  - `datasets > 3.6.0`: OpenCodeInstruct only, because script-based GitHub Code Clean is no longer supported.
  - Enable cross-sample concatenation for OpenCodeInstruct so short examples can form fixed-length calibration sequences.

- Apply automatic selection only to pure-text LLMs that require calibration.

- Preserve explicit user dataset choices, including both `NeelNanda/pile-10k` and its `pile-10k` alias.
  - If `dataset` / `--dataset` is omitted, normal models continue to use Pile-10k, while detected code models use the code-oriented calibration dataset.
  - If the user explicitly specifies any dataset, that choice is left unchanged.

- Keep the existing diffusion default behavior (`coco2014`) unchanged.

## Type of Change

New feature

## Related Issues

Fixes or relates to #https://github.com/intel/auto-round/issues/1986


auto_round/utils/model.py
```diff
@@ -1079,6 +1079,86 @@ def get_model_name_or_path(model_or_path: Union[str, torch.nn.Module]) -> Option
...
 
 
+_CODE_MODEL_TOKENS = {"code", "coder", "coding", "programming", "swe", "devstral"}
+_CODE_MODEL_FAMILIES = {
+    "codellama",
+    "codegemma",
+    "codestral",
+    "deepseekcoder",
+    "granitecode",
+    "magicoder",
```
wenhuach21, line 1089: Is it still doable if the user wants to use Pile-10k for a code model?

changwangss: fixed the issue. 
added a `dataset_was_explicitly_set` flag, changed the dataset default to None across all entry points, and then selected the appropriate default dataset only after determining whether the user had explicitly specified one.

auto_round/compressors/base.py
```diff
@@ -210,7 +210,7 @@ def __init__(
         ignore_layers: str = "",
         quant_lm_head: bool = False,
         to_quant_block_names: Optional[Union[str, list[str]]] = None,
-        dataset: Union[str, list, tuple, torch.utils.data.DataLoader] = "NeelNanda/pile-10k",
+        dataset: Optional[Union[str, list, tuple, torch.utils.data.DataLoader]] = None,
```
wenhuach21, line 213: I think the API entry of AutoRound is not this one, so you need to change auto_round/autoround.py /AutoRound  as well
@n1ck-guo  is it possible to merge your pr ASAP,  the different entries confuse developers

auto_round/compressors/base.py
```diff
@@ -220,9 +220,8 @@ def __init__(
         # instance onto the quantizer so the two share state.
         from auto_round.calibration.state import CalibrationContext
 
-        self.dataset = dataset
-        if self.dataset is None:
-            self.dataset = "NeelNanda/pile-10k"
+        self.dataset_was_explicitly_set = dataset is not None
```
wenhuach21: based on the code, there is no need to save dataset_was_explicitly_set  to self

auto_round/calib_dataset.py
```diff
@@ -23,12 +23,40 @@
...
+
+
+def get_code_calibration_dataset(nsamples, datasets_version=None):
+    """Build an exact-size code calibration mix compatible with datasets."""
+    if datasets_version is None:
+        import datasets
+
+        datasets_version = datasets.__version__
+    parsed_version = Version(str(datasets_version))
+    sources = [("opencode-instruct:concat=true", 50), ("github-code-clean", 40), ("mbpp:split=train:concat=true", 10)]
```
wenhuach21: is there any data to support this?

auto_round/calib_dataset.py
```diff
@@ -23,12 +23,40 @@
...
+    if parsed_version > _GITHUB_CODE_CLEAN_MAX_DATASETS_VERSION:
+        sources = [source for source in sources if source[0] != "github-code-clean"]
+        logger.warning(
```
wenhuach21: warning_once

auto_round/compressors/base.py
changwangss: Fixed. `dataset_was_explicitly_set` is now a local variable because it is only needed during initialization and does not need to be stored on `self`.

auto_round/calib_dataset.py
changwangss: Fixed. Changed `logger.warning()` to `logger.warning_once()` to avoid repeatedly reporting the same datasets-version compatibility warning when multiple AutoRound instances are created.

auto_round/calib_dataset.py
changwangss: Updated based on our discussion. MBPP has been removed from the automatically selected mixture. For `datasets <= 3.6.0`, OpenCodeInstruct and GitHub Code Clean are now selected with a 50/50 ratio. For newer `datasets` versions, the code falls back to OpenCodeInstruct only because script-based GitHub Code Clean is no longer supported.

auto_round/compressors/base.py
changwangss: Already addressed in commit 9961d8ee. The public `AutoRound` API in `auto_round/autoround.py` now defaults `dataset` to `None` and forwards it unchanged, allowing the downstream orchestrator to distinguish an omitted dataset from an explicit user choice.

n1ck-guo: There is no conflict， this is the correct modification.

n1ck-guo (approved):
LGTM

PR #2114, opened 2026-07-31
refine eval, add more option

WeiweiZhang1:
## Description

add few-shot, fewshot_as_multiturn option for eval API

## Type of Change

Bug fix

## Related Issues

Fixes or relates to #

### DIFF
--- a/auto_round/eval/eval_cli.py
+++ b/auto_round/eval/eval_cli.py
@@ -102,7 +102,9 @@ class EvalArgumentParser(argparse.ArgumentParser):
             "Float between 0-1: fraction of total examples.",
         )
         self.add_argument("--num_fewshot", "--num-fewshot", default=None, type=int, help="Number of few-shot examples.")
-        self.add_argument("--eval_gen_kwargs", "--eval-gen-kwargs", default=None, type=str, help="Generation kwargs for LM-Eval.")
+        self.add_argument(
+            "--eval_gen_kwargs", "--eval-gen-kwargs", default=None, type=str, help="Generation kwargs for LM-Eval."
+        )
         self.add_argument(
             "--eval_backend",
             default="hf",
--- a/auto_round/eval/eval_cli.py
+++ b/auto_round/eval/eval_cli.py
@@ -16,7 +16,7 @@ import importlib.util
 import os
 import time
 
-import torch.nn
+import torch
 from transformers.utils.versions import require_version
 
 from auto_round.utils import (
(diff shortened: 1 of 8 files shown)


n1ck-guo (approved):
LGTM. 
I'm considering adding a parameter to `eval`—something like `--eval_args "limit=xxx"`—and then passing that parameter directly to `simple_evaluate` after processing it; this would avoid the need to constantly add new arguments. However, we can discuss later whether this is necessary.

auto_round/eval/eval_cli.py
```diff
@@ -487,18 +547,33 @@ def _evaluate_tasks_with_retry(tasks, hflm, device_str, batch_size, limit, retry
...
+                    tasks=task,
+                    batch_size=batch_size,
+                    limit=limit,
+                    num_fewshot=num_fewshot,
+                    gen_kwargs=gen_kwargs,
+                    task_manager=task_manager,
+                    fewshot_as_multiturn=fewshot_as_multiturn,
                 )
                 break
             except Exception as e:
```
wenhuach21: [catch ](https://github.com/intel/auto-round/blob/f17d9cd4b36982006bad21ff87127aac739072e3/auto_round/algorithms/quantization/base.py#L126)

auto_round/eval/eval_cli.py
```diff
@@ -459,7 +491,35 @@ def _load_gguf_model_if_needed(model_path, eval_model_dtype=None):
     return model, tokenizer, is_gguf_file, gguf_file
 
 
-def _evaluate_tasks_with_retry(tasks, hflm, device_str, batch_size, limit, retry_times):
+def _get_lm_eval_task_manager(tasks):
```
wenhuach21: type annotation

auto_round/eval/eval_cli.py
```diff
@@ -459,7 +491,35 @@ def _load_gguf_model_if_needed(model_path, eval_model_dtype=None):
...
+
+        tasks_root = os.path.join(os.path.dirname(lm_eval.__file__), "tasks")
+        task_paths = []
+        for task in tasks:
+            task_path = os.path.join(tasks_root, task)
+            if not os.path.isdir(task_path):
+                return None
+            task_paths.append(task_path)
+        return TaskManager(include_defaults=False, include_path=task_paths)
+    except Exception:
```
wenhuach21: Where does this exception come from? It would be better to restrict the scope of the `try` block to only the code that can actually raise this exception.

Issue #2146, opened 2026-08-11
CPU CI AttributeError: 'QuantLinearGPTQ' object has no attribute 'weight'. Did you mean: 'qweight'?


XuehaoSun:
```
2026-08-11T02:15:06.1559561Z ______________________ TestQuantizationConv1d.test_quant _______________________
2026-08-11T02:15:06.1560222Z unit/test_cpu/models/test_conv1d.py:43: in test_quant
2026-08-11T02:15:06.1561528Z     model = AutoModelForCausalLM.from_pretrained(quantized_model_path, device_map="cpu", trust_remote_code=True)
2026-08-11T02:15:06.1563029Z             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-08-11T02:15:06.1563535Z ../auto_round/utils/common.py:140: in patched
2026-08-11T02:15:06.1564093Z     return underlying_func(klass, *args, **kwargs)
2026-08-11T02:15:06.1564560Z            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-08-11T02:15:06.1565157Z /home/hostuser/.venv/lib/python3.12/site-packages/transformers/models/auto/auto_factory.py:402: in from_pretrained
2026-08-11T02:15:06.1565782Z     return model_class.from_pretrained(
2026-08-11T02:15:06.1566429Z /home/hostuser/.venv/lib/python3.12/site-packages/transformers/modeling_utils.py:4349: in from_pretrained
2026-08-11T02:15:06.1567144Z     loading_info = cls._finalize_model_loading(model, load_config, loading_info)
2026-08-11T02:15:06.1567658Z                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-08-11T02:15:06.1568237Z /home/hostuser/.venv/lib/python3.12/site-packages/transformers/modeling_utils.py:4517: in _finalize_model_loading
2026-08-11T02:15:06.1568937Z     model._initialize_missing_keys(load_config.is_quantized)
2026-08-11T02:15:06.1569611Z /home/hostuser/.venv/lib/python3.12/site-packages/transformers/modeling_utils.py:4808: in _initialize_missing_keys
[...]

PR #2150, opened 2026-08-12
fix: restrict transformers version to <5.15.0 in requirements files

chensuyue:
## Description

Temp fix for https://github.com/intel/auto-round/issues/2146, need to be reverted in https://github.com/intel/auto-round/pull/2147.


chensuyue:
After reset transformers version, still got a diffusion issue.

PR #2193, opened 2026-08-19
Reduce SYCL-TLA Compiler Memory

luoyu-intel:
# Reduce SYCL-TLA Compiler Memory

## Description

This PR reduces peak compiler memory for SYCL-TLA translation units under oneAPI 2026.0. The change moves concrete dense GEMM, S8 GEMM, MoE, and SDPA template instantiations out of shared wrapper translation units and into CMake-generated implementation sources.

The public Python-facing APIs remain unchanged. Runtime routing is preserved through lightweight declaration-only interfaces, while concrete template definitions are compiled only by the generated source families that need them.

## Motivation

The previous include and instantiation structure caused large SYCL-TLA template graphs to be parsed and instantiated repeatedly. Representative translation units reached approximately 4 GiB peak RSS, making parallel builds unreliable on constrained build hosts.

The target is to keep representative per-translation-unit compiler memory below approximately 2 GiB while retaining the existing kernel routes and runtime behavior.

## Main Changes

- Generate dedicated SYCL-TLA implementation sources for:
  - Dense GEMM fixed-tile routes and dtype dispatch.
  - S8 GEMM normal and K-block routes.
  - MoE grouped GEMM, decode, and prefill routes.
  - Specialized INT8, FP8, low-bit, tensor, and dequantized MoE prefill paths.
- Add declaration-only common and helper headers so `ark.cpp` and lightweight wrappers do not include the full implementation headers.
- Split SDPA instantiations by dtype, head dimension, causal state, mask state, and varlen state.
- Preserve runtime validation and dispatch behavior for dense GEMM, S8 GEMM, grouped GEMM, decode, prefill, and SDPA.
- Add a direct packed symmetric S4 MoE prefill DPAS path, gated by `ARK_MOE_PREFILL_DPAS_S4`, with fallback to the existing routes when disabled or unsupported.
- Change the default SYCL-TLA compiler-memory scheduling policy to approximately 3 GiB per compiler job.
- Reject explicitly set empty, zero, negative, and non-numeric `ARK_SYCL_TLA_JOBS` values.
- Add static template-analysis and serial compiler-memory measurement tools.
- Document the build-memory controls and measurement workflow in English and Chinese.

## Validation

- Configured oneAPI XPU build completed successfully with `cmake --build auto_round_extension/ark/xbuild -j 2`.
- Static analysis completed across 193 translation units.
- The active compile database contains 72 generated SDPA translation units with unique causal, mask, and varlen variants.
- Measured representative peak RSS:
  - Worst measured SDPA sources: approximately 1,941-1,950 MiB.
  - Direct packed-S4 source: approximately 1,516.6 MiB.
- Python syntax checks passed for `setup.py` and both new analysis tools.
- Both new analysis tools passed their command-line help checks.
- The rebuilt shared library has no matching unresolved ARK API or generated-kernel symbols.
- `git diff --check` passed.

## Runtime Validation Limitation

The current validation host has no XPU device (`torch.xpu.is_available() == False`, device count `0`). The focused MoE prefill accuracy suite therefore skipped all 36 cases. Direct packed-S4 numerical validation, including nibble ordering, `cutlass::int4b_t` interpretation, DPAS results, and scale folding, still needs to be run on a BMG/XPU host with both settings:

```bash
ARK_MOE_PREFILL_DPAS_S4=1
ARK_MOE_PREFILL_DPAS_S4=0
```

The existing fallback path remains available for unsupported shapes and for environments where the direct S4 route is disabled.

## Related Issues

Not specified.

## Checklist

- [x] Build and static validation completed locally.
- [x] Documentation updated, including the Chinese counterpart.
- [x] New compiler-analysis tools added and checked.
- [ ] XPU runtime accuracy validation completed.
- [ ] XPU direct-S4 performance comparison completed.
- [ ] CUDA CI status checked. This PR targets the ARK SYCL-TLA/XPU backend.


a32543254 (approved):
LGTM

Zhenzhong1 (approved):

Issue #2185, opened 2026-08-18
[Bug]: auto-scheme wrongly changes  dataset when oom


wenhuach21:
### Problem Description

Generating AutoScheme:   9%|████████████████████                                                                                                                                                                                                   | 383/4096 [03:52<1:15:07,  1.21s/it][W818 19:47:57.206225105 CUDACachingAllocator.cpp:3933] memory allocation failed with OOM on device 0 while trying to allocate 8136949760 bytes (free: 7246446592, total: 85094825984).
[W818 19:47:57.305789898 CUDACachingAllocator.cpp:3933] memory allocation failed with OOM on device 0 while trying to allocate 8136949760 bytes (free: 7252738048, total: 85094825984).
2026-08-18 19:47:57 WARNING delta_loss.py L1133: Text-only calibration failed on VLM (CUDA out of memory. Tried to allocate 7.58 GiB. GPU 0 has a total capacity of 79.25 GiB of which 6.75 GiB is free. Process 230003 has 6.11 GiB memory in use. Process 230359 has 33.18 GiB memory in use. Including non-PyTorch memory, this process has 33.19 GiB memory in use. Of the allocated memory 24.51 GiB is allocated by PyTorch, and 8.13 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://docs.pytorch.org/docs/stable/notes/cuda.html#optimizing-memory-usage-with-pytorch-cuda-alloc-conf)); falling back to multimodal calibration dataloader.
Generating AutoScheme:   9%|████████████████████▏                                                                                                                                                                                                  | 384/4096 [03:53<1:14:02,  1.20s/it]2026-08-18 19:47:57 WARNING logging.py L340: AutoScheme(force_mllm): dataset='NeelNanda/pile-10k' is text-only, overriding to 'liuhaotian/llava_conv_58k'.
2026-08-18 19:47:57 WARNING logging.py L340: Unable to recognize qwen3_5, using default template instead.
2026-08-18 19:47:57 WARNING dataset.py L250: seqlen(1024) is greater than the maximum length supported by the liuhaotian/llava_conv_58k, reset to 512

### Reproduction Steps

~

### Error Logs

```shell

```

xin3he:
@wenhuach21 This issue is fixed, right?

wenhuach21:
I dont think so

PR #2220, opened 2026-08-24
fix: AutoScheme parallel scoring OOMs (MoE memory, VLM gate) and disk-stream tuning crash

avtc:
## Description

AutoScheme scoring with parallel workers OOMed on large models in two ways, and disk-streamed tuning crashed at the first MoE block (found on Qwen3.8-27B and tencent/hy3, a 300B A21B MoE; 24 GB GPUs, 4 workers, one per GPU):

* the parallel-scoring gate blocked text-only VLM runs, forcing Qwen3.8-27B's scoring into a single process that OOMed - language-tower scoring now streams blocks in workers like any text model
* the block-replay backward kept every block's autograd graph alive (retain_graph=True) and materialized weight-shaped gradient buffers through the in-graph quant chain - tens of GiB on expert blocks. Graphs are now freed per block, each layer is quantized once per pass, and the scoring forward recomputes from a CPU-bounded cache; scores are unchanged, quant searches drop from three per layer per batch visit to one
* a failed worker no longer aborts its siblings: per-scheme results persist, missing schemes raise a clear rerun hint (opt-in fail-fast via new `AR_AUTO_SCHEME_NO_SERIAL_FALLBACK`), and workers resume from batch-granularity partial checkpoints
* streamed block inputs are aligned to the block's device before the forward, and the offloader resolves checkpoint-conversion renames when materializing meta-skeleton blocks (previously crashed with "Cannot copy out of meta tensor")
* worker CUDA OOMs carry a self-diagnosing annotation: live-tensor census by (shape, dtype), retaining containers, and the failing op's traceback

The four-scheme scoring pass on tencent/hy3 that previously OOMed every worker at ~3% progress now completes with ~10 GB peak VRAM per worker.

## Type of Change

Bug fix

## Related Issues

Fixes the OOM that triggers the wrong-dataset fallback in #2185.
Relates to #1913 — text-only VLM scoring can now combine parallel workers with disk streaming to keep host RAM flat, and the new tensor census aids diagnosing the serial path.

- [x] The CUDA CI has passed. You can trigger it by commenting `/azp run Unit-Test-CUDA-AutoRound`.

---
Edit:

Exact commands to run auto-scheme i have tested with for hy3:
```
export CUDA_VISIBLE_DEVICES=0,1,2,3
export AR_SCHEME_MEM_INVENTORY=1
export AR_ENABLE_AUTO_SCHEME_PARALLEL=1
export AR_AUTO_SCHEME_BATCH_SIZE=16
export AR_DISK_STREAM_MODEL=1
export AR_AUTO_SCHEME_CACHE=~/as_cache
export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
export AR_AUTO_SCHEME_NSAMPLES=16
export AR_AUTO_SCHEME_SEQLEN=128
export AR_AUTO_SCHEME_NO_SERIAL_FALLBACK=1
export LD_LIBRARY_PATH=$(python -c "import torch, os; print(os.path.join(os.path.dirname(torch.__file__), 'lib'))"):$(python -c "import nvidia.cudnn; print(nvidia.cudnn.__path__[0])")/lib:$LD_LIBRARY_PATH
python -m auto_round --model /home/ubuntu/models/Hy3-tencent --enable_alg_ext \
    --options "W3A16,W4A16,W5A16,W8A16" --avg_bits 3.5 \
    --iters 0 --device_map 0,1,2,3 \
    --nsamples 128 --lr 2e-3 \
    --shared_layers "q_proj,k_proj,v_proj" \
                 "shared_experts.gate_proj,shared_experts.up_proj" \
                 "mlp.gate_proj,mlp.up_proj" \
    --enable_torch_compile \
    --low_gpu_mem_usage \
    --enable_opt_rtn \
    --format  auto_round:llm_compressor \
    --layer_config "{ '.*model.layers.0.':{'bits':16,'data_type':'float'}, '.*model.layers.80.':{'bits':8}, '.*shared_experts':{'group_size':64}, 'lm_head':{'bits':8} }" \
    --output_dir /home/ubuntu/models/autoround/tencent-Hy3-3.5bpw-iters0
```


wenhuach21:
@xin3he  if this pr could not catch up v0.15， could we disable parallel scoring in v0.15
cc @chensuyue

auto_round/utils/disk_stream_util.py
```diff
@@ -76,6 +77,65 @@ def tensor_names_with_prefix(self, prefix: str) -> list[str]:
         return [n for n in self.weight_map if n == prefix or n.startswith(dotted)]
 
 
+# Model-side module names sometimes differ from the checkpoint-side tensor
```
xin3he, line 80: Good point.

xin3he (approved):
LGTM

xin3he:
> @xin3he if this pr could not catch up v0.15， could we disable parallel scoring in v0.15 cc @chensuyue

@wenhuach21  If I understand correctly, the issue here is that AR_DISK_STREAM_MODEL is incompatible with VLM, and is unrelated to parallel processing. cc @avtc 
`Model-side module names sometimes differ from the checkpoint-side tensor`

avtc:
> @wenhuach21 If I understand correctly, the issue here is that AR_DISK_STREAM_MODEL is incompatible with VLM, and is unrelated to parallel processing. cc @avtc `Model-side module names sometimes differ from the checkpoint-side tensor`

@xin3he  Two different failure modes are getting conflated here. The quoted line ("model-side module names…") is from the disk-stream offloader fix: when AR_DISK_STREAM_MODEL=1 materializes meta blocks, checkpoint tensor names can diverge from module names (e.g. a router stored as mlp.router.gate behind a mlp.gate module), which left parameters on meta and crashed tuning (on hy3). That one is model-generic, not VLM-related. The parallel-processing part is the scoring gate: previously (not disk_stream_model or not is_vlm) disabled parallel scoring for any VLM whenever disk streaming was preferred, so text-only VLM scoring collapsed into a single process that OOMed on large models — this PR re-keys the gate to force_mllm, since only actual multimodal scoring needs the full-model backward. For #2185 specifically, the log shows no disk-stream involvement: the text-only calibration OOM came from per-worker memory accumulation (retained replay graphs + weight-grad buffers), which the memory commits here address — removing the OOM that triggers the wrong-dataset fallback.

avtc:
I have noticed a gap for the fix - when schema exists for all W, except of one - it fallbacks to serial path with cuda mismatch device, will test a fix and commit soon

update: fixed with commit below

avtc:
I am investigating failed cuda test

avtc:
@xin3he can you please re-run /azp run Unit-Test-CUDA-AutoRound

avtc:
fyi, all tests passed

wenhuach21:
Great work ! Really appreciate your effort!

PR #2230, opened 2026-08-25
fix: remove stale quantizer access in dual-transformer diffusion

changwangss:
## Description

the command and issue are occurred when quantized wan2.2 to nvfp4 datatype.
```
CUDA_VISIBLE_DEVICES=1 auto-round   --model /data3/changwa1/Wan2.2-T2V-A14B-Diffusers   --scheme NVFP4   --format llm_compressor   --device_map 0     --batch_size 1   --nsamples 64   --seqlen 2048   --iters 200   --num_inference_step 25 --low_gpu_mem_usage   --output_dir /mnt/disk3/changwa1/Wan2.2-T2V-A14B-Diffusers-NVFP4 --model-type bf16
```

```
2026-08-25 17:33:45 INFO device.py L1448: 'peak_ram': 564.85GB, 'peak_vram': 72.64GB
2026-08-25 17:33:45 INFO orchestrator.py L794: quantization tuning time 47823.08122754097
2026-08-25 17:33:45 INFO orchestrator.py L813: Summary: quantized 1/7 in the model, unquantized layers: condition_embedder.text_embedder.linear_1, condition_embedder.text_embedder.linear_2, condition_embedder.time_embedder.linear_1, condition_embedder.time_embedder.linear_2, condition_embedder.time_proj, proj_out
2026-08-25 17:33:45 INFO diffusion_mixin.py L445: Quantizing transformer_2
2026-08-25 17:33:47 INFO composer.py L208: Block-forward torch.compile is disabled because at least one quantized layer uses NVFP4.
Traceback (most recent call last):
  File "/data3/changwa1/.venv/bin/auto-round", line 10, in <module>
    sys.exit(run())
             ~~~^^
  File "/data3/changwa1/auto-round/auto_round/cli/main.py", line 507, in run
    start(argv=command_argv)
    ~~~~~^^^^^^^^^^^^^^^^^^^
  File "/data3/changwa1/auto-round/auto_round/cli/main.py", line 240, in start
    tune(args)
    ~~~~^^^^^^
  File "/data3/changwa1/auto-round/auto_round/cli/main.py", line 404, in tune
    model, folders = autoround.quantize_and_save(args.output_dir, format=args.format)  # pylint: disable=no-member
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data3/changwa1/auto-round/auto_round/compressors/base.py", line 2004, in quantize_and_save
[...]

PR #2236, opened 2026-08-26
fix: remove stale quantizer access in dual-transformer diffusion

changwangss:
## Summary

- Use orchestrator-owned `quant_block_list` and `layer_config` in the dual-transformer diffusion path.
- Avoid the missing `self.quantizer` crash when quantizing `transformer_2`.
- Add a regression test for stale `self.quantizer` access.

This forwards the fix from #2230, which was merged into `v0.15.0rc`, to `main`.

## Validation

- 9 focused tests passed.
- All pre-commit hooks passed.

Related: #2230

### DIFF
--- a/auto_round/compressors/diffusion_mixin.py
+++ b/auto_round/compressors/diffusion_mixin.py
@@ -459,8 +459,8 @@ class DiffusionMixin:
             self.layer_config = {}
 
             # Get new block names for caching
-            if bool(self.quantizer.quant_block_list):
-                all_blocks = self.quantizer.quant_block_list
+            if bool(self.quant_block_list):
+                all_blocks = self.quant_block_list
             else:
                 all_blocks = get_block_names(self.model_context.model)
             if len(all_blocks) == 0:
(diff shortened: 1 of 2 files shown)


changwangss:
@chensuyue @XuehaoSun 
The CUDA CI failure in build 73348 is unrelated to this PR. The diffusion suite affected by this change passed all selected tests (3 passed), while all failures came from `test_fp8_model.py` during Qwen FP8 model loading, before AutoRound's diffusion code was reached.

The failing build resolved `transformers==5.16.0` and crashes in `FineGrainedFP8HfQuantizer.update_tp_plan()`: `FP8Experts._impl_tp_layer_overrides.get(impl)` returns `None`, then 5.16.0 calls `layer_overrides.get(...)` unconditionally. The immediately preceding successful CUDA build 73341 used `transformers==5.15.1`, where the same code was guarded by `if layer_overrides:`, and `test_fp8_model` passed 5/5.

This is an upstream Transformers 5.16.0 regression / floating-dependency CI issue. Re-running without pinning will likely fail again; temporarily pinning Transformers to 5.15.1 or applying the upstream guard should unblock CI.

Build log: https://dev.azure.com/lpot-inc/b7121868-d73a-4794-90c1-23135f974d09/_apis/build/builds/73348/logs/51

XuehaoSun (approved):

PR #2123, opened 2026-08-05
[Experimental] Add NVFP4 E5M3 support and related components for model-free quantization

xin3he:
## Description

- Introduced NVFP4 E5M3 quantization scheme with high-precision QDQ weights.
- Implemented FakeActQuantLinear for runtime activation QDQ.
- Updated model-free compressor to handle NVFP4 E5M3 format.
- Enhanced FakeFormat to support NVFP4 E5M3 serialization.
- Added tests for NVFP4 E5M3 quantization and model conversion.

### Update: 
- support `fake`, `auto_round`, `llm_compressor` format
- support `auto_round:fake`, `auto_round:torch_nvfp4_e5m3`, `auto_round:cute_nvfp4_e5m3` backend

> Real Model Test [NVFP4_E5M3, Llama-3.1-8B-Instruct, lambada_openai] 
    1. BF16: 
     - time: 1min5s
     - acc: 0.7213
    
    2. auto_round format + auto_round:cute_nvfp4_e5m3 (dequant weight, qdq act)
    - time: 1min48s
    - acc: 0.7196
    
    3. auto_round format + auto_round:torch_nvfp4_e5m3 (dequant weight, qdq act)
    - time: 4min23s
    - acc: 0.7204
    
    4. llm_compressor format + auto_round:cute_nvfp4_e5m3 (dequant weight, qdq act)
    - time: 1min48s
    - acc: 0.7196
    
    5. fake format + auto_round:fake (bf16 weight, qdq act)
    - time: 3min 26
    - acc: 0.7204

## Type of Change

Experiment

## Related Issues

Fixes or relates to #

## Checklist Before Submitting

- [ ] My code has been tested locally.
- [ ] Documentation has been updated as needed.
- [ ] New or updated tests are included where applicable.
- [ ] The CUDA CI has passed. You can trigger it by commenting `/azp run Unit-Test-CUDA-AutoRound`.

PR #2245, opened 2026-08-27
Bug fix: Replace WAWrapperLinear with FakeActQuantLinear instead of Linear

xin3he:
## Description

Previous change breaks activation qdq hook https://github.com/intel/auto-round/pull/2123

## Type of Change

Bug fix

## Related Issues

Fixes or relates to #


auto_round/eval/evaluation.py
```diff
@@ -441,8 +448,11 @@ def run_model_evaluation(model, tokenizer, autoround, folders, formats, args):
         eval_folder = folders[-1] if folders else None
     else:
         eval_folder = folders
+
     # set model_type for ModelFreeCompressor.
-    model_type = detect_model_type(eval_folder)
+    # When fake+no-save, detect from the model object to avoid HF Hub access.
+    type_probe = model if (is_fake_format and not envs.AR_SAVE_FAKE_MODEL) else eval_folder
```
wenhuach21: It would be better to always use the online fake model for evaluation when combined with tuning.

wenhuach21: is torch compile supported in get_act_quant_func?

auto_round/export/formats/backends/fake.py
```diff
@@ -58,6 +80,20 @@ def save_quantized(
         serialization_dict: dict = None,
         **kwargs,
     ):
+        if not envs.AR_SAVE_FAKE_MODEL:
+            logger.warning(
```
wenhuach21: info

wenhuach21 (approved):
Thanks for the refinements. This really helps!

auto_round/export/formats/backends/fake.py
xin3he: The default behavior of other formats is saving model, so warning is better.

auto_round/eval/evaluation.py
xin3he: It is.

PR #2273, opened 2026-09-01
fix: asymmetric zero-point packing, format-scoped 8-bit asym, W5/W6/W7 llm_compressor presets

avtc:
## Description

This PR fixes silent-corruption bugs in asymmetric weight packing and scopes the 8-bit-asym policy by servability:

- **llm_compressor export**: two asymmetric-packing bugs, found while quantizing a tencent/hy3 MoE model with asymmetric group search.
(1) Group zero points arrive as float32 tensors but compressed-tensors' packer requires int8, crashing with `Tensor must be quantized to torch.int8 before packing` on the first asym layer.
(2) More subtle: compressed-tensors re-quantizes with a signed integer range and clamps after adding the zero point, while auto_round uses the unsigned level convention - the unsigned zp therefore clamped every level above `2^(b-1)-1`, silently collapsing the upper half of each group's range (such exports load fine but produce garbage). Zero points are now validated against the unsigned level range `[0, 2^bits-1]` (plus finiteness/integrality - NaN passes every min/max comparison), shifted into the signed convention, and range-checked; a round-trip test verifies `(q - zp) * s` is bit-exact through the compressed-tensors pack path.
- **Native/AWQ packers**: rounding overflow could shift bits into the neighboring packed value. Integer weights are clamped to the valid range in `qlinear_torch`, `qlinear_torch_zp` and the AWQ packing path; the classic `(zp-1)` packing clamps `zp = 0` to a decodable stored 0 instead of corrupting the whole packed word (tensor paths, 2/4/8-bit and 3-bit scalar paths, and the model-free GPTQ packer; scalars are normalized to int - a float scalar zp crashed the bit-shift packing).
- **8-bit asym is scoped by what can actually serve it**: native `auto_round` / `auto_gptq` / `auto_awq` / marlin refuse it at configuration time (vLLM's GPTQ-format W8 kernels are symmetric-only and Marlin's zero points are 4-bit only), with uniform `--asym` pinning 8-bit options back to symmetric and explicit `sym: false` on 8-bit entries raising instead of failing late at pack time. The `llm_compressor` format keeps 8-bit asym: compressed-tensors represents it and vLLM serves it (Machete on SM90+ at group sizes 64/128/-1, Conch on SM80+ at group size 128 or channelwise, Humming as the fallback where enabled)), with a one-time warning for group sizes only SM90 serves. AutoScheme option pins and fixed layer pins (lm_head etc.) follow the same rule, with the effective allowance derived from the export format at layer-config generation time. A new `allow_w8_asym` flag (CLI, `AutoRound`, and `ModelFreeCompressor`) lifts every refusal and pin for serving stacks beyond stock vLLM.
- **New presets**: `W7A16`, G64/G32 group-size variants (W2-W8), and W5/W6/W7 unlocked for the `llm_compressor` format (compressed-tensors is the vLLM serving route for these widths).

## Type of Change

Bug fix / Feature (opt-in flag + presets)

## Related Issues

Not found

- [x] The CUDA CI has passed. You can trigger it by commenting `/azp run Unit-Test-CUDA-AutoRound`.


avtc:
This is a prerequisite before NeUQI addition.

Manually verified with phi-2 (full-vocab fp32 KL vs the base model, openwebtext 8x8192, streaming dequant student - bit-exact against compressed-tensors' own decompressor modulo the intended bf16 cast):

**1. SignRound asym + quant_lm_head + llm_compressor** - the tensor-zp int8 cast + signed-convention shift on an outside-block layer. Branch: clean export. Main: crashes at export after tuning completes (`Tensor must be quantized to torch.int8 before packing`).

```bash
autoround --model microsoft/phi-2 --asym --iters 1 --quant_lm_head --nsamples 16 \
  --format auto_round:llm_compressor --output_dir /tmp/t1_asym_lmhead
```

**2. Uniform W8A16 --asym** - the silent-corruption comparison. A correct asym quantization is never worse than sym, so the W8-sym run is the control:

| same protocol | mean | median | p99 | top1 | logit bias |
|---|---|---|---|---|---|
| W8 sym (control) | 0.0230 | 0.0185 | 0.0980 | 94.2% | -0.004 |
| W8 asym, **main** | 0.0356 | 0.0273 | 0.1624 | 92.3% | -0.084 |
| W8 asym, this branch + llm_compressor | 0.0234 | 0.0189 | 0.0986 | 93.8% | -0.012 |

On main the export succeeds but zero points above 127 wrap negative in the int8 cast: 1.55x the control with a 20x larger logit bias, and vLLM refuses the artifact at load (`Unsupported packing_format: auto_round`). On this branch the native-format route is refused at startup with a message pointing at the servable alternatives, and the llm_compressor route lands at parity with the control and loads/serves in vLLM (compressed-tensors WNA16 kernel path).

```bash
autoround --model microsoft/phi-2 --iters 0 --scheme W8A16 --asym \
  --format auto_round:llm_compressor --output_dir /tmp/w8asym_llmc   # branch
```

**3. AutoScheme mixed-bits with --asym + llm_compressor** - W7A16G32 + W8A16 at avg 7.75 bits. Branch: completes, 0.0239 mean / 0.0192 median / p99 0.1009 / 93.9% top1 - parity with the same run where 8-bit entries were forced symmetric (0.0242 / 0.0195 / 0.1028 / 93.6%); at 8 bits both are near-lossless, the point is the path. Main cannot run it: no G32 preset, and with plain widths it dies at export with the same int8 error as test 1, after the whole scoring pass.

```bash
autoround --model microsoft/phi-2 --avg_bits 7.75 --options "W7A16G32,W8A16" \
  --asym --iters 0 --format auto_round:llm_compressor --output_dir /tmp/t3_as_llmc
```

avtc:
I have found that LLM compressor supports int8 asym pack and load in vllm and serve, so will do changes to this PR to block only unsupported paths, with explicit --allow_w8_asym to bypass checks/overrides.

Update: PR description/comment/logic updated with improvements and self-review fixes

auto_round/compressors/model_free.py
```diff
@@ -519,12 +519,13 @@ def __init__(
...
         layer_config: Optional[dict] = None,
         ignore_layers: str = "",
-        format: str = "auto_round",
+        format: Optional[str] = None,
         device: str = "cpu",
         quant_lm_head: bool = False,
         quant_nontext_module: bool = False,
         enable_torch_compile: Optional[bool] = None,
         disable_opt_rtn: bool = False,
+        allow_w8_asym: bool = False,
```
xin3he: I think this patch is not necessary for model_free.py. Only sym is supported here.

auto_round/compressors/model_free.py
```diff
@@ -519,12 +519,13 @@ def __init__(
         scheme: Union[str, QuantizationScheme] = "W4A16",
         layer_config: Optional[dict] = None,
         ignore_layers: str = "",
-        format: str = "auto_round",
+        format: Optional[str] = None,
```
xin3he, line 522: why removing the default format?

auto_round_extension/torch/qlinear_torch_zp.py
```diff
@@ -140,7 +148,12 @@ def pack_248_bits(self, linear, scales, zeros, g_idx=None, device=None):
                 col += 1
             self.qzeros = qzeros.cpu()
         else:
-            zeros -= 1
+            # classic (zp-1) packing cannot represent zp=0 (see the tensor
+            # branch above): clamp the scalar the same way so a zp of 0
+            # degrades to a decodable +1 instead of corrupting the word.
+            # Normalize to int like the 3-bit path: the bit-shift packing
+            # below cannot shift a float scalar.
+            zeros = int(min(max(zeros - 1, 0), self.maxq))
```
wenhuach21, line 156: Nice catch!

@WeiweiZhang1  Since this is the approach typically used by the community, and it definitely has an accuracy issue,  we'd better dynamically change the QDQ simulation to avoid zp=0 during the tuning phase.

docs/step_by_step.md
```diff
@@ -123,10 +123,13 @@ AutoRound supports several Schemes:
 
 - **W4A16**(bits:4,group_size:128,sym:True,act_bits:16)
 - **W8A16**(bits:8,group_size:128,sym:True,act_bits:16)
-- **W6A16**(bits:6,group_size:128,sym:True,act_bits:16) — `mlx` format only
-- **W5A16**(bits:5,group_size:128,sym:True,act_bits:16) — `mlx` format only
+- **W7A16**(bits:7,group_size:128,sym:True,act_bits:16) - `llm_compressor` format
+- **W6A16**(bits:6,group_size:128,sym:True,act_bits:16) — `mlx` and `llm_compressor` formats
+- **W5A16**(bits:5,group_size:128,sym:True,act_bits:16) — `mlx` and `llm_compressor` formats
```
wenhuach21, line 128: @n1ck-guo  @xin3he  I assume this is supported by humming kernel and humming supports awq/gptq packing, so we'd better support 5/6/7 bits in AR format too

wenhuach21:
Really nice work, thanks!

auto_round/auto_scheme/delta_loss.py
```diff
@@ -3547,6 +3555,47 @@ def _select_embedding_scheme_index():
...
+    a servable W8-asym path (native auto_round, auto_gptq, marlin) cannot use
+    8-bit asym, and fixed layer_config pins (lm_head, MTP layer, ...) bypass
+    options. Entries WITHOUT an explicit sym would silently inherit the global
+    --asym and are pinned; an EXPLICIT sym=False is refused up front -- there
+    is no point spending quantization time on a configuration that can never
+    be served. allow_w8_asym skips both -- set by the builder from the
+    explicit --allow_w8_asym flag OR from an export format whose serving
+    stack accepts W8 asym (llm_compressor).
+    """
+    if allow_w8_asym:
```
wenhuach21, line 3573: @avtc Based on your accuracy data, it seems that at least for 8-bit quantization, W8 asymmetric performs similarly to W8 symmetric, while symmetric quantization is a little simpler.

Since we have added quite a bit of tricky code for this special scheme, I was wondering if we could just force it to symmetric quantization at the entry point for now, and re-enable asymmetric quantization for 8 bits later if it gets well supported by the community.  It's up to you

auto_round/auto_scheme/delta_loss.py
```diff
@@ -2548,6 +2548,7 @@ def _gen_layer_config(
     processor=None,
     is_vlm: bool = False,
     disk_index=None,
+    w8_asym_allowed: bool = False,
```
xin3he: I would prefer to use a environment variable instead of introducing a new argument for this case, since it's not a normal case. cc @wenhuach21

docs/step_by_step.md
avtc: Here's the factual matrix:

| route | W5/W6/W7 support today |
|---|---|
| **Humming kernel itself** | unconstrained - `can_implement` = CUDA + installed + no `g_idx`; any bit width, zp or not, any group size. The kernel *could* serve int5/6/7 from gptq/awq packings |
| **compressed-tensors (llm_compressor artifacts)** | **symmetric 5/6/7 accepted** - `WNA16_SUPPORTED_TYPES_MAP` has `uint5b16/uint6b32/uint7b64` > router can pick Humming > **served today**. Asym 5/6/7 refused (`WNA16_ZP_SUPPORTED_TYPES_MAP = {4, 8}` only) |
| **INC path (native auto-round artifacts)** | **hard refusal at config level** - `INCConfig.SUPPORTED_BITS = {2, 3, 4, 8}`; and inside, the gptq path only routes Marlin for `{(4,sym),(8,sym)}` (`gptq_type_map`), else `NotImplementedError: "Only 4-bit and 8-bit symmetric..."`; the awq map is `{4, 8}`. Crucially, **zero humming references in the INC loader** - Humming is reachable only via compressed-tensors today |

auto_round/auto_scheme/delta_loss.py
avtc: I'd keep it format-scoped rather than forcing sym at the entry: the phi-2 parity reflects a near-symmetric weight distribution, not a general property - on skewed per-group ranges asym retains real dynamic range, and the upstream NeUQI joint (scale, zp) search this PR is a prerequisite for is asym-native, so a hard sym pin would block it. The llm_compressor path is measured clean (parity with the sym control) and vLLM-servable, while native formats already pin/refuse; happy to revisit if the community converges on sym-only W8.

auto_round/compressors/model_free.py
avtc: Model-free does support asymmetric int schemes (the scheme validator explicitly accepts sym/asym for bits 2/4/8, and the shard packer has a real asym branch), but you're right that the parameter had no model-free-specific purpose - it only forwarded the opt-in through construction/re-check/fallback hops. It's removed in c37c0bcd together with the switch to the env variable.

auto_round/auto_scheme/delta_loss.py
avtc: Agreed - done in c37c0bcd: `AR_ALLOW_W8_ASYM` replaces the argument everywhere (including the CLI flag), read directly at the policy gates; documented in docs/environments.md.

docs/step_by_step.md
wenhuach21: Yes, we’ll support it ourselves.

auto_round/compressors/model_free.py
avtc: Two parts to the answer:

**Why the default changed at all:** the fallback compressor consumes `quantize_and_save(format=...)` only while its formats are unset. We had added format forwarding to carry the old `allow_w8_asym` argument across the model-free fallback, and with the default being a concrete `"auto_round"` every fallback got silently pinned to native - gguf saves wrote `model.safetensors` instead of `.gguf`, llm_compressor saves lost the compressed-tensors config, and gptq/awq saves hit the native deepcopy path on the meta skeleton (the 10 failures in the previous CI run all trace to this). Now that the opt-in is the `AR_ALLOW_W8_ASYM` env, that original purpose is gone.

**Why limited forwarding remains:** model-free falls back to the regular flow whenever it is asked to do something it cannot do itself - when `quantize_and_save(format=...)` requests a format outside its own packing set (for integer schemes it only produces `auto_round` / `auto_round:auto_gptq` natively, so `gguf:*`, `auto_awq`, `marlin`, or `llm_compressor` requests are re-routed), when an AutoScheme's format is incompatible with its option family, and when `quantize()` is called directly on it. In those cases a caller who constructed with an explicit `format="auto_round:llm_compressor"` and then calls `quantize_and_save()` without a format argument gets that construction format silently dropped on main - the export lands on the default instead of what was requested. Forwarding *only* an explicitly requested format closes that gap while leaving the implicit path exactly as on main: no explicit format -> nothing forwarded -> the save-time format is honored. The CLI is unaffected either way, since it passes the same value to construction and save.

Happy to drop the forwarding entirely if you prefer the smaller diff - implicit-path behavior is identical to main either way; only the explicit-construction corner would revert to upstream behavior.

auto_round/compressors/model_free.py
avtc: I am testing model-free asym right now...

avtc: Update on the model-free surface: testing the `asym` opt-in exposed a pre-existing packing bug (present on main, not introduced here). Model-free `asym` int artifacts declare packing_format: "auto_round" - the family whose consumers read the qzeros nibble as the zero point directly (qlinear_torch) - but the packer stored `zp - 1` (the auto_round:auto_gptq / qlinear_torch_zp convention). Every group's zero point decoded one step low, shifting all weights by +scale: W4A16-asym on phi-2 measured full-vocab KL `5.15` vs ~`0.12` after the fix (regular-flow exports were never affected - asym native routes to the AWQ packer there). The zero-point convention is now derived from the declared `packing_format` via the backend registry constants, shared by the packer and the config builder so the bytes and the declaration cannot drift apart, and the declared packing is a required argument on the packing call.

avtc:
@wenhuach21 something wrong with pipeline ( https://dev.azure.com/lpot-inc/neural-compressor/_build/results?buildId=74046&view=logs&j=abf08a52-6fca-57e2-8281-2cd20bcb00b9&t=47e98654-b9d3-59c3-85c9-e73592dce793&l=135 ):
```
Using Python 3.12.13 environment at: /root/.venv
   Updating https://github.com/ggml-org/llama.cpp.git (HEAD)
error: Git operation failed
  Caused by: failed to clone into: /root/.cache/uv/git-v0/db/d69dd14cf260707a
  Caused by: failed to fetch commit `0f3a71be15af836d277c9f918adfafb45732677e`
  Caused by: process didn't exit successfully: `/usr/bin/git fetch --force --update-head-ok 'https://github.com/ggml-org/llama.cpp.git' '+0f3a71be15af836d277c9f918adfafb45732677e:refs/commit/0f3a71be15af836d277c9f918adfafb45732677e'` (exit status: 128)
    --- stderr
    fatal: could not read Username for 'https://github.com'/: terminal prompts disabled
    fatal: expected flush after ref listing
```

avtc:
Looks like it was an issue with github login, encountered it as well locally, please retry run:
/azp run Unit-Test-CUDA-AutoRound

wenhuach21:
@chensuyue Could we also grant the collaborator privilege to run CUDA UT?

avtc:
c37c0bcd lost its sign-off during a history squash; all other commits are signed and the tree is unchanged - can you dismiss the DCO check, or shall I rebase with signoff?

wenhuach21:
> [c37c0bc](https://github.com/intel/auto-round/commit/c37c0bcd58aa2c678159cd049ce0db9f1d5c3031) lost its sign-off during a history squash; all other commits are signed and the tree is unchanged - can you dismiss the DCO check, or shall I rebase with signoff?

done

wenhuach21 (approved):

PR #2306, opened 2026-09-06
V0.15.1rc-testonly

chensuyue:
for CI test.


chensuyue:
Performance test failed caused by vram reduce, which is as expected.

Issue #2010, opened 2026-07-07
[Feature] Support quant parallelism on multi-GPU


chensuyue:

avtc:
I have been playing with different calibration-free quantization methods (offline SINQ, NeUQI, PERQ b512 MassDiff), another streaming implementation with layers weights prefetch on next device, then block-parallel-quantization for calibration-free iters0 path, and data-parallel quantization with `iters>0`, in this branch: https://github.com/avtc/auto-round/tree/feature/mixed-bits-low-vram-NeUQI-preSINQ-and-optimizations

I am still experimenting with cuda graphs caching across layers in layer-by-layer quantization and planned to take a look at batching expert modules so 300B MoE models will be able to be quantized on consumer hardware like RTX 3090 (or multiple of them).

I have made tests on Qwen3.8-27B model on server with 8x3090, with KLD eval.
From what I have found:

1. **NeUQI** works for `iters=0` improving KLD by **10%** (w4a16g128 asym) on top of opt-rtn (sym) and with optimized implementation it works fast and adds tiny fraction of time to whole quantization. For `iters=50` KLD improved by **7.3%** (w4a16g128 both asym). For `iters=200` by **1.0-1.7%** (w4a16g32 both asym). For different models it can give different effect, but have seen only positive.

2. **Data parallel** when batch distributed over several GPUs, each keeping whole layer - improves quantization speed, i.e. lowering time
- for `iters=50`, from `53m` to `31m` on 4x3090
- for `iters=200`, from `2h57m` to `1h37m` on 4x3090
note: using 8x3090 improves only by few minutes.

3. **Cuda graphs** (captured on each layer and copied to 3 gpus), lower the time for whole `iters=200` by only 2 minutes. (cached graph from another layer with identical modules ~~do not work for me, but if it would work it will~~ give another `1 minute`).

Feel free to use the branch and its improvements.
If you interested only in specific features from it - like NeUQI or Data-parallel (maybe with cuda-graphs as well) or offline SINQ/PERQ b512 MassDiff - I can try to extract this functionality and prepare a PR, just let me know (the branch derived by 300+ commits ahead in short time, and I would like to contribute).

---
There are also features:
- pack on gpu
- streaming reading and writing in a way no need to have much RAM or disk space (i have 64GB only and less than 300GB free space and was able to quant hy3 with iters=0).

wenhuach21:
@avtc That’s really great. I think all of them would be useful. 
[...]

PR #2341, opened 2026-09-10
NeUQI grid search for optimized RTN (--enable_neuqi: joint asym scale/zp search, two-stage sym search, frozen-init anchor)

avtc:
## Description

This PR adds the **NeUQI** grid search ([arXiv 2505.17595](https://arxiv.org/abs/2505.17595)) to the optimized-RTN path, opt-in via `--enable_neuqi` (CLI, or `enable_neuqi=True` in the API; carried by both `RTNConfig` and `SignRoundConfig`):

- **Asymmetric layers — joint (scale, integer zero-point) search**: a two-stage weighted grid search per group (coarse log-spaced scale candidates, then fine refinement; all zero points evaluated per candidate). Asymmetric layers previously had no optimized initializer at all — plain min/max RTN was the only path.
- **Symmetric layers — two-stage signed scale search**: replaces the incumbent min/max clip search when enabled; which clamp convention wins per group is searched, not assumed.
- **On the zero-shot path (`iters=0`) both searches are weighted by the activation imatrix**, collected automatically when the flag is set (asymmetric layers previously never consumed one — plain min/max RTN was their only path). With `iters > 0` the anchor runs unweighted; tuning itself is data-driven.
- **`iters > 0`: frozen-init anchor.** The search runs once at wrapper init and anchors the SignRound tuning grid — range margins are pinned to the winner, tuning proceeds on the rounding values only. Measured quality-neutral at `iters=50` on sym with SignRoundV2 (0.02191 vs 0.02203 without), so the anchor is a no-cost initialization once tuning runs.
- **MoE expert batching**: same-shape expert projections (both symmetry classes) quantize in one stacked search call — bit-identical to per-module results.
- **Backends**: the search runs on the fastest available backend — a Triton kernel (`auto_round_extension`), a torch.compile implementation, or plain eager PyTorch — and falls back to the next one if a backend fails at runtime (the failed one is disabled for the rest of the process, with a warning). `AR_NEUQI_BACKEND=triton|compile|eager` pins a backend, mainly for debugging. Measured on the joint search (2M groups, group size 128, 64+32 candidates, RTX 3090): eager 23.8 s, torch.compile 0.43 s (~55x faster than eager), Triton 0.33 s (~72x).
- **Grid sizes**: `AR_NEUQI_COARSE` / `AR_NEUQI_FINE` control the grid. Defaults are 256/64 when Triton or torch.compile serves the search and 64/32 with eager — the eager cost is proportional to the grid size and measured quality is flat between the two (Qwen3.8-27B W4A16 g128, iters=0: mean KL 0.02570 at 64/32 vs 0.02579 at 256/64). Explicit values always win.

Validation — full-vocabulary fp32 KL vs the BF16 base, openwebtext 8x8192 positions; Qwen3.8-27B, W4A16, group size 128, single RTX 3090. Protocol and full statistics: [docs/neuqi_acc.md](https://github.com/intel/auto-round/blob/main/docs/neuqi_acc.md). Sorted by mean KL (lower is better):

| mean | median | p90 | p99 | top1 | top5 | wall | recipe |
|--:|--:|--:|--:|--:|--:|--:|:--|
| 0.01921 | 0.00905 | 0.03591 | 0.17297 | 0.9370 | 0.8727 | 2:57:21 | iters=200, asym, SignRoundV2 + NeUQI (64/32 grid) |
| 0.02106 | 0.01050 | 0.04071 | 0.18038 | 0.9317 | 0.8661 | 0:49:07 | iters=50, asym, SignRoundV2 + NeUQI |
| 0.02191 | 0.01116 | 0.04237 | 0.18473 | 0.9304 | 0.8627 | 0:56:52 | iters=50, sym, SignRoundV2 + NeUQI |
| 0.02203 | 0.01112 | 0.04307 | 0.18239 | 0.9295 | 0.8630 | 0:55:40 | iters=50, sym, SignRoundV2, no NeUQI |
| 0.02579 | 0.01361 | 0.05046 | 0.21753 | 0.9234 | 0.8514 | 0:11:00 | iters=0, asym, NeUQI |
| 0.02734 | 0.01468 | 0.05358 | 0.22014 | 0.9213 | 0.8463 | 0:09:00 | iters=0, sym, NeUQI |
| 0.02852 | 0.01533 | 0.05678 | 0.23453 | 0.9191 | 0.8429 | 0:13:19 | iters=0, sym, incumbent OptRTN search (no NeUQI) |

All `iters > 0` rows use SignRoundV2 (`enable_alg_ext`).

## Type of Change

New feature

## Related Issues

Relates to #2010 — NeUQI support was requested there.

- [x] The CUDA CI has passed.

### DIFF
--- a/auto_round/algorithms/quantization/sign_round/config.py
+++ b/auto_round/algorithms/quantization/sign_round/config.py
@@ -120,6 +120,16 @@ class SignRoundConfig(QuantizationConfig):
             action="boolean_optional",
             help="Enable last-block LM cross-entropy (LFQ) loss for the final transformer block (experimental).",
         )
+        registry.add_argument(
+            "--enable_neuqi",
+            field="enable_neuqi",
+            action="store_true",
+            help=(
+                "Enable the NeUQI grid search (arXiv 2505.17595): the tuning grid is anchored to "
+                "the searched (scale, zero-point) result once at layer init and frozen during "
+                "tuning (joint search for asym layers, two-stage scale search for sym layers)."
+            ),
+        )
 
     def __init__(
         self,
@@ -140,6 +150,7 @@ class SignRoundConfig(QuantizationConfig):
         optimizer: str | None = None,  # TODO later wenhuach delete this
         enable_adam: bool = False,  # TODO later  wenhuach delete this
         enable_lfq: bool = False,
+        enable_neuqi: bool = False,
         **kwargs,
     ) -> None:
         """Initialize a SignRound configuration.
@@ -170,11 +181,18 @@ class SignRoundConfig(QuantizationConfig):
                 quantized output of previous blocks during calibration.
             optimizer: Optional optimizer name override.
             enable_adam: Whether to use the Adam-based SignRound variant.
+            enable_neuqi: Opt into the NeUQI grid search (arXiv 2505.17595):
+                the joint (scale, integer zero-point) search runs once at
+                layer init and anchors the tuning grid to the winner, frozen
+                during tuning (the two-stage symmetric scale search for sym
+                layers). The search grid sizes are tunable via
+                ``AR_NEUQI_COARSE`` and ``AR_NEUQI_FINE``.
             **kwargs: Common quantization arguments forwarded to
                 QuantizationConfig, such as bits, group_size, sym,
                 data_type, and activation quantization fields.
         """
         super().__init__(**kwargs)
+        self.enable_neuqi = enable_neuqi
         self.gradient_accumulate_steps = gradient_accumulate_steps
         self.iters = iters
         if self.iters < 0:
--- a/auto_round/data_type/neuqi.py
+++ b/auto_round/data_type/neuqi.py
@@ -1,0 +1,1333 @@
+# Copyright (c) 2026 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Near-optimal uniform (scale, zero-point) search for asymmetric integer quantization.
+
+This is a clean-room reimplementation of the integer-zero-point variant of the NeUQI
+objective (arXiv 2505.17595, "NeUQI: Near-Optimal Uniform Quantization Parameter
+Initialization for Low-Bit LLMs"). The reference implementation carries no license,
+so no code is copied from it; this module implements the published optimization from
+scratch. Only the integer-zero-point variant is implemented because the paper's
+floating-point zero point is not consumable by stock int4 kernels (GPTQ/AWQ/Marlin
+assume an integral zero point).
+
+For each quantization group the following problem is solved:
+
+    min_{s > 0, z in {0, .., 2**bits - 1}}
+        sum_i qw_i * (s * (clamp(round(w_i / s) + z, 0, 2**bits - 1) - z) - w_i)**2
+
+where ``qw`` is an optional per-element weighting (e.g. the activation ``imatrix``).
+For a fixed scale the loss decomposes over the integer zero points, so the optimal
+``z`` is found exactly by enumerating all ``2**bits`` candidates (16 for 4-bit).
+The scale is searched on a coarse log grid bounded by the min-max scale (scales
+beyond the min-max range are never optimal) followed by a fine additive refinement
+around the best coarse entry.
+
+The search replaces the plain min/max initialization of the asymmetric optimized-RTN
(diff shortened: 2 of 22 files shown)


wenhuach21:
@avtc Really appreciate this PR! May I know the additional quantization cost and accuracy results on a few representative tasks, like mmlu? If the accuracy looks good and cost is low, we could consider setting it as the default.

avtc:
@wenhuach21 

Both NeUQI sym/asym are faster than opt-rtn sym at least on CUDA (`triton` or `torch.compile` backend), as can be seen in table for `iters=0`: sym optrtn=`13m`, sym optrtn+neuqi=`9m`, asym optrtn+neuqi=`11m`.
`eager` mode is 74x times slower than `triton` and 54x times slower than `torch.compile`, so on CPU could be slow.

`iters=50` passes were running in parallel on the same host, so could be less representative.

`mmlu` | score | stderr
|---|---|---|
iters=0, optrtn, sym | 0.8281 | ± 0.0030
iters=0, optrtn, neuqi, sym | 0.8271 | ± 0.0031
iters=0, optrtn, neuqi, asym | 0.8294 | ± 0.0030

auto_round/data_type/neuqi.py
```diff
@@ -0,0 +1,1333 @@
...
+``z`` is found exactly by enumerating all ``2**bits`` candidates (16 for 4-bit).
+The scale is searched on a coarse log grid bounded by the min-max scale (scales
+beyond the min-max range are never optimal) followed by a fine additive refinement
+around the best coarse entry.
+
+The search replaces the plain min/max initialization of the asymmetric optimized-RTN
+path and is a strict generalization of the symmetric ``search_scales`` grid (which
+fixes ``z = 0``).
+
+The symmetric counterpart (:func:`neuqi_search_scale_sym`) fixes the zero point at
```
wenhuach21, line 40: How about moving this to the algorithm? It feels more like an algorithm than a data type, although there may be some coupling if different algorithms want to use it.

@n1ck-guo, what do you suggest?

auto_round/algorithms/quantization/sign_round/config.py
```diff
@@ -120,6 +120,16 @@ def register_args(cls, registry: AlgorithmParameterRegistry) -> None:
             action="boolean_optional",
             help="Enable last-block LM cross-entropy (LFQ) loss for the final transformer block (experimental).",
         )
+        registry.add_argument(
+            "--enable_neuqi",
+            field="enable_neuqi",
```
wenhuach21, line 125: @n1ck-guo Should we consolidate this with `disable_opt_rtn` into a unified `qparams_init_method` ?

docs/step_by_step.md
```diff
@@ -595,6 +595,8 @@ AutoRound also supports Optimized RTN (Round-To-Nearest) mode for fast, calibrat
 
 For the GGUF format, we have optimized the RTN algorithm inspired by llamacpp. To use the original (pure) RTN algorithm instead, enable the `--disable_opt_rtn` option.
 
+Passing `--enable_neuqi` opts the optimized path into the **NeUQI** grid search ([arXiv 2505.17595](https://arxiv.org/abs/2505.17595)): asymmetric layers run a joint (scale, integer zero-point) search and symmetric layers a two-stage signed scale search, both weighted by the activation imatrix on the zero-shot path, which is collected automatically (with `iters > 0` the anchor runs unweighted). With `iters > 0`, the search result anchors the SignRound tuning grid (frozen init). Grid sizes are tunable via `AR_NEUQI_COARSE`/`AR_NEUQI_FINE` (see [environments](./environments.md)); the unpinned default is backend-aware (wide only on the Triton/torch.compile lanes). Accuracy and wall-time results are shown in [NeUQI accuracy validation](./neuqi_acc.md).
```
wenhuach21, line 598: @WeiweiZhang1 If the author doesn’t have enough resources to evaluate accuracy on some typical models and tasks, could you help run some evaluations and link the results as we usually do? It would also be helpful to provide users with some guidance on when to trigger this.

wenhuach21 (approved):

wenhuach21:
@copilot resolve the merge conflicts in this pull request

avtc:
@wenhuach21 
We'll rebase this ourselves — the datatype refactor needs a semantic port, should be up shortly.

avtc:
We've rebased onto current main ourselves - single commit force-pushed.

wenhuach21:
> @wenhuach21 We'll rebase this ourselves — the datatype refactor needs a semantic port, should be up shortly.

Ok，we may revert pr 2370，i‘d suggest you wait for us to avoid further conflict

wenhuach21:
sorry，2360

avtc:
@wenhuach21 there are no conflicts vs main - only conversations left unresolved

auto_round/data_type/neuqi.py
avtc: The current home follows the existing optimized-RTN family: `search_scales` (the clip search) and `quant_tensor_opt_rtn_sym` already live in `data_type`, and the NeUQI searches are their grid-search counterparts - same call sites, same (scale, zp) contract. The coupling you mention is also what keeps it there: RTN and SignRound both consume the searches through the quantizer lifecycle, and `data_type` is the shared layer between the algorithms, so the search stays algorithm-agnostic. Happy to move it if you and @n1ck-guo prefer a different home - the move is mechanical; the engagement points stay as they are.

Issue #2358, opened 2026-09-14
[Bug]: Preserve FP32 modules during model dtype conversion


changwangss:
### Problem Description

  AutoRound may cast the entire model to BF16 during AMP initialization, including parameters declared in _keep_in_fp32_modules.

  For example, with a preloaded FP32 Wan transformer:

  1. The model declares certain modules should remain FP32.
  2. ModelContext._set_amp_dtype() calls model.to(torch.bfloat16).
  3. Diffusion pipeline dtype alignment later restores protected parameters to FP32, but their original precision has already been lost.

  A protected parameter demonstrates the issue:

  Original FP32:        1.0010000467
  After BF16 cast:     1.0
  Restored to FP32:    1.0

  Expected behavior

  Model dtype conversion should preserve the original values of parameters declared to remain FP32, while converting other parameters to the requested dtype.

### Reproduction Steps

  Example: Wan2.2

  WanTransformer3DModel declares time_embedder, scale_shift_table, and norm1/norm2/norm3 in _keep_in_fp32_modules. Their parameters should remain FP32.

  For example:

  condition_embedder.time_embedder.linear_1.weight
  Expected: FP32
  After AutoRound AMP dtype conversion: BF16

  When AutoRound calls model.to(torch.bfloat16), these protected parameters are also converted to BF16. Subsequent pipeline alignment may restore their dtype to FP32, but cannot recover the precision
  already lost.

### Error Logs

```shell

```

PR #2359, opened 2026-09-14
fix: preserve declared FP32 tensors during model dtype conversion

changwangss:
## Description

Example: Wan2.2

WanTransformer3DModel declares time_embedder, scale_shift_table, and norm1/norm2/norm3 in _keep_in_fp32_modules. Their parameters should remain FP32.

For example:

condition_embedder.time_embedder.linear_1.weight
Expected: FP32
After AutoRound AMP dtype conversion: BF16

When AutoRound calls model.to(torch.bfloat16), these protected parameters are also converted to BF16. Subsequent pipeline alignment may restore their dtype to FP32, but cannot recover the precision
already lost.

## Type of Change

Bug fix

## Related Issues

Fixes or relates to #2358 


auto_round/utils/model.py
```diff
@@ -1789,17 +1789,45 @@ def check_seqlen_compatible(input_seqlen, tokenizer=None, model=None):
...
+    def convert(tensor):
+        if not (tensor.is_floating_point() or tensor.is_complex()):
+            return tensor
+        target_dtype = torch.float32 if id(tensor) in protected else dtype
+        return tensor.to(dtype=target_dtype)
+
+    return model._apply(convert)
+
+
 def _to_model_dtype(model, model_dtype):
```
n1ck-guo, line 1820: Please help add a check to confirm that model_dtype is of type str

n1ck-guo (approved):
LGTM

auto_round/utils/model.py
changwangss: Added an isinstance(model_dtype, str) guard in 1a646965. Non-string inputs, including None and torch.dtype, leave the model unchanged.

Issue #1651, opened 2026-04-03
[Bug]: gemma 4: RuntimeError: The size of tensor a (512) must match the size of tensor b (256) at non-singleton dimension 3


XuehaoSun:
### Problem Description

```
2026-04-03 09:25:38 INFO __main__.py L599: start to quantize google/gemma-4-26B-A4B
2026-04-03 09:25:39 INFO autoround.py L178: using MLLM mode for multimodal model.
model.safetensors.index.json: 103kB [00:00, 7.08MB/s]
Fetching 2 files: 100%|██████████| 2/2 [09:06<00:00, 273.07s/it]
Download complete: 100%|██████████| 51.6G/51.6G [09:06<00:00, 94.5MB/s]00:17, 61.7MB/s]
Loading weights: 100%|██████████| 1013/1013 [00:00<00:00, 4362.49it/s]]
generation_config.json: 100%|██████████| 181/181 [00:00<00:00, 729kB/s]
processor_config.json: 1.69kB [00:00, 426kB/s]181 [00:00<?, ?B/s]
2026-04-03 09:35:42 INFO base.py L517: using torch.bfloat16 for quantization tuning
2026-04-03 09:35:42 INFO base.py L834: 'enable_torch_compile' is set to `False` by default. Enabling it can reduce tuning cost by 20%, but it might throw an exception.
2026-04-03 09:35:42 WARNING formats.py L166: some layers are skipped quantization (shape not divisible by 32): model.vision_tower.encoder.layers.[0-26].mlp.down_proj.linear, model.vision_tower.encoder.layers.[0-26].mlp.gate_proj.linear, model.vision_tower.encoder.layers.[0-26].mlp.up_proj.linear
2026-04-03 09:35:42 INFO replace_modules.py L107: Experts (before replacement) [model.language_model.layers.0.experts] (Gemma4TextExperts):
Gemma4TextExperts(
  (act_fn): GELUTanh()
)
2026-04-03 09:35:42 WARNING modeling_utils.py L4432: `loss_type=None` was set in the config but it is unrecognized. Using the default loss: `ForCausalLMLoss`.
2026-04-03 09:35:42 INFO device.py L1690: Before applying custom replacements 'peak_ram': 3.72GB, 'peak_vram': 48.32GB
2026-04-03 09:35:53 INFO moe_experts_interface.py L642: [MoE Prep] Unfused 30 MOE experts modules
[...]

PR #2366, opened 2026-09-16
feat: add NVFP4 static KV cache quantization support

yiliu30:
## Description

Add NVFP4 static KV cache quantization: calibrate per-layer K/V global scales on the full calibration set and export them as `k_global_scale`/`v_global_scale` parameters together with a compressed-tensors `kv_cache_scheme` (NVFP4 preset input-activation args) in the `llm_compressor` export format.

Key points:

- **KV cache calibration**: per-side running amax with NVFP4 QDQ simulation during calibration; the output hook writes the static global scales only after magnitudes have been observed.
- **Full-calibration-set activation amax pass** merged onto the layers before tuning, so static activation global scales are calibrated over the whole dataset.
- **Stored KV scale convention**: the checkpoint stores the dequantization multiplier `amax / 2688` (reciprocal of the weight-style global scale). This matches the vLLM NVFP4 KV store kernel, which computes `global_scale = 1 / k_scale`; storing `2688 / amax` instead would underflow the fp8 e4m3 block scales for typical activation magnitudes (e4m3 min subnormal 2^-9) and zero the served KV cache.
- **CLI**: `--static_kv_dtype` now accepts `nvfp4`.
- **Validation**: Qwen3-8B (w4a4 NVFP4 + NVFP4 KV) exported with this branch loads in vLLM with `kv_cache_dtype=nvfp4` and generates coherent text; all 36 layers serve scales that match the checkpoint exactly.

## Type of Change

New feature

## Related Issues

Fixes or relates to #

Notes:

- New/updated tests: `test/unit/test_cpu/export/test_llmc_format.py` (`TestLLMCNVFP4KV`: config, stored scales, vLLM-runtime round trip) and `test/unit/test_cpu/calibration/` (K/V amax tracking, stored-scale convention, full-set act amax wiring).
- CPU unit tests for the touched areas pass; full CPU suite failures on the dev box are pre-existing environment issues (missing optional deps, CUDA assert cascade) and reproduce on a clean tree.


auto_round/experimental/kv_cache.py
```diff
@@ -107,16 +110,28 @@ def __new__(cls, *args, **kwargs):
...
+            # NVFP4 block scales are computed per 16-element group at runtime;
+            # only a static per-tensor global scale is calibrated, so "head"
+            # granularity does not apply.
+            self.granularity = FP8_GRANULARITY_TENSOR
+        else:
+            assert dtype == torch.float8_e4m3fn, "Only fp8_e4m3fn is supported for now."
+            self.granularity = normalize_fp8_granularity(granularity)
+        # Set when a layer turns out incompatible with NVFP4 KV (e.g. head_dim
+        # not divisible by 16); all subsequent updates become identity passes.
+        self.disabled = False
```
lvliang-intel: self.disabled is global, but the head_dim % 16 check is per layer. One unsupported layer may disable NVFP4 KV quantization for all layers. Can we make this per-layer or fail early?

auto_round/experimental/kv_cache.py
lvliang-intel, line 1: The current implementation uses a running amax for QDQ during calibration, but the final exported scale is calculated from the maximum amax across all calibration data. This means different calibration samples may use different scales, while the exported scale is fixed.

Could we collect the final K/V amax first and then use the fixed scale for QDQ? This would make the calibration behavior consistent with the exported static KV scale.

auto_round/export/export_to_llmcompressor/export_to_fp.py
```diff
@@ -236,6 +264,32 @@ def _build_mixed_fp_quantization_config(
...
+            "static_kv_dtype 'nvfp4' conflicts with the FP8 static attention/KV options; "
+            "please set only one of them."
+        )
+    if use_nvfp4_kv:
+        if not _attention_modules_have_nvfp4_kv_scales(model):
+            logger.warning(
+                "No NVFP4 KV cache global scales were collected (calibration may not have run); "
+                "skipping the kv_cache_scheme in the exported compressed-tensors config."
+            )
+            return None
```
lvliang-intel: If NVFP4 KV scale collection fails, the code only logs a warning and continues the export. Since the user explicitly requested NVFP4 KV cache, could we fail with an error instead of silently falling back?

auto_round/experimental/kv_cache.py
yiliu30: Addressed in 318a8a25: we fail fast per layer instead. initialize_quantized_kv_cache now raises ValueError at setup time for any attention layer whose head_dim is not divisible by 16 (error includes the layer name, layer_idx and head_dim), and the runtime guard in _nvfp4_track_amax raises as well, so a single unsupported layer can never silently disable KV quantization for all layers. The global disabled flag is removed. Covered by new unit tests (test_nvfp4_kv_incompatible_head_dim_fails_early / test_nvfp4_kv_incompatible_head_dim_raises_at_update).

auto_round/experimental/kv_cache.py
yiliu30: Addressed in 318a8a25: we aligned with the current observer-based KV calibration in llm-compressor (post-#1651), where the K/V calibration hooks are observe-only and the static global scale is computed once from the final amax over the full calibration set. The NVFP4 update() path now accumulates the per-side running amax and lets K/V pass through unmodified, so calibration no longer fake-quantizes with an incomplete scale; k/v_global_scale is still derived from the final amax (amax/2688, vLLM dequant-multiplier convention). Validated on the tiny model: all 24 W4A4 input/weight global scales stay bit-identical to the no-KV calibration and the k/v global scales are unchanged.

auto_round/export/export_to_llmcompressor/export_to_fp.py
yiliu30: Addressed in 318a8a25: _resolve_kv_cache_scheme now raises ValueError when an explicit NVFP4 KV request has no collected k/v global scales (including models without any attention modules), instead of logging a warning and silently exporting without the kv_cache_scheme. Covered by test_nvfp4_kv_missing_scales_raises.

lvliang-intel (approved):
Comments fixed. LGTM.

n1ck-guo (approved):
LGTM

Issue #2332, opened 2026-09-10
[Bug]:  some weights should kept as FP32 in tuning phase


wenhuach21:
### Problem Description

<img width="1275" height="671" alt="Image" src="https://github.com/user-attachments/assets/1ff347b5-db89-4767-ac1e-17e588ed07f6" />

<img width="1286" height="647" alt="Image" src="https://github.com/user-attachments/assets/84957338-ba88-43b6-8f2f-7674f48b694a" />

### Reproduction Steps

~

### Error Logs

```shell

```

xin3he:
related https://github.com/intel/auto-round/pull/2359/changes

wenhuach21:
We should root-cause this issue and fix it if it turns out to be an AutoRound issue. Xin’s approach should only be used as a backup, since the tuning process is not aware of it.

PR #2375, opened 2026-09-17
fix: restore FP32 tensors saved as FP16/BF16 during tensor copying

xin3he:
## Description

This pull request enhances the logic for restoring FP32 tensors that were incorrectly saved as FP16 or BF16, improves the handling of tensor files during copying, and adds comprehensive tests to ensure correct restoration of tensor data types. The changes also include a minor adjustment to the list of tensor block names to ignore during quantization.

**Restoration and handling of FP32 tensors:**

* Added the `_restore_special_fp32_tensors` function to detect and restore tensors that were originally FP32 but were saved as FP16 or BF16, updating the target files in place. This ensures data fidelity when copying tensors between models.
* Modified `copy_missing_tensors_from_source` to use a mapping of tensor names to their file locations (`saved_tensor_to_file`), enabling more precise restoration and file handling. The new restoration logic is now invoked before identifying missing tensors.

**Testing improvements:**

* Added two new unit tests: one to verify that FP32 tensors saved as FP16 are restored to FP32, and another to check restoration in the case of sharded tensor files with index files. These tests ensure the new restoration logic works for both single and multi-shard scenarios.

**Other improvements:**

* Updated the list of block names to ignore during quantization to better match tensor naming patterns, changing `".mlp.gate."` to `".gate."`.
* Added import of `tempfile` to support atomic file replacement during tensor restoration.

## Type of Change

Bug fix

## Related Issues

Fixes or relates to #2332 


auto_round/utils/missing_tensors.py
```diff
@@ -52,6 +53,70 @@
...
+
+    target_names_by_shard: dict[str, list[str]] = {}
+    for tensor_name in candidates:
+        target_names_by_shard.setdefault(saved_tensor_to_file[tensor_name], []).append(tensor_name)
+
+    tensors_to_restore: dict[str, torch.Tensor] = {}
+    for target_shard, tensor_names in target_names_by_shard.items():
+        with safe_open(target_shard, framework="pt", device="cpu") as target_file:
+            for tensor_name in tensor_names:
+                if target_file.get_slice(tensor_name).get_dtype() in {"F16", "BF16"}:
```
changwangss: The FP32 → FP16/BF16 dtype difference alone does not mean a tensor was incorrectly saved. For an FP32 checkpoint loaded in FP16/BF16 with enable_norm_bias_tuning=True, this also matches tuned norm weights and biases. Replacing them with the original checkpoint values would silently discard the tuning results during export.

Could we restrict restoration to parameters explicitly required to remain FP32 and ensure that legitimate parameter updates are preserved? A regression test covering modified target values would help.

auto_round/utils/missing_tensors.py
```diff
@@ -52,6 +53,70 @@
...
 
+def _restore_special_fp32_tensors(
+    source_tensor_to_file: dict[str, str],
+    saved_tensor_to_file: dict[str, str],
+) -> None:
+    """Restore source FP32 tensors that were saved as FP16 or BF16."""
+    from safetensors import safe_open
+    from safetensors.torch import save_file
+
+    common_tensor_names = source_tensor_to_file.keys() & saved_tensor_to_file.keys()
```
changwangss: The exact key intersection misses tensors renamed during checkpoint loading, such as Gemma’s language_model.model.* ↔ model.language_model.* and Nemotron’s backbone.* → model.*.

The existing missing-tensor check recognizes these aliases and treats the tensors as already present, so neither path restores their FP32 values. Could we reuse the existing name mappings when pairing source and target tensors, and add a test covering a renamed FP32 tensor saved as BF16?

auto_round/utils/missing_tensors.py
```diff
@@ -52,6 +53,70 @@
...
+    for target_shard, tensor_names in restore_names_by_shard.items():
+        with safe_open(target_shard, framework="pt", device="cpu") as target_file:
+            metadata = target_file.metadata()
+            shard_tensors = {name: target_file.get_tensor(name) for name in target_file.keys()}
+        for tensor_name in tensor_names:
+            shard_tensors[tensor_name] = tensors_to_restore[tensor_name]
+
+        temporary_path = None
+        try:
+            with tempfile.NamedTemporaryFile(
```
changwangss, line 112: The temporary file has 0600 permissions. After os.replace(), the target shard also gets these permissions, so other users may lose read access. Could we preserve the original file permissions before
  replacing it?

auto_round/utils/missing_tensors.py
xin3he: Right, but we aims to provide highest accuracy if the tensor is not intended to be quantized.
Usually, the original checkpoint are using bf16 tensors, if FP32 is used, it must be important.

auto_round/utils/missing_tensors.py
xin3he: Not sure whether this issue exists.
The designed process is reverting tensor name during saving. so missing_tensors.py could detect them correctly.
vLLM requires original name for loading so reverting tensor name is necessary and it's implemented already I think.

auto_round/utils/missing_tensors.py
xin3he: Is this temporary file removed automatically, it should not be awared by users.

auto_round/utils/missing_tensors.py
xin3he: Thanks for raising this concern. The norm bias tuning edge case is valid in principle, but in practice:
1. Most modern models use BF16 for all tensors including norms, so the source checkpoint's norm weights are typically BF16 (not FP32).
2. The condition for restoration is narrow: source FP32 + target FP16/BF16. If the source norm weights are BF16 (the common case), they won't be restored.
3. The rare edge case (FP32 source norms + norm bias tuning) would require a model that explicitly uses FP32 for norms AND enables norm bias tuning, which is uncommon.

I agree that a more precise check (e.g., comparing with pre-tuning values) would be ideal, but that's not feasible without access to the model's internal state. I'll note this as a known limitation. For now, the heuristic is safe because the condition is narrow and the common case (BF16 source norms) is not affected.

auto_round/utils/missing_tensors.py
xin3he: will fix

auto_round/utils/missing_tensors.py
xin3he: image

xin3he:
Update with warning log for rename logic

auto-round /models/gemma-3-4b-it --iters 0
```
2026-09-18 13:47:18 WARNING logging.py L340: Transformers checkpoint->model renaming detected: 'model.language_model' -> 'language_model.model'. Using reverted checkpoint name.

```

auto_round/utils/missing_tensors.py
```diff
@@ -465,7 +477,7 @@ def _woq_quantize_missing_tensors(target_dir: str, missing_tensors_dict: dict) -
     """
     import re as _re
 
-    BLOCK_NAME_TO_IGNORE = [".shared_expert_gate.", ".mlp.gate.", ".g_proj.", "mtp.fc."]
+    BLOCK_NAME_TO_IGNORE = [".shared_expert_gate.", ".gate.", ".g_proj.", "mtp.fc."]
```
n1ck-guo, line 480: mlp.gate was changed to .gate, which expands the scope of impact. Please confirm its correctness. In addition, if BLOCK_NAME_TO_IGNORE is to be used with regular expressions, please pay attention to the usage of the . symbol to avoid introducing unnecessary bugs.

n1ck-guo (approved):
LGTM

auto_round/utils/missing_tensors.py
xin3he: Thanks, this change is to suit the sparse_moe.gate, and re is not used so no need to worry the `.`

Issue #2142, opened 2026-08-11
[Feature]: accuracy issue Intel/Qwen3.6-27B-int4-AutoRound


wenhuach21:
### Feature Description

https://huggingface.co/Intel/Qwen3.6-27B-int4-AutoRound/discussions/3

### Motivation and Use Case

~

### Alternatives Considered

~

lkk12014402:
hi @wenhuach21  @chensuyue  the model was not created by me

wenhuach21:
I have no time to check this issue. This model is created by AR, we should check whether using other recipe or other algorithms could fix this issue

wenhuach21:
**W/O fp8 KV CACHE
CUDA_VISIBLE_DEVICES=2 vllm serve Qwen3.6-27B-int4-AutoRound/ --served_model_name qwen --reasoning-parser qwen3
**Task 1 mixed-bits problem****


(base) wenhuach@mlp-dgx-01:~$ curl -s http://localhost:8000/v1/chat/completions   -H "Content-Type: application/json"   -d '{
    "model": "qwen",
    "messages": [
      {
        "role": "user",
        "content": "Explain how to implement layer-wise mixed-precision quantization under a global average-bit budget using dynamic programming. Provide the algorithm, explanation, Python implementation, complexity, and a small example."
      }
    ],
    "temperature": 0.2,
    "max_tokens": 4096
  }' | jq -r '
    (.choices[0].message.reasoning_content // ""),
    (.choices[0].message.content // "")
  '



Here is a complete guide to implementing layer-wise mixed-precision quantization under a global average-bit budget using dynamic programming.

---
## 1. Problem Formulation

Given:
- $L$ layers in a neural network
- A set of allowed bit-widths $\mathcal{B} = \{b_1, b_2, \dots, b
(base) wenhuach@mlp-dgx-01:~$ curl -s http://localhost:8000/v1/chat/completions   -H "Content-Type: application/json"   -d '{
    "model": "qwen",
    "messages": [
      {
        "role": "user",
        "content": "Explain how to implement layer-wise mixed-precision quantization under a global average-bit budget using dynamic programming. Provide the algorithm, explanation, Python implementation, complexity, and a small example."
      }
    ],
    "temperature": 0.2,
    "max_tokens": 4096
  }' | jq -r '
    (.choices[0].message.reasoning // ""),
    (.choices[0].message.content // "")
  '
Here's a thinking process:

1.  **Understand User Request:**
[...]

PR #2383, opened 2026-09-20
feat: chunked tuning for lm_head-class layers

avtc:
## Description

Full-vocabulary lm_heads become quantizable and tunable on a single 24 GB GPU, at every recipe: RTN, optimized RTN (iters=0), and SignRound (iters>0).

Even for a model like Qwen3.8-27B, lm_head on upstream can be quantized only in RTN: OptRTN on 4×3090 finishes with lm_head quantized with RTN, and iters>0 on 4×3090 crashes at the end of the full-model quantization, at lm_head start. On the PR branch, 1×3090 is able to quantize RTN/OptRTN/SignRoundV2 iters>0. Comparison table below.

**How it works**

- lm_head tune: input positions are sliced to an output-element budget, and fake-quant runs over row blocks of the weight — one chunk of the `[tokens × vocab]` output is live at a time, and each row block's forward, loss, and backward complete before the next. A failed tune falls back to export-time completion; missing tail inputs fall back to zero-shot RTN.
- Calibration reuses the block loop's cached chain outputs — the last block's fp and quantized-chain rows, the exact tensors lm_head sees at inference. The imatrix-weighted search runs over those rows, as for block layers.
- The best-parameters snapshot is weight-sized, and a device↔host copy of a tensor that large is slow — it happens on every improving iteration. The snapshot stays on the layer's device when it fits beside the row-window floor, moves to an idle peer GPU otherwise, and falls back to the host when both are tight. The previous snapshot is released before its replacement is cloned.
- Quantized out-of-block layers pack immediately, so shards stream per block. On Qwen3.8-27B with the lm_head quantized (64 GB host), host RAM during save drops from 59 GB + 35 GB swap (whole-model pack at the end) to ~16 GB peak.

**Parity**

- Chunked lm_head tune reproduces the single-shot tuned weights bit-for-bit (`torch.equal`, full tune run both ways).
- e2e on a real tiny model: identical tuner inputs and identical final quantized weights versus the capture lane.

**Measured** — Qwen3.8-27B, W4A16 g128 sym on lm_head, body bf16. Branch runs on one RTX 3090; main runs: RTN on 1 GPU, OptRTN and SignRound on 4. KL = full-vocab fp32 vs bf16 base, openwebtext 8×8192. RAM = host peak; VRAM = per-GPU peak (max of 4 where 4 GPUs).

| recipe | wall, main | wall, this branch | lm_head stage, main | lm_head stage, branch | RAM/VRAM, main | RAM/VRAM, branch | mean KL | p99 | top1 |
|---|---|---|---|---|---|---|---:|---:|---:|
| RTN iters=0 | 2:36, 1 GPU (`--disable_opt_rtn --disable_model_free`) | **2:25, 1 GPU** | ~6 s on CPU (remaining-layers path) | ~7 s on GPU | 10.0 / 4.8 GB | 9.9 / 7.3 GB | 0.01272 | 0.056 | 0.9274 |
| OptRTN iters=0 | 18:38, 4 GPUs | **8:06, 1 GPU** | search OOMs on GPU; CPU fallback runs plain RTN (~17 s) | ~22 s on GPU | 56.2 / 22.8 GB | 23.1 / 9.3 GB | **0.00898** | 0.040 | 0.9392 |
| SignRound iters=10 | 25:02, 4 GPUs — crashed | **12:28, 1 GPU** | input capture 5:18, then OOM at tune start | ~62 s on GPU | 48.2 / 21.8 GB (pre-crash) | 15.9 / 22.6 GB | **0.00807** | 0.036 | 0.9419 |

Optimized-RTN buys ~29% over plain RTN; 10 SignRound iterations buy a further ~10%. The rest of each branch wall is the 64 16-bit-pinned body blocks and calibration.

**Out of scope: MTP predictor trees.** A separate in-run tree-tuning lane (a materialized predictor tree fed embedding-side inputs) was explored on a precursor branch and measured in isolation (tree W4A16 g128 sym, lm_head and body bf16, same KL protocol on the tree's draft output): zero-shot optimized RTN stays the best tree recipe on this model — draft KL 0.00739 (top1 0.9445) versus 0.00754 (0.9423) for plain RTN and 0.00836 (0.9446) for a 10-iteration SignRound tune. Trees: not in this PR.

**Tests**: 90 CPU-runnable tests — chunked tune math, tail derivation, snapshot placement, packing gates, and the e2e parity pair.

## Type of Change

New feature (capability + performance) + bug fix

## Related Issues

- #2142 motivated the lm_head tuning work.


auto_round/algorithms/quantization/sign_round/quantizer.py
```diff
@@ -42,6 +45,38 @@
     from auto_round.algorithms.composer import BlockContext
 
 
+# Output-element budget for one forward/backward while tuning an outside-block
+# layer: a full-vocabulary lm_head at long seqlen cannot hold its fp32 logits
+# plus backward transients on a single GPU, so the tune loop slices the
+# sample's rows to stay within this many output elements per slice (64M = a
+# 256MB fp32 logits tensor; the accumulated gradient is unchanged)
+_OUTSIDE_TUNE_CHUNK_OUT_ELEMS = 2**26
```
wenhuach21, line 53: Could we leverage or learn from Unsloth’s implementation, which uses a Triton kernel? I assume it would be faster.
https://github.com/unslothai/unsloth/blob/04c9951b4d371860491a6a0b1098f57e90eff270/unsloth/kernels/cross_entropy_loss.py#L114

avtc: Thanks for the pointer! The chunking here is memory-driven rather than speed-driven: a 248k-vocabulary logits tensor plus its backward transients doesn't fit a 24 GB card, so the tune computes one row-block at a time and accumulates gradients unchanged - the result is identical to the single-shot loss. At this scale the lane is already fast: on Qwen3.8-27B (248k vocab, one RTX 3090) the whole lm_head stage takes about 22 s for the optimized-RTN search and about a minute for ten SignRound iterations; the run's wall is dominated by the decoder blocks and calibration, not this layer.

On Unsloth's kernel specifically: it fuses a chunked linear + cross-entropy (forward and backward) so logits never materialize - the right pattern for CE against label indices in training. Our loss is MSE against a dense reference computed by the unquantized layer on the same chunk, so a fused version would need to fold two GEMMs plus the elementwise difference and a backward into the weight and rounding values - and the memory it saves is already bounded here, because only one [rows × block] prediction/reference pair is ever live (the whole search peaks at 22.6 GiB on a 24 GB card). So we're not adopting it in this PR: the stage it would accelerate is a sub-minute fraction of runs measured in minutes.

---
note: I have updated PR description with exact wall-clock time and lm_head time for each run

avtc: Follow-up with numbers: the iteration is GEMM-bound - of the ~5.5-6 s per tune iteration, roughly 4-5 s is the prediction/reference/backward matrix multiplications, which a fused loss kernel leaves untouched; fusing would remove only the elementwise part (fp32 casts, mask, MSE, chunk materialization), about 0.5-1.5 s per iteration - 10-20 % of this stage, single-digit percent of a whole-model run. On complexity: our backward has to flow through the fake-quant math (round/clamp/scale) into the rounding-value and min/max-scale parameters, per datatype family, where the CE kernel handles standard weight/input gradients for one loss - and it would cost the bit-exact parity this lane currently guarantees against the single-shot computation

avtc:
Performance-Test-AutoRound is too strict for improvements:
```
Evaluating Model: Qwen/Qwen3-0.6B
------------------------------------------------------------
  [*] Tuning Time (s)     : Current = 330.783  | Baseline = 348.8896 (Diff: -5.19%) -> FAIL
  [*] Peak RAM (GB)       : Current = 3.63     | Baseline = 3.65     (Diff: -0.55%) -> PASS
  [*] Peak VRAM (GB)      : Current = 27.29    | Baseline = 27.21    (Diff: +0.29%) -> PASS
  [*] Output Size (GB)    : Current = 0.5137   | Baseline = 0.5137   (Diff: +0.00%) -> PASS
============================================================
❌ Performance check failed: Current metrics exceed acceptable limits compared to baseline.
```

wenhuach21:
> Performance-Test-AutoRound is too strict for improvements:
> 
> ```
> Evaluating Model: Qwen/Qwen3-0.6B
> ------------------------------------------------------------
>   [*] Tuning Time (s)     : Current = 330.783  | Baseline = 348.8896 (Diff: -5.19%) -> FAIL
>   [*] Peak RAM (GB)       : Current = 3.63     | Baseline = 3.65     (Diff: -0.55%) -> PASS
>   [*] Peak VRAM (GB)      : Current = 27.29    | Baseline = 27.21    (Diff: +0.29%) -> PASS
>   [*] Output Size (GB)    : Current = 0.5137   | Baseline = 0.5137   (Diff: +0.00%) -> PASS
> ============================================================
> ❌ Performance check failed: Current metrics exceed acceptable limits compared to baseline.
> ```

You can ignore this;  i think the performance test does not involve any LM‑head logic

auto_round/compressors/orchestrator.py
```diff
@@ -578,12 +598,23 @@ def _quantize_data_driven(self) -> tuple[torch.nn.Module, dict[str, Any]]:
...
                 quant_block_list=self.quant_block_list,
             )
+        # lm_head-class external layers are tail-fed: the block loop's chain
+        # output (fp reference + quantized rows through the final norm) is
+        # their input, so they neither join the upfront capture call nor the
+        # outside-block q-capture pass - no extra whole-model forwards
+        self._tail_fed_layers_ = []
+        self._lm_head_chain_tail_ = None
+        self._lm_head_norm_name_ = None
+        lm_head_name = self._resolve_lm_head_name_(layer_names)
```
wenhuach21: we already  have a  func get_lm_head_name in utils/models.py

auto_round/compressors/orchestrator.py
```diff
@@ -578,12 +598,23 @@ def _quantize_data_driven(self) -> tuple[torch.nn.Module, dict[str, Any]]:
...
+        if lm_head_name is not None:
+            self._tail_fed_layers_ = [lm_head_name]
+            self._lm_head_norm_name_ = self._discover_final_norm_(all_blocks)
```
wenhuach21: This makes a strong assumption about the model architecture, which does not align with AutoRound’s accuracy-first design philosophy. I’d suggest 
1 triggering this behavior only for verified models, e.g., [this check](https://github.com/intel/auto-round/blob/main/auto_round/algorithms/quantization/sign_round/quantizer.py#L182), and falling back to the legacy behavior otherwise.
or
2 At a minimum, we should log a warning and provide an option for users to fall back to the legacy behavior for unverified models immediately before tuning.

auto_round/compressors/orchestrator.py
avtc: **1. Stronger lm_head detection via `get_output_embeddings`**

We checked transformers: `PreTrainedModel.get_output_embeddings()` is an authoritative mapping — the base implementation returns `self.lm_head`, and 125 model classes override it per family (gpt_neox → `lm_head`, rwkv → `head`, …). Proposal: `get_lm_head_name` resolves via `get_output_embeddings()` first (reverse-mapping the returned module to its name), and keeps the current last-leaf walk as fallback for classes that define neither. This would improve all call sites of the helper. OK to include here, or would you prefer it as a separate PR, or skip?

auto_round/compressors/orchestrator.py
avtc: **2. Post-block wiring: measured, with legacy fallback**

Background we found while looking into it:
- the modules between the last decoder layer and lm_head vary more than a single norm: DeepSeek-V4 ends with `norm(hc_head(x))` — a stream-merging head followed by the norm; OPT ends with `project_out(final_layer_norm(x))` — norm followed by a projection. GPTQModel's per-family `pre_lm_head_norm_module` lists agree: one norm for nearly every family, with these same two-module exceptions.
- the family paths in the check you linked predate the Qwen3.5-wrapper family and return an empty post-block list there — its norm sits at `model.language_model.norm`, which none of the paths matches.
- name-based scans must also dodge the vision towers of multimodal checkpoints: Qwen3.6 carries a dozen outside-block norm-named leaves (`model.visual.merger.norm`, `model.visual.blocks.N.norm1/2`, …), and picking by module order works only while registration order happens to put the language norm last.

Proposal: resolve the chain by measurement. Before calibration starts, we temporarily replace the decoder layers with pass-through modules and run one dummy forward (dummy token ids, batch 1, length 2). Hooks record every module that executes between the last layer and lm_head, in order. We verify the recorded chain is linear: each module's input is the previous module's output, the first module's input is the last layer's output, and the last module's output is lm_head's actual input. The verified chain is then applied in order, in full precision, when deriving lm_head's inputs from the chain tail — so multi-module chains like DeepSeek-V4's and OPT's work too. If anything mismatches, or the dummy forward raises, we log a warning and run the whole outside-block handling the way current main does, from the start.

One edge case: a chain module can itself be a quantizable layer — OPT's `project_out` is a plain Linear between the final norm and lm_head. If it is in the quantization plan — whether the user marked it or the default scheme swept it in — the inputs we derive for lm_head would depend on a module that is being tuned in the same run: they become meaningful only after `project_out` finishes tuning, and lm_head would train on quantized-chain inputs where current main trains it on full-precision-chain inputs. We keep that combination on the legacy path for now and say why in the warning.

Risks we see: layer-call conventions differ across families (some return the hidden tensor bare, some as a tuple) — the pass-through tries the bare form first and the tuple form on error; models with unusual top-level control flow may raise on the dummy forward; both cases land in the legacy fallback described above.

Implementation options — pick what you prefer:

**1. Full version as described** — measured chain, multi-module chains supported (DeepSeek-V4, OPT with `project_out` kept full precision).

**2. Simplified version** — measured chain, but we adopt only single-norm chains for now; every other shape, including DeepSeek-V4's two-module chain, runs the legacy path until someone needs it. Smaller change, covers the common families (llama-style and the Qwen3.5 wrapper included).

**3. Verified-models gate** — your original option 1: the lane engages only for model types we have validated, with the list extended to the Qwen3.5-wrapper family; everything else keeps current main behavior. Smallest change, and the measurement probe becomes a follow-up.

Our preference is 1 or 2 — both replace architecture assumptions with a measured check and keep an automatic legacy fallback. OK to proceed with one of these, or do you have another approach in mind?

auto_round/compressors/orchestrator.py
wenhuach21: It's fine to use either approach to detect the LM head. My point is that we should consolidate these two functions.

auto_round/compressors/orchestrator.py
wenhuach21: I don't think this is straightforward, since not all ops are implemented as modules. For example, `F.relu` is a functional op rather than a module. I worked on graph tracing when I implemented SmoothQuant in Intel Neural Compressor before, and it can get quite complex.

avtc: @wenhuach21 
Agreed on tracing complexity - we dropped the detection approach and propose a simpler idea.

**Smoke check (before quantization).** We temporarily replace the decoder layers with pass-through stubs and run a two-token forward with the real lm_head. This checks per model that the wrapper is callable text-only, that the stub shapes fit the layer loop, and that the model's own post-block code runs. The result sets immediate packing and model materialization for the run.

**Capture (after the block loop).** By then the decoder blocks are streamed and released, and the last block's output (fp + quantized) is cached. We swap lm_head for a capture stub, place the cached tail at the last layer slot, and run one forward. Transformers executes its own post-block code - final norm and any other ops, functional or modular - and the capture stub records the exact rows the real model feeds the head, returning a tiny dummy instead of logits. We then restore the real lm_head.

**Tuning.** The captured rows (one set per fp/quantized tail) drive the existing chunked tuning unchanged. During the mocked pass only the embeddings, the post-block modules, and the head are resident.

**Fallback.** If the smoke check fails, we fall back to full model materialization - the conservative path, matching main's behavior today.

An audit of the post-block region across the transformers architectures backs the design: most models end with a single final norm; a few add pre-head scalings, dtype casts, or stream-mixing modules after the loop. All of it executes as the model's own code in the mocked pass.

What do you think - will it work? We can implement it in this PR.

avtc: @wenhuach21

We implemented the design from our earlier reply in a second commit (64931b988) - please take a look.

Smoke check (init time). The decoder layers are temporarily replaced with pass-through stubs and a two-token forward runs with the real lm_head. This validates per model that the model is callable text-only, that the stub return shape fits the layer loop (single-element tuple, residual pair, or bare tensor - discovered empirically), and that the model's own post-block code runs. Pass -> mocked capture with immediate packing; fail -> the ordinary capture walk with the lane's relaxations reverted, matching main's behavior.

Capture (after the block loop). lm_head is swapped for a capture head, the cached last-block output (fp + quantized variants) is placed at the last layer slot, and one forward per sample executes the model's own post-block code - final norm, pre-head scalings, dtype casts, stream mixers - while the capture head records the exact rows the real model feeds the head. Captured rows are validated against the head input width: post-block code that reshapes or splits streams (e.g. ProphetNet's ngram split) declines loudly and falls back to zero-shot RTN.

Name resolution. get_lm_head_name now prefers model.get_output_embeddings() (guarded), then a named_modules identity reverse-map, then the original last-leaf walk. The name-based final-norm detection is replaced by executing the model's own code: a 255-architecture audit of the transformers post-block region showed name-based reconstruction silently wrong on pre-head scalings, dtype casts, and multi-module chains, while the mocked pass covers every tail class by construction.

Validation. An architecture-matrix e2e runs one model per tail class - qwen3 (final norm), minicpm3 (pre-head scaling), mamba (dtype cast), gpt_bigcode (post-loop view), bart (empty post-loop tail), electra (head-side chain), xlm_roberta (compound head engaging through the inner vocab Linear) - all produce rows and quantized weights identical to the ordinary capture walk; ProphetNet asserts the loud decline. On Qwen3.8-27B (248k vocab, lm_head-only W4A16 g128 sym, iters=10, single 3090) the lane engages end to end (imatrix from 262144 chain-tail rows), lm_head stage 53 s.

**Two input-cache findings from the matrix (shared calibration path, upstream of this lane).** The block-input collection (`auto_round/calibration/llm.py`) caches layer kwargs of type Tensor/list/tuple/str/bool/None; dict-valued kwargs fall through to a skip branch (it prints "attention_mask key is not currently used in quantization fine-tuning"). Two current architectures pass dict kwargs to their layers and hit this:

1. Zaya passes `attention_mask` as a dict `{"causal": ..., "conv": ...}`. The capture walk runs materialized masks while block replays run implicit-mask attention; on a tiny random-weight model the ~5e-5 hidden difference is amplified by the pre-head RMSNorm to a 0.0098 rows difference in our parity test. Caching the dict kwarg experimentally drops it to 2.7e-05, so a shared-raw-storage fix for dict kwargs in the collection would close it.
2. deepseek_v4 takes a required `position_embeddings` dict `{"main", "compress"}` plus a raw `input_ids` kwarg (its MoE gate routes tokens by id through a tid2eid table). The dropped dict makes block replay fail with "missing required argument"; after caching it experimentally, the next failure is that `split_inputs` (`auto_round/calibration/inputs.py`) pops the layer's cached `input_ids` as the chain-input slot, so the replay passes `input_ids=None` into the gate. The official Intel DeepSeek-V4 artifact used the model-free lane, so data-driven block replay is unexercised upstream for this arch.

Both are reproducible without real weights - the two cases in test_lm_head_tail_arch_matrix_e2e.py build tiny config-only models and document the exact mechanisms (kept as local reproducers, marked skip_ci so they stay out of CI runtime).

PR #2385, opened 2026-09-20
Cast float32 weights to the export dtype when saving

### LINKED ISSUE #2293: [Bug]: CPU quantization can't get proper model size
### Problem Description

Qwen3.8-27B-FP8/Qwen3.8-27B-fp-w8afp8/, total 83G
Quantized model8 size larger than BF16

### Reproduction Steps

auto-round \
  --model_name Qwen/Qwen3.8-27B \
  --scheme FP8_STATIC \
  --fp_layers "visual,linear_attn,mtp,lm_head,embed_tokens" \
  --iters 0 \
  --device_map auto \
  --format auto_round \
  --output_dir ./Qwen3.8-27B-FP8

### Environment Information

_No response_

### Error Logs

```shell

```

### Additional Context

_No response_

Shubham-Padkonde:
Fixes #2293

## Problem

On a CPU without bfloat16 support, `ModelContext._set_amp_dtype` falls back to float32, so the model is tuned and held in float32. Export still writes `dtype`/`torch_dtype` into the config (bfloat16 by default), but `save_pretrained` serializes whatever the model currently holds — float32. The saved artifact is therefore roughly twice the expected size and its tensors disagree with its own config. The same quantization run on a bf16-capable device produces a correctly sized model, which is what the issue reports.

## Fix

`save_model` now builds the state dict it hands to `save_pretrained`, casting float32 tensors to the declared export dtype. The model itself is not mutated, and:

- only float32 tensors are touched, so already-quantized tensors (fp8, int32 zeros/indices/scales) pass through untouched;
- modules the model deliberately keeps in float32 (`_keep_in_fp32_modules`, `_keep_in_fp32_modules_strict`) are preserved;
- the helper returns `None` when the export dtype is not a 16-bit float or no float32 tensors exist, so the previous call path is used unchanged;
- the manual `save_file` fallback (for third-party configs that break `save_pretrained`) uses the same state dict.

## Verification

Reproduced with a tiny BF16 model quantized to FP8_STATIC on CPU, on `main`:

| run | output size | float32 tensors |
|---|---|---|
| CPU (`supports_bf16` False), before fix | 9.77 MB | 39 |
| bf16-capable device (emulated by patching `CpuARDevice.supports_bf16`) | 4.90 MB | 0 |
| CPU (`supports_bf16` False), after fix | 4,899,284 bytes | 0 |

The post-fix CPU output is byte-for-byte the same size as the bf16-device output, with **0 dtype differences** across all tensors when the two state dicts are compared key by key.

## Tests

Added `TestGetStateDictForExportDtype` and a `save_model` case in `test/unit/common/export/test_export_utils.py` covering: casting float32 to the export dtype, leaving quantized/integer tensors alone, honouring `_keep_in_fp32_modules`, not mutating the model, and the `None` short-circuits.

```
$ pytest test/unit/common/export/test_export_utils.py
53 passed
```

The three pre-existing collection errors in `test/unit/common/export` (`test_conversion_base.py`, `test_gguf_dtype_helpers.py`, `test_gguf_moe_adapter.py`) are `ModuleNotFoundError: No module named 'gguf'` and are present on a clean tree as well. `black`, `isort` and `ruff` are clean on both changed files.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The test, lint and reproduction output quoted above comes from real runs in my local environment; I am accountable for what is submitted here and will follow up on review feedback.


Shubham-Padkonde:
The `Performance-Test-AutoRound (Qwen3_W4A16)` failure looks unrelated to this change — it is the peak-RAM comparison, and the metric is reached before export:

```
  [*] Tuning Time (s)     : Current = 349.8516 | Baseline = 351.1587 (Diff: -0.37%) -> PASS
  [*] Peak RAM (GB)       : Current = 3.66     | Baseline = 3.52     (Diff: +3.98%) -> FAIL
  [*] Peak VRAM (GB)      : Current = 27.21    | Baseline = 27.4     (Diff: -0.69%) -> PASS
  [*] Output Size (GB)    : Current = 0.5137   | Baseline = 0.5137   (Diff: +0.00%) -> PASS
```

Two things point away from this PR:

- That job logs `Immediate saving mode: weights already saved by ShardWriter, saving configs only`, which returns from `save_model` before the branch this PR touches. The code added here is never executed in that run — consistent with the output size being identical to the baseline.
- `peak_ram` grows monotonically through tuning in the job log (2.04 GB → 2.88 → … → 3.66 GB across the 28 blocks), so the reported peak is reached during quantization, well before export.

The `Qwen3_FP8_STATIC` performance job, which does exercise the changed path, passes.

Happy to rebase or push an empty commit if that helps re-run the job — just let me know.

wenhuach21 (approved):

PR #2386, opened 2026-09-20
Don't rewrite packed modules as float weights when resuming

### LINKED ISSUE #2350: [Bug]: crash + resume (AR_RESUME_DIR, immediate saving) silently duplicates already-quantized modules as stale fp weights in the final artifact
## Evidence (opt-125m e2e)

crash after 2 blocks -> resume -> artifact contains 12 duplicated tensors = exactly the 6 quantized Linears of the 2 skipped blocks (each with both `qweight` and `.weight`). No warning is logged.

Qwen3.8-27B W4A16 artifact came out 47.4 GB instead of ~17.7 GB with 318/407 modules duplicated this way — silently.

## Root cause

`ShardWriter.finalize()`'s remaining-weights pass writes every tensor still present in the model tree whose exact NAME was not saved yet. Resumed-over blocks present plain `nn.Linear` modules (no wrapper), so their packed names (`qweight` etc., saved before the crash) never match the plain names — the name-exact dedup misses and the stale fp weight is written on top of the packed one.

### Reproduction Steps

1. run any quantization with immediate saving + `AR_RESUME_DIR=<dir>`
2. kill the process after >= 1 block completes (manifest marks it done)
3. rerun the same command with the same `AR_RESUME_DIR` -> run resumes, skipped blocks are not re-quantized
4. inspect the final artifact: every module of a skipped block exists TWICE — the packed tensors saved before the crash AND a plain fp `.weight` written at save time

### Environment Information

- auto-round: reproduced on a build based on upstream `main` @ `32369a69`; `ShardWriter.finalize()` is unchanged through current `main` @ `9e2650f4`. transformers 5.16.1, torch 2.11.0+cu128, immediate-saving path

### Error Logs

```shell

```

### Additional Context

_No response_

Shubham-Padkonde:
Fixes #2350

## Problem

When a run is resumed through `AR_RESUME_DIR`, the blocks a previous run already quantized are skipped, so their modules are still plain `nn.Linear` in the model tree while their packed tensors sit in the shard files the crashed run flushed.

`ShardWriter.finalize()`'s remaining-weights pass deduplicates by exact tensor name, so `<layer>.weight` never matches the saved `<layer>.qweight`: the stale floating-point weight is written on top of the packed one. Every module of every skipped block ends up in the checkpoint twice, with no warning. The issue reports a Qwen3.8-27B W4A16 artifact coming out at 47.4 GB instead of ~17.7 GB, with 318/407 modules duplicated.

There is a second half to it: discovery of the previous run's shards (`_discover_existing_shards`) is deferred to the first `_flush_shard()` call, which `finalize()` only reaches *after* the remaining-weights pass. So that pass ran with an empty `_all_saved` for anything the crashed run wrote — even tensors whose names match exactly would have been written twice.

## Fix

- Adopt the previous run's shards at the start of `finalize()`, before deciding which weights are still missing. `output_dir` is already the final path there, which is the reason the call was deferred out of `__init__` in the first place.
- Skip a module's unpacked `weight` when packed tensors for that module are already saved (`qweight` / `weight_packed`), and log how many were skipped so this is no longer silent.

Modules that were never packed are unaffected, and non-resuming runs never reach the new discovery call because it stays gated on `AR_RESUME_DIR`.

## Test

`test_finalize_skips_unpacked_weight_of_resumed_packed_module` writes a shard in the crashed run's temp layout containing `linear.qweight` / `linear.scales` / `linear.bias`, points `AR_RESUME_DIR` at it, and finalizes a writer whose model still holds the plain `nn.Linear`.

Without the change:

```
AssertionError: the packed module must not also be saved as a floating-point weight
1 failed, 7 passed
```

With it, all 8 pass and `proj_out.weight` — a module that was never packed — is still saved.

```
$ pytest test/unit/common/utils test/unit/common/export/test_export_utils.py
5 failed, 923 passed, 6 skipped
```

The same 5 failures (`test_alg_ext`, `test_layer_config_resolution`, `test_weight_handler`, two `TestIsImmediateSavingMode` cases) reproduce on an unmodified `main` with the same selection — 922 passed there, so this change adds a test and breaks nothing. `black`, `isort` and `ruff` are clean.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The test, lint and reproduction output quoted above comes from real runs in my local environment; I am accountable for what is submitted here and will follow up on review feedback.


xin3he (approved):
Nice catch! Thanks

xin3he:
@Shubham-Padkonde  Could you please address these high-priority comments?

PR #2393, opened 2026-09-20
Apply autocast only when mixed precision is enabled

Shubham-Padkonde:
## Problem

The autocast context in `SignRoundQuantizer._get_loss` is selected with an inverted condition:

```python
autocast_ctx = (
    nullcontext()
    if self.model_context.amp                # mixed precision ON  -> no autocast
    else autocast(device_type=..., dtype=self.model_context.amp_dtype)   # OFF -> autocast
)
```

`ModelContext._set_amp_dtype` sets `amp_dtype = torch.float32` when `amp` is disabled, so with mixed precision off every loss computation enters an autocast block with a dtype autocast does not support. On CPU that is visible in any run:

```
UserWarning: In CPU autocast, but the target dtype is not supported. Disabling autocast.
CPU Autocast only supports dtypes of torch.bfloat16, torch.float16 currently.
  else autocast(device_type=str(device).split(":")[0], dtype=self.model_context.amp_dtype)
```

and conversely, a run that asked for mixed precision computes its loss outside autocast.

The two other call sites — one further down the same file (line ~758) and one in `sign_roundv2/quantizer.py` — already write the condition the right way round, so this looks like a copy/paste slip rather than intent.

## Fix

One character of logic: `if not self.model_context.amp`, matching the sibling call sites.

## Test

`test/unit/common/algorithms/test_autocast_context.py` calls `_get_loss` with a stub context. With mixed precision disabled it turns `UserWarning` into an error, so entering autocast with float32 fails the test; it also checks the loss is still finite with mixed precision enabled for bfloat16 and float32.

Without the fix:

```
E   UserWarning: In CPU autocast, but the target dtype is not supported. Disabling autocast.
1 failed, 2 passed
```

With it:

```
$ pytest test/unit/common/algorithms/test_autocast_context.py
3 passed

$ pytest test/unit/common/algorithms
650 passed, 1 skipped
```

`black`, `isort` and `ruff` are clean.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The output quoted above comes from real runs in my local environment; I am accountable for what is submitted here and will follow up on review feedback.


wenhuach21:
Nice catch, thanks for the fix!

PR #2395, opened 2026-09-20
Honour the padding mask in SignRoundV2's plain loss

Shubham-Padkonde:
## Problem

`SignRoundV2Quantizer._get_loss` takes a `valid_token_mask` and applies it on the outlier-suppressed path, but drops it when it hands off to the base implementation:

```python
        return super()._get_loss(pred_output, ref_output, indices, mse_loss, device)
```

`SignRoundQuantizer._get_loss` masks padded tokens out of the loss when it receives that argument. So with the algorithm extension enabled and padded calibration batches — that is, whenever `_use_outlier_suppressed_loss` is false — padding is scored as if it were real data and the tuning objective is wrong.

For a two-row batch whose second row is padding:

```
without the mask: 13.0
with the mask:     0.5
```

## Fix

Pass the mask through.

## Test

`test/unit/common/algorithms/test_signroundv2_loss.py` calls `_get_loss` on a `SignRoundV2Quantizer` with a mask that marks the second row as padding, and checks the loss is the masked value. It also covers the outlier-suppressed path, which already handled the mask.

Without the change:

```
E    +    and   tensor(13.) = _loss(use_outlier_suppressed_loss=False)
E    +    and   tensor(0.5000) = torch.tensor(0.5)
1 failed, 1 passed
```

With it:

```
$ pytest test/unit/common/algorithms/test_signroundv2_loss.py
2 passed

$ pytest test/unit/common/algorithms
649 passed, 1 skipped
```

`black`, `isort` and `ruff` are clean.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The output quoted above comes from real runs in my local environment; I am accountable for what is submitted here and will follow up on review feedback.


wenhuach21 (approved):
Thanks for the fix! How do you manage to find so many subtle bugs?

PR #2340, opened 2026-09-10
feat: add opt-in diffusion tuning cache prefetch

changwangss:
## Motivation                                                                                                                                                                                             
                                                                                                                                                                                                            
  Repeated CPU-to-GPU cache transfers can become a bottleneck during Wan2.2 tuning with `low_gpu_mem_usage=True`.

  For Wan2.2 T2V-A14B at 480×832 resolution and 81 frames:
  - Video tokens: 21 × 30 × 52 = 32,760.
  - One BF16 hidden-state tensor: 32,760 × 5,120 × 2 bytes ≈ 0.312 GiB.
  - A cached block input and reference output together occupy approximately 0.625 GiB per sample, assuming both are BF16.
  - For 128 cached samples, these tensors alone require approximately 80 GiB, excluding conditioning tensors, model weights, and tuning activations.

  Keeping these caches on CPU reduces GPU memory usage, but repeatedly transferring selected batches can leave the GPU waiting for data.


## Description
  Add `diffusion_tuning_cache_size` to enable bounded prefetching with two reusable buffers and optional GPU storage for best-parameter snapshots. Set it to `"auto"` to determine the budget after warmup,
  or specify a budget in GiB. Insufficient memory falls back to the existing path.

  Disabled by default and limited to single-GPU diffusion SignRound tuning.

  Usage example (single-GPU Wan2.2 T2V-A14B):

  ```bash
  CUDA_VISIBLE_DEVICES=0 auto-round \
      --model Wan-AI/Wan2.2-T2V-A14B-Diffusers \
      --scheme MXFP4 \
      --model_dtype bf16 \
      --format fake \
      --batch_size 1 \
      --dataset coco2014 \
      --calib_num_inference_steps 8 \
      --low_gpu_mem_usage \
      --diffusion_tuning_cache_size auto \
      --output_dir ./wan2.2-a14b-mxfp4
  ```

  `auto` selects the extra GPU cache budget after the first tuning iteration. Alternatively, use a numeric value such as `8` for an 8 GiB cache budget. This budget applies to the additional cache buffers,
  not total GPU memory usage.

  The feature requires `low_gpu_mem_usage`. Omit `diffusion_tuning_cache_size` or set it to `0` to retain the existing behavior.

## performance improve with wan2.2 T2V
device: b300

main branch, quantize transformer about need 5 hours 44 mins.
```
[...]

PR #2404, opened 2026-09-21
refactor: move diffusion tuning data handling out of SignRound

changwangss:
## Description

  Follow up on #2340 review comments:

  - Move cache eligibility checks into DiffusionTuningCache.
  - Extract batch staging and fallback into BlockForwardRunner.forward_tuning_batch, reusing the existing forward.
  - Remove try/finally and close the cache after the tuning loop.

  Keep the existing diffusion-only scope, warmup timing, and best-parameter logic.

## Type of Change

Bug fix

## Related Issues

Fixes or relates to #


auto_round/algorithms/quantization/sign_round/quantizer.py
```diff
@@ -477,101 +477,97 @@ def quantize_block(
...
-                if self.not_use_best_mse and i == self.iters - 1:
+            use_tuning_cache = DiffusionTuningCache.is_enabled(
+                self.model_context, self.compress_context, device, loss_device
+            )
+
+        for i in range(self.iters):
+            # Auto observes a complete forward/backward/optimizer iteration
+            # on the legacy path before allocating any extra GPU buffers.
+            if use_tuning_cache and i == (1 if cache_budget == "auto" else 0):
+                tuning_cache = DiffusionTuningCache.create(
```
wenhuach21, line 492: Is it possible to move all the cache-related functionality into the base quantizer or other relevant classes, and expose a common interface for accessing it. This would allow us to support caching across all algorithms through the same interface and make it easier to extend the support to LLMs and other use cases.

Personally, I don't think algorithm developers should have to care about implementation details that are not directly related to their algorithms.

